Issue #137, opened 2024-03-26
GCP: Persistent Volume pricing is incorrect 


Pokom:
After rolling out `cloudcost-exporter` I discovered that quite a few pvc's were reporting roughly twice as expensive as they should be. I setup some basic unit tests to validate that single sku's were being parsed correctly, which led to me recognize that some storage classes were getting overwritten with regional prices as well. 

`cloudcost-exporter` should be not overwrite existing entries in the storage map.

PR #168, opened 2024-05-07
docs(gke): Update persistent volume cost metric name

Pokom:
Related to #137, after rolling out the change to update the metric name, I forgot to update the documentation. This updates the docs to reflect that the cost is hourly.


Pokom:
This is unnecessary, I had stale docs in my local branch 🤦🏻

PR #169, opened 2024-05-07
feat(eks): Implement base eks module

### LINKED ISSUE #67: Implement AWS EKS cost metrics in `cloudcost-exporter`
# Overview

CloudCost Exporter needs to implement a module within aws that performs the following:
- Generates a pricing map for ec2 instances 
- Collects a list of running ec2 instances within an AWS profile
- Implements a collector that exports the hourly cost of a machine

There was a proof of concept done many months ago: https://github.com/grafana/deployment_tools/pull/96916
Note that the project structure has been completely refactored so you can't use it as is.
However the api endpoints and clients could be useful for helping bootstrap the issue.

## Pricing Map

One major deviation from other Cloud Providers is that AWS only provides pricing by instance type. 
The implication is that we need to make an estimate of how much an instance costs by CPU and Memory. 
One way we can do this is by looking at GCP prices and finding the ratio of CPU:Memory costs. 

Unknowns:
- Can we pull spot pricing information from api?

Some considerations:
- Cache the pricing map for a specific period of time
- Can we infer the regions the profile is running in and limit the amount of skus to parse?

## Collecting a list of running ec2 instances

In order to export metrics, we need to list of each of the ec2 instances that are running within a project/profile.
We need to keep track of the following bits of metadata:
- instance name
- region it's running in
- family of the machine(T2, M2, etc)
- machine type(m7g.medium, m5.large)
- Number of Cores
- GiB of ram
- profile/project associated with the instance
- OnDemand or a spot instance

Some considerations:
- Performance here will be key, the list of instances needs to be collected in under 60 seconds
- We can use a cache and be okay with the current exported metrics to be from a previous pull
- Parallize the requests by regions/zones if possible

## Export Metrics

Use the pricing map and ec2 instances to export the following metrics:
- `cloudcost_aws_ec2_instance_cpu_usd_per_core_hour`
- `cloudcost_aws_ec2_instance_cpu_usd_per_gibyte_hour`

Both metrics should have the following labels
- instance
- region
- family
- machine_type
- project|profile (depending on terminology, profile may make more sense)
- price_tier (spot|ondemand)

Pokom:
This is the first attempt at implementing the eks module. The following is implemented
- List of list prices of on demand ec2 instances for regions profile has activated
- Generate pricing map to look up by region and machine type
- List of instances in all regions for configured profile
- Export cost metrics by ec2 instance for compute and memory

TODO:
- [x] Generate cost map for spot instances
- [x] List instances for other profiles
- [x] Collect instances in parallel 

- closes #67


pkg/aws/eks/eks.go
```diff
@@ -0,0 +1,524 @@
...
+	options = append(options, awsconfig.WithSharedConfigProfile(profile))
+	ac, err := awsconfig.LoadDefaultConfig(context.Background(), options...)
+	if err != nil {
+		return nil, err
+	}
+	client := ec2.NewFromConfig(ac)
+	dii := &ec2.DescribeInstancesInput{
+		// TODO: Is 1000 appropriate?
+		MaxResults: aws.Int32(1000),
+	}
```
logyball, line 466: I wonder if something like this may improve performance?

```suggestion
	dii := &ec2.DescribeInstancesInput{
		// TODO: Is 1000 appropriate?
		MaxResults: aws.Int32(1000),
		Filters:    []ec2Types.Filter{{Name: aws.String("tag-key"), Values: clusterTags}, {}},
	}
```

This way, only ec2 instances that have the one of the cluster tags [(values are OR'd)](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/ec2@v1.161.1/types#Filter) will be returned?

Pokom: Right now we're able to process the compute instances in prod in ~10s. Let me add it and see what happens 🤔

Pokom: I'm going to run this a bit longer tomorrow and report back. IE, let me run it for ~30 minutes with the filter and 30 minutes without so I can get a proper comparison.

Pokom: image
The first half of the time series is with the filter suggested, and the last half is without the filter. It's really hard to tell, but it doesn't make an obvious improvement in speed. I'm going to leave out the filter for now, but leave a comment for the future that it's possible to filter by tags and to benchmark it.

pkg/aws/eks/README.md
```diff
@@ -0,0 +1,23 @@
+# EKS Module
+
+This module is responsible for collecting pricing information for EKS clusters.
+Specifically it collects the pricing information for the following resources:
+- compute instances
+- storage
+
+## Compute Instances
```
the-it, line 8: Let me have a look at this in the morning ... I have some calculations that I wanna try

pkg/aws/eks/eks.go
```diff
@@ -0,0 +1,360 @@
...
+	pricingClient "github.com/grafana/cloudcost-exporter/pkg/aws/services/pricing"
+	"github.com/grafana/cloudcost-exporter/pkg/provider"
+)
+
+const (
+	subsystem = "eks"
+)
+
+// These were generated by analysing Grafana Labs spend in GCP and finding the ratio of CPU to Memory spend by instance type.
+// It's an imperfect approximation, but it's better than nothing.
```
the-it, line 29: do we have an idea how other solutions are dealing with this?

pkg/aws/eks/eks.go
```diff
@@ -0,0 +1,360 @@
...
+	pricingMap      *StructuredPricingMap
+	pricingService  pricingClient.Pricing
+	ec2Client       ec2client.EC2
+	NextScrape      time.Time
+	ec2RegionClient map[string]ec2client.EC2
+}
+
+func (c *Collector) CollectMetrics(metrics chan<- prometheus.Metric) float64 {
+	//TODO implement me
+	panic("implement me")
```
the-it, line 102: is here something todo?

pkg/aws/eks/eks.go
```diff
@@ -0,0 +1,360 @@
...
+}
+
+func (c *Collector) Collect(ch chan<- prometheus.Metric) error {
+	if c.pricingMap == nil {
+		var prices []string
+		var spotPrices []ec2Types.SpotPrice
+		for _, region := range c.Regions {
```
the-it, line 109: is this fine from a runtime perspective? If I understand this correctly, we do a (two) serial calls to the pricing api for every region we have to deal with? ... I guess we can revisit some channel magic if we have to, (or cache the prices for a little while)

the-it: reading further in the function: Can't we adapt the strategy taken for listing the instances?

pkg/aws/eks/eks.go
```diff
@@ -0,0 +1,360 @@
...
+	ErrClientNotFound       = errors.New("no client found")
+	ListSpotPricesErr       = errors.New("error listing spot prices")
+	GeneratePricingMapErr   = errors.New("error generating pricing map")
+)
+
+var (
+	InstanceCPUHourlyCostDesc = prometheus.NewDesc(
+		prometheus.BuildFQName(cloudcostexporter.MetricPrefix, subsystem, "instance_cpu_usd_per_core_hour"),
+		"The cpu cost a compute instance in USD/(core*h)",
+		[]string{"instance", "region", "family", "machine_type", "cluster", "price_tier", "provider"},
```
the-it, line 49: is there a point to have provider here? ... eks is already in the name of the metric

pkg/aws/eks/eks.go
```diff
@@ -0,0 +1,360 @@
...
+				ch <- prometheus.MustNewConstMetric(InstanceCPUHourlyCostDesc, prometheus.GaugeValue, price.Cpu, labelValues...)
+				ch <- prometheus.MustNewConstMetric(InstanceMemoryHourlyCostDesc, prometheus.GaugeValue, price.Ram, labelValues...)
+			}
+		}
+	}
+}
+
+func CollectMetrics(_ chan<- prometheus.Metric) error {
+	//TODO implement me
+	panic("implement me")
```
the-it, line 205: ?

the-it (approved):
nothing is blocking ... but I have some questions

pkg/aws/eks/README.md
the-it: Yeah it's probably the best estimation we can do at the moment ... I will leave this rabbit hole again.

pkg/aws/eks/eks.go
Pokom: THe main thing is that the `CollectMetrics` interface is a leaky abstraction and should be deprecated 😅 For now I'll just return 1 or whatever the appropriate method is.

pkg/aws/eks/eks.go
Pokom: The main reason why I didn't is because performance isn't important in this part of the code path and fetching prices adds a bit of complexity. I'm not opposed to fetching in parallel since it's not the hardest to add/handle. Also looks like I missing the most important bit here: Adding logic to refresh the pricing 😅

pkg/aws/eks/eks.go
Pokom: I suppose not. It doesn't make sense since we can't aggregate metrics between providers anyway.

pkg/aws/eks/eks.go
Pokom: 🤔 thought I provided a response. `CollectMetrics` is a leaky abstraction that I'm looking to remove in a future PR. For now I'll return nil since panic'ing seems excessive.

pkg/aws/aws.go
```diff
@@ -187,3 +207,15 @@ func (a *AWS) Collect(ch chan<- prometheus.Metric) {
...
+func newEc2Client(region, profile string) (*ec2.Client, error) {
+	options := []func(*awsconfig.LoadOptions) error{awsconfig.WithEC2IMDSRegion()}
+	options = append(options, awsconfig.WithRegion(region))
+	options = append(options, awsconfig.WithSharedConfigProfile(profile))
+	ac, err := awsconfig.LoadDefaultConfig(context.Background(), options...)
+	if err != nil {
+		return nil, err
+	}
+	client := ec2.NewFromConfig(ac)
+	return client, nil
```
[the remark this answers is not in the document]
Pokom: Personally don't think it's an ultra nit. This is a solid suggestion that follows Go's styleguide pretty well that `golangci-lint` wouldn't catch. 👍🏻

pkg/aws/aws.go
```diff
@@ -87,41 +93,55 @@ var (
...
+			}
+			regionClientMap := make(map[string]ec2client.EC2)
+			for _, r := range regions.Regions {
+				client, err := newEc2Client(*r.RegionName, config.Profile)
+				if err != nil {
+					return nil, fmt.Errorf("error creating ec2 client: %w", err)
+				}
+				regionClientMap[*r.RegionName] = client
+			}
+			collector := eks.NewCollector(config.Region, config.Profile, config.ScrapeInterval, pricingService, computeService, regions.Regions, regionClientMap)
```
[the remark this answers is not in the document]
Pokom: > One other super NIT is that the s3 New() function returns a checked error, but it's always nil, can we remove that error check? Just so we have a consisted interface for the "New Collector" function

💯 I was considering making the change to S3 to no longer return an error. That's from a previous times where `s3.New` creating the clients within the method instead of having them passed in.

pkg/aws/aws.go
```diff
@@ -87,41 +93,55 @@ var (
...
 			client := costexplorer.NewFromConfig(ac)
 			collector, err := s3.New(config.ScrapeInterval, client)
 			if err != nil {
 				return nil, fmt.Errorf("error creating s3 collector: %w", err)
 			}
 			collectors = append(collectors, collector)
+		case "EKS":
+			pricingService := pricing.NewFromConfig(ac)
+			computeService := ec2.NewFromConfig(ac)
+			regions, err := computeService.DescribeRegions(context.TODO(), &ec2.DescribeRegionsInput{AllRegions: aws.Bool(false)})
```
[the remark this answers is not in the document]
Pokom: > [NIT] - we should consider implementing a ctx.WithTimeout() here? But that's outside of the scope of this. For this PR I'd consider either:

In the future yes I think we should. I like your recommendation and will implement that. Thanks 👍🏻

pkg/aws/eks/pricing_map.go
```diff
@@ -0,0 +1,167 @@
...
+
+// GeneratePricingMap accepts a list of ondemand prices and a list of spot prices.
+// The method needs to
+// 1. Parse out the ondemand prices and generate a productTerm map for each instance type
+// 2. Parse out spot prices and use the productTerm map to generate a spot price map
+func (spm *StructuredPricingMap) GeneratePricingMap(ondemandPrices []string, spotPrices []ec2Types.SpotPrice) error {
+	for _, product := range ondemandPrices {
+		var productInfo productTerm
+		if err := json.Unmarshal([]byte(product), &productInfo); err != nil {
+			return err
```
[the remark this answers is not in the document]
Pokom: I think this scenario stuck out the most because if we're unable to unmarshal one price, we're likely unable to unmarshal others. The other scenarios we skip are less problematic(imo).

scripts/aws-spot-pricing/main.go
```diff
@@ -0,0 +1,70 @@
...
+	"github.com/aws/aws-sdk-go-v2/config"
+	"github.com/aws/aws-sdk-go-v2/service/ec2"
+	"github.com/aws/aws-sdk-go-v2/service/ec2/types"
+)
+
+func main() {
+	options := []func(*config.LoadOptions) error{config.WithEC2IMDSRegion()}
+	options = append(options, config.WithRegion("us-east-2"))
+	options = append(options, config.WithSharedConfigProfile("workloads-dev"))
+	cfg, err := config.LoadDefaultConfig(context.TODO(), options...)
```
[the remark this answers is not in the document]
Pokom: Ah I'm leaking information with this script, good catch Logan!

pkg/aws/eks/eks.go
```diff
@@ -0,0 +1,362 @@
...
+			if c.ec2RegionClient[*region.RegionName] == nil {
+				return ErrClientNotFound
+			}
+			client := c.ec2RegionClient[*region.RegionName]
+			spotPriceList, err := ListSpotPrices(context.Background(), client)
+			if err != nil {
+				return fmt.Errorf("%w: %w", ErrListSpotPrices, err)
+			}
+			spotPrices = append(spotPrices, spotPriceList...)
+			c.NextScrape = time.Now().Add(c.ScrapeInterval)
```
logyball, line 131: I think this is just supposed to be the next scrape on line 137

```suggestion
```

pkg/aws/eks/eks.go
```diff
@@ -0,0 +1,362 @@
...
+			instanceCh <- reservations
+		}(region)
+	}
+	go func() {
+		wg.Wait()
+		close(instanceCh)
+	}()
+	c.emitMetricsFromChannel(instanceCh, ch)
+	return nil
+}
```
logyball, line 162: In a very unscientific test, parallelizing this to 5 concurrent goroutines took a ton of time off:

```sh
# without parallelizing
❯ curl localhost:8080/metrics | grep -v "go_"                              48m 32s
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 52989    0 52989    0     0   1036      0 --:--:--  0:00:51 --:--:-- 11127# HELP cloudcost_eks_instance_cpu_usd_per_core_hour The cpu cost a compute instance in USD/(core*h)

# with parallelizing to 5 goroutines
❯ curl localhost:8080/metrics | grep -v "go_"                                  10s
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:--  0:00:19 --:--:--     0# HELP cloudcost_eks_instance_cpu_usd_per_core_hour The cpu cost a compute instance in USD/(core*h)
# TYPE cloudcost_eks_instance_cpu_usd_per_core_hour gauge
```

```suggestion
// Collect satisfies the provider.Collector interface.
func (c *Collector) Collect(ch chan<- prometheus.Metric) error {
	if c.pricingMap == nil || time.Now().After(c.NextScrape) {
		var prices []string
		var spotPrices []ec2Types.SpotPrice
		eg := new(errgroup.Group)
		eg.SetLimit(5)
		lock := sync.Mutex{}
		for _, region := range c.Regions {
			eg.Go(func() error {
				priceList, err := ListOnDemandPrices(context.Background(), *region.RegionName, c.pricingService)
				if err != nil {
					return err
				}
				lock.Lock()
				prices = append(prices, priceList...)
				lock.Unlock()
				if c.ec2RegionClient[*region.RegionName] == nil {
					return ErrClientNotFound
				}
				client := c.ec2RegionClient[*region.RegionName]
				spotPriceList, err := ListSpotPrices(context.Background(), client)
				if err != nil {
					return fmt.Errorf("%w: %w", ErrListSpotPrices, err)
				}
				lock.Lock()
				spotPrices = append(spotPrices, spotPriceList...)
				lock.Unlock()
				return nil
			})
		}

		err := eg.Wait()
		if err != nil {
			return err
		}

		c.pricingMap = NewStructuredPricingMap()
		if err := c.pricingMap.GeneratePricingMap(prices, spotPrices); err != nil {
			return fmt.Errorf("%w: %w", ErrGeneratePricingMap, err)
		}
		c.NextScrape = time.Now().Add(c.ScrapeInterval)
	}

	wg := sync.WaitGroup{}
	wg.Add(len(c.Regions))
	instanceCh := make(chan []ec2Types.Reservation, len(c.Regions))
	for _, region := range c.Regions {
		go func(region ec2Types.Region) {
			defer wg.Done()
			client := c.ec2RegionClient[*region.RegionName]
			reservations, err := ListComputeInstances(context.Background(), client)
			if err != nil {
				log.Printf("error listing instances: %s", err)
				return
			}
			log.Printf("found %d instances in region %s", len(reservations), *region.RegionName)
			instanceCh <- reservations
		}(region)
	}
	go func() {
		wg.Wait()
		close(instanceCh)
	}()
	c.emitMetricsFromChannel(instanceCh, ch)
	return nil
}
```

So I'd say its worth doing!

logyball: I hacked this together, and added the 5 goroutine limit to get around concurrent request rate-limiting, I imagine there's a better way to do this.  But it does seem like these API calls are expensive and should be parallelized

pkg/aws/eks/pricing_map.go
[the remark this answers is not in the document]
logyball: 👍

logyball (changes requested):
Real quick pushback on the parallelization - everything else looks good!

logyball (approved):
up to you with the parallelization (as discussed via slack) - nothing else blocking :letsfuggingo:

Issue #173, opened 2024-05-16
AWS S3 Costs only pull from single profile


Pokom:
The S3 module was developed in such a way that restricts us to only pulling cost data from a single profile. There should be a way for `cloudcost-exporter` to pull cost data from multiple profiles and merge the results together in a way that exports cost metrics for many profiles. This has already been started in https://github.com/grafana/cloudcost-exporter/pull/169/files#diff-6010e30d87bcb66ac6445f4924d0b0df711747cfc91710fea379a591f4ff28f3R134-R160

--- 

Goals
- [ ] Update `s3` to accept a set of profiles
- [ ] Run costexplorer query against each profile
- [ ] Emit cost metrics for multiple profiles

Pokom:
@Pokom will drop my notes here and then likely pivot to deployment per profile.

nikimanoledaki:
👋 This was partially solved by https://github.com/grafana/cloudcost-exporter/pull/464 🥳

nikimanoledaki:
@Pokom is it fair to say that this can be marked as completed by https://github.com/grafana/cloudcost-exporter/pull/487/files?

PR #174, opened 2024-05-16
Fix/aws s3 collect multiple profiles

Pokom:
- fixes https://github.com/grafana/cloudcost-exporter/issues/173

Cherry picked a change from https://github.com/grafana/cloudcost-exporter/pull/169 to extend the configuration for
AWS to have profiles which represents the profiles you want to pill data
from.

Updated `aws/s3` to become aware of multiple profiles and create a new
costexplorer client per profile when fetching billing data.
Updated `s3.parseBillingData` to return a slice of outputs so that we
can merge them with other profiles before parsing out billing data.

### DIFF
--- a/pkg/aws/aws.go
+++ b/pkg/aws/aws.go
@@ -118,7 +118,7 @@ func New(config *Config) (*AWS, error) {
 			}
 
 			client := costexplorer.NewFromConfig(ac)
-			collector, err := s3.New(config.ScrapeInterval, client)
+			collector, err := s3.New(config.ScrapeInterval, client, config.Profiles, config.Region)
 			if err != nil {
 				return nil, fmt.Errorf("error creating s3 collector: %w", err)
 			}
--- a/pkg/aws/aws.go
+++ b/pkg/aws/aws.go
@@ -20,6 +20,7 @@ type Config struct {
 	Services       []string
 	Region         string
 	Profile        string
+	Profiles       []string
 	ScrapeInterval time.Duration
 }
 
--- a/pkg/aws/s3/s3.go
+++ b/pkg/aws/s3/s3.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
+	awsconfig "github.com/aws/aws-sdk-go-v2/config"
 	awscostexplorer "github.com/aws/aws-sdk-go-v2/service/costexplorer"
 	"github.com/aws/aws-sdk-go-v2/service/costexplorer/types"
 	"github.com/prometheus/client_golang/prometheus"
@@ -128,6 +129,8 @@ type Collector struct {
 	metrics     Metrics
 	billingData *BillingData
 	m           sync.Mutex
+	profiles    []string
+	region      string
 }
 
 // Describe is used to register the metrics with the Prometheus client
@@ -145,7 +148,7 @@ func (c *Collector) Collect(ch chan<- prometheus.Metric) error {
 }
 
 // New creates a new Collector with a client and scrape interval defined.
-func New(scrapeInterval time.Duration, client costexplorer.CostExplorer) (*Collector, error) {
+func New(scrapeInterval time.Duration, client costexplorer.CostExplorer, profiles []string, region string) (*Collector, error) {
 	return &Collector{
 		client:   client,
 		interval: scrapeInterval,
@@ -153,6 +156,8 @@ func New(scrapeInterval time.Duration, client costexplorer.CostExplorer) (*Colle
 		nextScrape: time.Now().Add(-scrapeInterval),
 		metrics:    NewMetrics(),
 		m:          sync.Mutex{},
+		profiles:   profiles,
+		region:     region,
 	}, nil
 }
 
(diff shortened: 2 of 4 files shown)


pkg/aws/aws.go
```diff
@@ -98,26 +102,30 @@ func New(config *Config) (*AWS, error) {
...
+				// I'm going to use the AWS SDK to handle this for me. If the user has provided a region and profile, it will use that.
+				// If not, it will use the EC2 instance metadata service to determine the region and credentials.
+				// This is the same logic that the AWS CLI uses, so it should be fine.
+				options := []func(*awsconfig.LoadOptions) error{awsconfig.WithEC2IMDSRegion()}
+				if config.Region != "" {
+					options = append(options, awsconfig.WithRegion(config.Region))
+				}
+				if config.Profile != "" {
+					options = append(options, awsconfig.WithSharedConfigProfile(profile))
+				}
```
inkel, line 119: I'm not sure what's going on here. You check here that the single `-profile` flag has some value, and if it does, then forget about that value and use the one from the iteration 🤔 

Basically the `-profile` value never gets used, and in fact, I'd argue than adding this flag adds confusion to the user, as now they have `-profile` and `-profiles`. It seems to me that the single case is the same as `-profiles` with just one value.

What am I missing?

inkel (commented):
First pass at review, I've found something that I'm not entirely sure I understand. Once this is addressed I'd gladly do another run.

pkg/aws/aws.go
Pokom: Pretty on point question. Mainly, naming is hard 😅 

It looks like we're not using the `-profile` flag in dev or production today:

```
 spec:
  ...
      containers:
        - args:
            - --provider=aws
            - --aws.services=s3
```

I think the initial intent of the check you see is when doing local development to figure out which profile we should target. In dev/production that should be handled by:

```go
options := []func(*awsconfig.LoadOptions) error{awsconfig.WithEC2IMDSRegion()}
```

Let me remove the `-profile` flag as it's not used anymore. I'll build a local image and deploy this into development and make sure it still works 😬

logyball: agreed here, it looks like this might've been a misstep.  It now appears that we can't even use the flags separately, they _must_ be used together, as specifying `-aws.profile` will not get used unless there's a profile from `-aws.profiles` as well, in which case it'll use the value(s) from `aws.profiles`, but _unless_ you supply `aws.profile`, it'll never use any config values 😕

logyball:
```suggestion
				options = append(options, awsconfig.WithSharedConfigProfile(profile))
```

I'd suggest something like this, perhaps with a conditional flag to look if `aws-profile` is defined first rather than a loop through Profiles?

Or better yet, I'd just eliminate the `aws.profiles` flag entirely and allow `aws.profile` to be specified multiple times OR eliminate `aws.profile` and allow `aws.profiles` to be one or more comma-seperated items.

logyball (changes requested):
Same question as inkel, with a couple of suggestions how to resolve

pkg/aws/aws.go
Pokom: I've opted to stick with `-aws.profiles` as a flag, as I think it makes it clear that multiple values can be passed in. @inkel @logyball I've pushed a commit that removes the reference, and appreciate the thoughts/suggestions!

`cloudcost-exporter` when deployed infers the profile and region from this line:

```golang
				options := []func(*awsconfig.LoadOptions) error{awsconfig.WithEC2IMDSRegion()}
```

I need to be a bit mindful when rolling this change out. If the `-profiles` isn't set, I've added logic to return an error. So I'll need to likely couple rolling out the wave with adding the `-aws.profiles` flag to the environment.

pkg/aws/aws.go
```diff
@@ -98,26 +101,31 @@ func New(config *Config) (*AWS, error) {
...
-			// 2. Running within an EC2 instance and the region and profile can be derived
-			// I'm going to use the AWS SDK to handle this for me. If the user has provided a region and profile, it will use that.
-			// If not, it will use the EC2 instance metadata service to determine the region and credentials.
-			// This is the same logic that the AWS CLI uses, so it should be fine.
-			options := []func(*awsconfig.LoadOptions) error{awsconfig.WithEC2IMDSRegion()}
-			if config.Region != "" {
-				options = append(options, awsconfig.WithRegion(config.Region))
+			var clients []costexplorer.CostExplorer
+			if len(config.Profiles) == 0 {
+				return nil, fmt.Errorf("no profiles provided")
```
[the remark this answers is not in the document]
logyball: Only problem is this doesn't quite work.  If you split an empty string,[ it results in a length one array](https://go.dev/play/p/IAJ7VCgjkZZ) 😡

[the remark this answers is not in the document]
logyball: Could we either check that:

- `cfg.Providers.AWS.Profiles` != "" when defined
- move the check to the loop:

```go
case "S3":
			var clients []costexplorer.CostExplorer
			for _, profile := range config.Profiles {
				if profile == "" {
					return nil, fmt.Errorf("invalid profile provided")
				}
				// There are two scenarios:
				// 1. Running locally, the user must pass in a region and profile to use
				// 2. Running within an EC2 instance and the region and profile can be derived
				// I'm going to use the AWS SDK to handle this for me. If the user has provided a region and profile, it will use that.
				// If not, it will use the EC2 instance metadata service to determine the region and credentials.
				// This is the same logic that the AWS CLI uses, so it should be fine.
				options := []func(*awsconfig.LoadOptions) error{awsconfig.WithEC2IMDSRegion()}
				if config.Region != "" {
					options = append(options, awsconfig.WithRegion(config.Region))
				}
				options = append(options, awsconfig.WithSharedConfigProfile(profile))
				ac, err := awsconfig.LoadDefaultConfig(context.Background(), options...)
				if err != nil {
					return nil, err
				}
				client := awscostexplorer.NewFromConfig(ac)
				clients = append(clients, client)
			}

			collector, err := s3.New(config.ScrapeInterval, clients)
			if err != nil {
				return nil, fmt.Errorf("error creating s3 collector: %w", err)
			}
			collectors = append(collectors, collector)```

File: pkg/aws/s3/s3.go [hunk 3]
```diff
@@ -171,24 +171,29 @@ func (c *Collector) Register(registry provider.Registry) error {
 	return nil
 }
 
-// Collect is the function that will be called by the Prometheus client anytime a scrape is performed.
+// CollectMetrics Collect is the function that will be called by the Prometheus client anytime a scrape is performed.
```
logyball, line 174: Note, I have not tested this at length, but anecdotally it has improved the performance of collecting multiple profiles:

```go
// CollectMetrics Collect is the function that will be called by the Prometheus client anytime a scrape is performed.
func (c *Collector) CollectMetrics(ch chan<- prometheus.Metric) float64 {
	c.m.Lock()
	defer c.m.Unlock()
	now := time.Now()
	// :fire: Checking scrape interval is to _mitigate_ expensive API calls to the cost explorer API

	if c.billingData == nil || now.After(c.nextScrape) {
		eg := new(errgroup.Group)
		outputsLock := sync.Mutex{}
		var billingOutputs []*awscostexplorer.GetCostAndUsageOutput
		for _, client := range c.clients {
			eg.Go(func() error {
				endDate := time.Now().AddDate(0, 0, -1)
				// Current assumption is that we're going to pull 30 days worth of billing data
				startDate := endDate.AddDate(0, 0, -30)
				billingData, err := getBillingData(client, startDate, endDate, c.metrics)
				if err != nil {
					return err
				}
				outputsLock.Lock()
				billingOutputs = append(billingOutputs, billingData...)
				outputsLock.Unlock()
				return nil
			})
		}

		err := eg.Wait()
		if err != nil {
			log.Printf("Error getting billing data: %v\n", err)
			return 0
		}
		c.nextScrape = time.Now().Add(c.interval)
		c.metrics.NextScrapeGauge.Set(float64(c.nextScrape.Unix()))
		c.billingData = parseBillingData(billingOutputs)
	}

	exportMetrics(c.billingData, c.metrics)
	return 1.0
}
```

I can't speak to the validity of the results, but running it this way seemed to speed things up....

logyball (changes requested):
couple more suggestions, already liking this direction!!

pkg/aws/aws.go
inkel, line 106: (suggestion) We probably want to test this, and in order to make it simpler, I'd suggest adding a package variable `var ErrNoProfiles = errors.New("no profiles provided")` (and BTW, better to use `errors.New` instead of `fmt.Errorf` for simple strings). By adding this error then you can use `errors.Is` in the test.

inkel (commented):
I had some pending comments from last week. I'll do another review later on today.

Pokom:
Thanks @inkel and @logyball for the reviews!  Unfortunately this approach isn't going to work the way I had thought 😞 I have a strong preference not to create and maintain secrets to access multiple profiles, and the recommended route is to use roles and the [stscreds Assume Role](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/credentials/stscreds#hdr-Assume_Role) method. [Yet Another Cloudwatch Exporter](https://github.com/nerdswords/yet-another-cloudwatch-exporter/blob/c45e2404feca3feb5ca12f9dd2c8d03cb6944724/pkg/clients/v2/factory.go#L438) uses this technique to access multiple profiles from the same workload, but I still don't understand the underlying mechanism well enough. 

As a temporary workaround, I've deployed an instance of `cloudwatch-exporter` to EKS clusters in each profile needed, and will follow pattern for the short term.

PR #177, opened 2024-05-22
fix(gcp): Remove GB to GiB conversation on disk size

Pokom:
I believe the documentation for GCP's api is misleading and that we do not need to convert from GB to GiB when calculating the cost of a disk. This was observed by comparing the measured costs vs our bill over a period of a month and found that we were off by ~7%.

Removes the methods to convert disk.Size to GiB.

### DIFF
--- a/pkg/google/gke/README.md
+++ b/pkg/google/gke/README.md
@@ -34,7 +34,7 @@ For simplicity, `cloudcost-exporter` has implemented the following disk types:
 - Local SSD
 
 According to the [documentation](https://cloud.google.com/compute/disks-image-pricing#disk-and-image-pricing), pricing for storage is for [JEDEC Binary GB or IEC gibibytes(GiB)](https://en.wikipedia.org/wiki/Gigabyte).
-One of the more confusing bits is that the Documentation for disk implies that the size is in GB, but doesn't specify if it's in GB or GiB.
-The initial assumption was to convert the size to GiB, however after collecting data for a month and comparing against billing, we were off by just about ~7%.
-Coincidentally, the conversion from GB to GiB is ~7%, so we removed the code to convert from GB to GiB.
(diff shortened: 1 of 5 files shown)


pkg/google/gke/README.md
```diff
@@ -18,3 +18,23 @@ The primary motivation for this module was to ensure we could support the follow
...
+
+For simplicity, `cloudcost-exporter` has implemented the following disk types:
+- Standard(hard disk drives)
+- SSD(solid state drives)
+- Local SSD
+
+According to the [documentation](https://cloud.google.com/compute/disks-image-pricing#disk-and-image-pricing), pricing for storage is for [JEDEC Binary GB or IEC gibibytes(GiB)](https://en.wikipedia.org/wiki/Gigabyte).
+One of the more confusing bits is that the Documentation for disk implies that the size is in GB, but doesn't specify if it's in GB or GiB.
+The initial assumption was to convert the size to GiB, however after collecting data for a month and comparing against billing, we were off by just about ~7%.
+Coincidentally, the conversion from GB to GiB is ~7%, so we removed the code to convert from GB to GiB.
```
logyball, line 39: I might remove these and just go with the conclusion, if we're really curious we can read through the code history, but I don't imagine us re-reading this in a year and finding a lot of value from it.  This seems more like a PR description or a commit message than something to be preserved in documentation, in my opinion

logyball (approved):
LGTM! One small documentation NIT

pkg/google/gke/README.md
Pokom: Heh, yeah the docs actually started as a  bit of a way to keep breadcrumbs. Let me update with the conclusion 👍🏻

paulajulve (approved):

PR #180, opened 2024-05-30
fix(eks): Support Storage optimized and unkown instances

### LINKED ISSUE #179: AWS EKS: i4i.xlarge machine type reports CPU cost as 0
After deploying #169 I discovered that `i4i.xlarge` machine types cpu cost is being exported as 0 in `us-east-2`:
image

The price should be closer to $0.17/hour: https://instances.vantage.sh/aws/ec2/i4i.large

---

The root cause is because the `i4i` is part of the storage optimized family and we don't handle that. 
image

The root cause is from the lookup here: https://github.com/grafana/cloudcost-exporter/blob/ffce31a953d77aba4fe1f5bc6dc642e310d12a18/pkg/aws/eks/pricing_map.go#L147

We do not have any catch all/defaults in the cpu to memory map: https://github.com/grafana/cloudcost-exporter/blob/ffce31a953d77aba4fe1f5bc6dc642e310d12a18/pkg/aws/eks/eks.go#L32-L36

Pokom:
Adds unit tests to confirm a bug where an unknown instance type would return a value of 0 for the vCPU price.
Adds `Storage optimized` to the cpuToCostRatio map. Sets the default instance family to `General purpose` Moves over `cpuToCostRatio` to `pricing_map.go`.

- fixes #179


pkg/aws/eks/pricing_map.go
```diff
@@ -144,7 +155,12 @@ func weightedPriceForInstance(price float64, attributes Attributes) (*ComputePri
 	if err != nil {
 		return nil, fmt.Errorf("%w: %w", ErrParseAttributes, err)
 	}
-	ratio := cpuToCostRation[attributes.InstanceFamily]
+	ratio := cpuToCostRatio[attributes.InstanceFamily]
+	if ratio == 0 {
+		log.Printf("no ratio found for instance type %s, defaulting to %s", attributes.InstanceType, defaultInstanceFamily)
+		ratio = cpuToCostRatio[defaultInstanceFamily]
+	}
+
```
Pokom, line 163: @inkel is there a more idiomatic way of achieving this in Go? Plainly, what I want to do is if the instance family doesn't exist, I want to use a default key instead.

inkel: The more idiomatic way would be using the following:
```suggestion
	ratio, ok := cpuToCostRatio[attributes.InstanceFamily]
	if !ok {
		log.Printf("no ratio found for instance type %s, defaulting to %s", attributes.InstanceType, defaultInstanceFamily)
		ratio = cpuToCostRatio[defaultInstanceFamily]
	}

```

The reason why is because `0` is the zero value for `float64`, and you could have a element in the map whose value is `0` (whether that's a valid value or not depends on your use case), thus by using the _two-form access_ to the map, `ok` will return if the key was present or not in the map, ej:

```go
m := map[string]int{"zero": 0, "one": 1}
a, ok := m["zero"] // a == 0, ok == true
a, ok = m["cero"] // a == 0, ok == false
```

Basically is a way of checking that the returned value was found in the map and not the zero value of the element type.

inkel (approved):

Issue #32, opened 2023-12-06
Introduce slog with support for levels


Pokom:
There's many different logging patterns across the codebase with no easy path to refactoring them. It would be nice for the project to converge on using [slog](https://pkg.go.dev/log/slog) and provide a mechanism to pass a shared logging instance across all of the parts of the application. THis would provide:
1. Unified logging across the application
2. Easier to maintain and make changes across the codebase
3. Modifications can be made at the package level if necessary

This is the first problem, which is to introduce a module within the program that enables us to use structured logging and some mechanism to limit logs to specific levels.

Pokom:
Safe to call this complete with https://github.com/grafana/cloudcost-exporter/pull/197. We're working through the tailend now of updating existing providers + collectors with slog, but nothing is blocking that but time.

PR #188, opened 2024-06-03
feat(logger): Convert to using slog

Pokom:
This introduces using slog as a structured logger for the project. Created a `LevelHandler` struct that is responsible for only logging out messages below a configured log level.
Extended the config object to have a Log configuration to set
- Level
- Output
- Type

Added the following command flags:
- `log.level` (debug, **info**, warn, error)
- `log.output` (**stdout**, stderr)
- `log.type` (json, **text**)

Updated the csp and collector entrypoints to have accept a `*slog.Logger` parameter.
Started updating existing log.Printf statements to use appropriate logging.

TODO:
- [ ] Update all collectors to accept a logger parameter
- [ ] Determine best way to pass logger to helper methods that current log out via `log.Printf`

- relates #32

### DIFF
--- a/cmd/exporter/config/config.go
+++ b/cmd/exporter/config/config.go
@@ -29,4 +29,9 @@ type Config struct {
 		Path    string
 		Timeout time.Duration
 	}
+	Logger struct {
+		Level  string // debug, info, warn, error
+		Output string // stdout, stderr
+		Type   string // json, text
+	}
 }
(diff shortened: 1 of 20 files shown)


Pokom:
I'm going to close this PR in favor of breaking this up into smaller chunks. I'll keep the branch around temporarily as I use it for reference

PR #197, opened 2024-06-18
Feat/add leveled logger interface

Pokom:
Add base level log 
There are two main requirements for improving the logging of cloudcost-exporter:
1. Structured logging with consistent labels across providers and collectors
2. Ability to define a log level and runtime that limits the logs
       
There are no shortage of existing Go packages that could have been to meet the first requirement, such that:
- [logrus](https://pkg.go.dev/github.com/sirupsen/
- [zap](https://github.com/uber-go/zap)
- [zerolog](https://github.com/

However each of these comes with tradeoffs and then yet dependency to maintain. The go team released [slog](https://go.dev/blog/slog) officially in go 1.21, which provides a structured logging implemention as part of the standard library. This makes it a prime candidate to immediate meet the first requirement.
    
The second requirement(levels) is a bit more nuanced since the slog is  thin implementation. The go team provides a reference example to use: https://pkg.go.dev/log/slog@master#example-Handler-LevelHandler. 
This PR aims to meet both requirements by:
1. Adding `slog` as a common logging library to use
2. Creating a levels implementation that allows us to toggle the emitted level

    

There are two three methods added as well:
- `GetLogLevel` to convert a string to the log level
- `WriterForOutput` to convert a string to an `io.Writer`
- `HandlerForOutput` that takes a string and returns a slog handler for json or text

### DIFF
--- a/cmd/exporter/exporter.go
+++ b/cmd/exporter/exporter.go
@@ -6,6 +6,7 @@ import (
 	"flag"
 	"fmt"
 	"log"
+	"log/slog"
 	"net/http"
 	"os/signal"
 	"strings"
@@ -23,9 +24,38 @@ import (
 	"github.com/grafana/cloudcost-exporter/cmd/exporter/web"
 	"github.com/grafana/cloudcost-exporter/pkg/aws"
 	"github.com/grafana/cloudcost-exporter/pkg/google"
+	"github.com/grafana/cloudcost-exporter/pkg/logger"
 	"github.com/grafana/cloudcost-exporter/pkg/provider"
 )
 
+func main() {
+	var cfg config.Config
+	providerFlags(flag.CommandLine, &cfg)
+	operationalFlags(&cfg)
+	flag.Parse()
+
+	ctx, cancel := signal.NotifyContext(context.Background(), syscall.SIGINT, syscall.SIGTERM)
+	defer cancel()
+
+	logs := setupLogger(cfg.Logger.Level, cfg.Logger.Output, cfg.Logger.Type)
+	logs.LogAttrs(ctx, slog.LevelInfo, "Starting cloudcost-exporter",
+		slog.String("version", cversion.Info()),
+		slog.String("build_context", cversion.BuildContext()),
+	)
+
+	csp, err := selectProvider(&cfg)
+	if err != nil {
+		log.Fatalf("Error setting up provider %s: %s", cfg.Provider, err)
+	}
+
+	err = runServer(ctx, &cfg, csp, nil)
+	if err != nil {
+		log.Fatal(err)
+	}
+}
+
+// providerFlags is a helper method that is responsible for setting up the flags that are used to configure the provider.
+// TODO: This should probably be moved over to the config package.
 func providerFlags(fs *flag.FlagSet, cfg *config.Config) {
 	flag.StringVar(&cfg.Provider, "provider", "aws", "aws or gcp")
 	fs.StringVar(&cfg.Providers.AWS.Profile, "aws.profile", "", "AWS Profile to authenticate with.")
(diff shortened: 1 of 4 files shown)


cmd/exporter/exporter.go
```diff
@@ -23,9 +24,38 @@ import (
...
+		slog.String("version", cversion.Info()),
+		slog.String("build_context", cversion.BuildContext()),
+	)
+
+	csp, err := selectProvider(&cfg)
+	if err != nil {
+		log.Fatalf("Error setting up provider %s: %s", cfg.Provider, err)
+	}
+
+	err = runServer(ctx, &cfg, csp, nil)
```
logyball, line 51:
```suggestion
	err = runServer(ctx, &cfg, csp, logs)
```

cmd/exporter/exporter.go
```diff
@@ -100,13 +99,15 @@ func runServer(ctx context.Context, cfg *config.Config, csp provider.Provider) e
...
+		log.LogAttrs(ctx, slog.LevelInfo, "Starting server",
+			slog.String("address", cfg.Server.Address),
+			slog.String("path", cfg.Server.Path))
 		errChan <- server.ListenAndServe()
 	}()
 
 	select {
 	case <-ctx.Done():
-		log.Print("shutting down server")
+		log.LogAttrs(ctx, slog.LevelInfo, "Shutting down server")
```
logyball, line 110: oops - this should actually be `log`, but the arg should be passed into the `runServer` function (see above suggestion)

docs/contribute/logging.md
```diff
@@ -0,0 +1,22 @@
...
+2. Ability to define a log level and runtime that limits the logs emitted
+
+With `slog` being part of Go's stdlib since 1.21, we decided to use it as the logging library with a wrapper so that we can get log levels as well.
+
+## Guidelines
+
+1. Every provider _must_ accept a `*slog.Logger` in the constructor
+1. Every collector _must_ accept a `*slog.Logger` in the constructor
+1. Each provider and collect _must_ add a `provider` or `collector` group when initializing using the [slog.WithGroup](https://pkg.go.dev/golang.org/x/exp/slog#Logger.WithGroup) method
+1. Always prefer to use the `logger.WithAttr(...)` method to add structured data to the log message for both performance and consistency
```
logyball, line 17: Personally, I prefer things like `<loggingPackage>.Info(message, args...)`, it flows a lot more intuitively to me.

For example:

```go
		log.Info("Starting server",
			slog.String("address", cfg.Server.Address),
			slog.String("path", cfg.Server.Path))
```

just looks better than:

```go
		log.LogAttrs(ctx, slog.LevelInfo, "Starting server",
			slog.String("address", cfg.Server.Address),
			slog.String("path", cfg.Server.Path))
```

in my opinion.

logyball: if `ctx` is important (doesn't seem to show up in the logs, but I suppose could be used for things like timeouts, we could use `InfoCtx`

logyball (changes requested):
one blocking thing (crash) and a couple questions, lemme know what you think!

docs/contribute/logging.md
Pokom: So unfortunately this isn't just a semantic difference on looks. There's both a performance reason for using `logAttrs` and gaining access to context, which is simply not possible with using the pre built `log.(Debug|Info|Warn|Error`. See https://go.dev/blog/slog and search for Performance for a better overview on why.

pkg/logger/level.go
```diff
@@ -0,0 +1,89 @@
+package logger
```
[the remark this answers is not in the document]
Pokom: IMO `log.Fatal()` is better represented as:

```
logger.Attr(_, slog.LevelError, ...)
os.Exit(1)
```

The only place that you should be doing that is at the root of the application.

docs/contribute/logging.md
Pokom: Let me update that line to link out to the go blog post, as I feel this is going to be a somewhat controversial/highly opinionated topic 😅

pkg/logger/level.go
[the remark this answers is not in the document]
logyball: sounds good!

cmd/exporter/exporter.go
```diff
@@ -39,74 +74,50 @@ func providerFlags(fs *flag.FlagSet, cfg *config.Config) {
...
+		log.LogAttrs(ctx, slog.LevelInfo, "Starting server",
+			slog.String("address", cfg.Server.Address),
+			slog.String("path", cfg.Server.Path))
 		errChan <- server.ListenAndServe()
 	}()
 
 	select {
 	case <-ctx.Done():
-		log.Print("shutting down server")
+		log.Info("Context done", slog.String("test", "this"))
```
logyball, line 119:
```suggestion
```

logyball (commented):
one leftover testing line to remove, otherwise LGTM!

logyball (approved):

docs/contribute/logging.md
logyball: 👍

PR #198, opened 2024-06-21
POC: Azure Get Prices

logyball:
PoC of getting prices from Azure


scripts/azure-price-fetch/types.go
```diff
@@ -0,0 +1,17 @@
...
+	MachineType string
+}
+
+type VmMap struct {
+	RegionMap map[string]map[string]VirtualMachineInfo
+}
+
+type PriceMap struct {
+	RegionMap map[string]map[string]sdk.ResourceSKU
+}
```
Pokom, line 17: IMO having a dedicated `types` package is a bit of an antipattern from other languages and makes it harder to follow the code. I'd have these at the top of `azureClientWrapper.go` so one doesn't need to jump between multiple files. Often times I'll see a `types.go` like file for generated code :)

scripts/azure-price-fetch/main.go
```diff
@@ -0,0 +1,93 @@
...
+	debugSummary(priceMap, vmMap)
+}
+
+func debugSummary(priceMap *PriceMap, vmMap *VmMap) {
+	type Summary struct {
+		RegionName    string
+		TotalCost     float64
+		TotalMachines int
+		MachineTypes  map[string]bool
+	}
```
Pokom, line 63: Any reason for this to be scoped just to the method?

scripts/azure-price-fetch/main.go
```diff
@@ -0,0 +1,93 @@
...
+
+	for region, vmInformation := range vmMap.RegionMap {
+		fmt.Printf("Prices for region: %s\n", region)
+		totalCost := float64(0.0)
+		totalMachines := 0
+		machineTypes := map[string]bool{}
+
+		for vmName, vmInfo := range vmInformation {
+			vmType := vmInfo.MachineType
+			vmPrice := priceMap.RegionMap[region][vmType].RetailPrice
```
Pokom, line 74: There's a chance for a panic here if the region key doesn't exist. Consider doing 

```suggestion
if _, ok := priceMap.RegionMap[region]; !ok { continue }
```

scripts/azure-price-fetch/azureClientWrapper.go
```diff
@@ -0,0 +1,177 @@
...
+	return vmInfo, nil
+}
+
+func (a *AzureClientWrapper) getRegionalVmInformationFromRgVmss(ctx context.Context, rgMap map[string]string) (*VmMap, error) {
+	vmMap := VmMap{
+		RegionMap: make(map[string]map[string]VirtualMachineInfo),
+	}
+
+	for rg, location := range rgMap {
+		m, err := a.getVmInfoFromResourceGroup(ctx, rg)
```
Pokom, line 163: What does `m` represent in this case? Since it's not immediately obvious, consider renaming the variable to what it represents.

scripts/azure-price-fetch/azureClientWrapper.go
```diff
@@ -0,0 +1,177 @@
...
+
+	rgClient, err := armresources.NewResourceGroupsClient(subId, cred, nil)
+	if err != nil {
+		return nil, fmt.Errorf("failed to build resource group client: %w", err)
+	}
+
+	computeClientFactory, err := armcompute.NewClientFactory(subId, cred, nil)
+	if err != nil {
+		return nil, fmt.Errorf("failed to build compute client: %w", err)
+	}
```
Pokom, line 37: Very minor, but would likely get flagged once this is moved over to the `pkg` module, consider creating error objects for each scenario(or one for all three?) to make it easier to test each scenario. IE

```
var ErrClientCreationFailure = errors.New("failed to create client")
...
return nil, fmt.Printf("w: %w", ErrClientCreationFailure, err)
```

scripts/azure-price-fetch/azureClientWrapper.go
```diff
@@ -0,0 +1,177 @@
...
+	}, nil
+}
+
+func (a *AzureClientWrapper) buildQueryFilter(locationList []string) string {
+	locationListFilter := []string{}
+	for _, t := range locationList {
+		locationListFilter = append(locationListFilter, fmt.Sprintf("armRegionName eq '%s'", t))
+	}
+
+	locationListStr := strings.Join(locationListFilter, " or ")
```
Pokom, line 55: This is surprisingly elegant way of building the query. I do wonder if there's edge cases for us to worry about, such as if the list only has one item in it.

scripts/azure-price-fetch/azureClientWrapper.go
```diff
@@ -0,0 +1,177 @@
...
+	return fmt.Sprintf(`serviceName eq 'Virtual Machines' and priceType eq 'Consumption' and (%s)`, locationListStr)
+}
+
+func (a *AzureClientWrapper) getPrices(ctx context.Context, locationList []string) (*PriceMap, error) {
+	pm := PriceMap{
+		RegionMap: make(map[string]map[string]sdk.ResourceSKU),
+	}
+
+	pager := a.priceClient.NewListPager(&sdk.RetailPricesClientListOptions{
+		APIVersion:  to.StringPtr("2023-01-01-preview"),
```
Pokom, line 65: What's the significance of 2023-01-01-preview? I'd suggest moving it to a const and possibly link out to what it implies.

Pokom (commented):
Left some general comments around go style and what not, nothing would block. One thing I'd recommend so you don't have to follow up with it: You should leave comments on all exported methods and types. The comments are used for if we ever publish go docs, and also by IDE's for code generation. 

Let me move on to the correctness of what's being done.

scripts/azure-price-fetch/main.go
```diff
@@ -0,0 +1,93 @@
...
+			totalCost += vmPrice
+			totalMachines++
+			machineTypes[vmInfo.MachineType] = true
```
Pokom, line 77: Use the existing variable 😉 

```suggestion
			machineTypes[vmType] = true
```

scripts/azure-price-fetch/types.go
logyball: moved to the other file

scripts/azure-price-fetch/main.go
logyball: mostly just because it's only ever used here in a debug method ¯\_(ツ)_/¯

scripts/azure-price-fetch/main.go
logyball: updated

scripts/azure-price-fetch/azureClientWrapper.go
logyball: updated to `virtualMachineInfoByRg`

scripts/azure-price-fetch/azureClientWrapper.go
logyball: updated

scripts/azure-price-fetch/azureClientWrapper.go
logyball: added

scripts/azure-price-fetch/azureClientWrapper.go
logyball: fortunately no,  if I hardcode just a single location:

```go
func (a *AzurePriceInformationCollector) buildQueryFilter(locationList []string) string {
	locationList = []string{"uscentral"}
	locationListFilter := []string{}
	for _, region := range locationList {
		locationListFilter = append(locationListFilter, fmt.Sprintf("armRegionName eq '%s'", region))
	}

	locationListStr := strings.Join(locationListFilter, " or ")
	return fmt.Sprintf(`serviceName eq 'Virtual Machines' and priceType eq 'Consumption' and (%s)`, locationListStr)
}
```

We get the locationListStr as: `"armRegionName eq 'uscentral'"`

logyball:
closing in favor of #221

PR #199, opened 2024-06-24
Azure Provider Skeleton

logyball:
Adding the skeleton of the Azure Provider.

Note that this doesn't do much, just adds some building blocks and a test to ensure that we can instantiated it.

### DIFF
--- a/pkg/azure/azure.go
+++ b/pkg/azure/azure.go
@@ -1,0 +1,51 @@
+package azure
+
+import (
+	"context"
+	"log/slog"
+
+	"github.com/prometheus/client_golang/prometheus"
+
+	"github.com/grafana/cloudcost-exporter/pkg/provider"
+)
+
+const (
+	subsystem = "azure"
+)
+
+var (
+// TODO - add prometheus metrics here
+)
+
+type Azure struct {
+	Logger *slog.Logger
+}
+
+type Config struct {
+	Logger *slog.Logger
+}
+
+// New is a TODO
+func New(config *Config) (*Azure, error) {
+	providerGroup := config.Logger.With("provider", "azure")
+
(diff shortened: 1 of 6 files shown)


pkg/azure/azure.go
```diff
@@ -0,0 +1,55 @@
...
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+
+	"github.com/grafana/cloudcost-exporter/pkg/provider"
+)
+
+const (
+	// subsystem = "azure"
+	collectTimeout time.Duration = 5 * time.Minute
```
Pokom: I really like this notion, but don't know if this is appropriate at a per provider level. How do you feel about making this an app level configuration that we can then pass to other providers as well?

pkg/azure/azure.go
```diff
@@ -0,0 +1,55 @@
...
+	Logger  *slog.Logger
+}
+
+type Config struct {
+	Logger *slog.Logger
+}
+
+// New is a TODO
+func New(ctx context.Context, config *Config) (*Azure, error) {
+	providerGroup := config.Logger.WithGroup("azure")
```
Pokom, line 33: Suggestion: Make `azure` refer to the `subsystem` const that's been commented out

pkg/azure/azure.go
logyball: Good point, added it as a CLI arg and object on the top-level config

pkg/azure/azure.go
logyball: done

Pokom (approved):
Really appreciate chunking this out into a smaller PR. I think I'm going to borrow your `Collect` method creating a context with timeout to pass down to collectors!

Issue #200, opened 2024-06-24
[AWS]: Implement Compute module


Pokom:
There is a need for Grafana Labs to track the cost of ec2 instances and be able to attribute the costs to specific teams. This is very similar to what has already been implemented via the eks module, but with a few distinct differences:
1. Need to filter out instances associated with EKS clusters
2. Export the _total_ cost of the ec2 instances
3. Infer the team associated to the ec2 instance by looking for tags

Associating a team with an ec2 instance has many different ways of implementing. Since a majority of Grafana Labs infrastructure is provisioned as code, the easiest path would be to add a `team` and `workload` tag to each ec2 instance provisioned and then set the respective label to the value stored in the tag.

The follow metrics need to be emitted:
- `cloudcost_aws_ec2_instance_total_hourly_cost` 

The following labels must exist:
- instance
- region
- family
- machine_type
- price_tier
- team
- workload

Pokom:
@the-it posed the question on whether or not we even need to associate the the team label. There's a possibility yace already exposes this, and I've posed the question here: https://raintank-corp.slack.com/archives/C050DTQUXUP/p1719324648684559

Pokom:
Solved with https://github.com/grafana/cloudcost-exporter/pull/211

PR #202, opened 2024-06-24
feat(ec2): Add skeleton of ec2 module

Pokom:
This starts the process of refactoring `eks` and creating the skeleton for an ec2 module. Since `eks` and `ec2` are going to share a significan amount of code, I've introduced a `compute` module within `aws`. At a bare minimum the `pricing_map.go` will be shared, so that lives within the `compute` module.

Adds a base skeleton implementation for `ec2` that implements the collector interface and has 100% test coverage to start off.

Updates `aws.go` to handle an `ec2` flag, but does not document it's usage just yet.

First step of #200


pkg/aws/compute/ec2/ec2.go
```diff
@@ -0,0 +1,61 @@
...
+
+func (c *Collector) Describe(ch chan<- *prometheus.Desc) error {
+	return nil
+}
+
+func (c *Collector) Name() string {
+	return subsystem
+}
+
+func New(ps pricingClient.Pricing, ec2s ec2client.EC2, regions []ec2Types.Region, regionClientMap map[string]ec2client.EC2) *Collector {
```
logyball, line 50: Should we consider a CollectorConfig object here instead of a ton of args to the Constructor?

pkg/aws/compute/ec2/ec2.go
```diff
@@ -0,0 +1,61 @@
...
+	return &Collector{
+		pricingService:  ps,
+		ec2Client:       ec2s,
+		Regions:         regions,
+		ec2RegionClient: regionClientMap,
+	}
+}
+
+func (c *Collector) Register(_ provider.Registry) error {
+	return nil
```
logyball, line 60: [NIT] - not required, but I would add a logging statement here to indicate it was registered (and possibly in EKS as well?)

pkg/aws/compute/ec2/ec2.go
```diff
@@ -0,0 +1,61 @@
...
+func (c *Collector) Describe(ch chan<- *prometheus.Desc) error {
+	return nil
+}
+
+func (c *Collector) Name() string {
+	return subsystem
+}
+
+func New(ps pricingClient.Pricing, ec2s ec2client.EC2, regions []ec2Types.Region, regionClientMap map[string]ec2client.EC2) *Collector {
+	return &Collector{
```
logyball, line 51: You'll want to register a logger with the "ec2" Group here as well

pkg/aws/aws.go
```diff
@@ -142,6 +143,23 @@ func New(config *Config) (*AWS, error) {
 			}
 			collector := eks.New(config.Region, config.Profile, config.ScrapeInterval, pricingService, computeService, regions.Regions, regionClientMap)
 			collectors = append(collectors, collector)
+		case "EC2":
+			pricingService := pricing.NewFromConfig(ac)
+			computeService := ec2.NewFromConfig(ac)
```
logyball, line 148: Question: since these need not be independent of the `eks` collector, should we consider creating `pricingService` and `computeService` once and just passing them to the constructor for either `eks` or `ec2` ?

I realize that you could define one or the other, and not necessarily both, but perhaps initializing them outside the `switch` statement might then allow them to be defined just once.

logyball (changes requested):
Just a couple of questions, I like where this refactor is going

pkg/aws/aws.go
Pokom: I don't have a strong opinion. I was going back and forth on this and opted to create them twice. I opted for twice because there's no harm in them being created twice, but it feels a bit wrong to instantiate them if they're not going to be used.

pkg/aws/compute/ec2/ec2.go
Pokom: For now let me focus on getting the ec2 instance skeleton setup, then I can follow up with eks. Simply because I'll want to update all of EKS to use the newly available logger 😬

pkg/aws/compute/ec2/ec2.go
```diff
@@ -0,0 +1,78 @@
...
+	return nil
+}
+
+func (c *Collector) Name() string {
+	return subsystem
+}
+
+// New creates an AWS EC2 collector.
+func New(ctx context.Context, config *Config, ps pricingClient.Pricing, ec2s ec2client.EC2, regionClientMap map[string]ec2client.EC2) *Collector {
+	logger := config.Logger.With("collector", "ec2")
```
logyball, line 63: SUPER NIT - should this be 

```suggestion
	logger := config.Logger.With("collector", subsystem)

```

pkg/aws/compute/ec2/ec2.go
logyball: SGTM

logyball (approved):

pkg/aws/compute/ec2/ec2.go
Pokom: Only reason why I didn't is because then I think it becomes redundant. The logs would go from

```
time=2024-06-25T06:17:58.286-04:00 level=INFO msg="Registering AWS EC2 collector" provider=aws collector=ec2
```

to 

```
time=2024-06-25T06:18:33.250-04:00 level=INFO msg="Registering AWS EC2 collector" provider=aws collector=aws_ec2
```

Pokom: I think we'll need to rethink a bit the `sub_system` const and how it's being used. In EKS it's used for the metric name, which makes total sense: https://github.com/grafana/cloudcost-exporter/blob/7e496948ea16859fa5bd8bbcd45b911243a34344/pkg/aws/compute/eks/eks.go#L42

That translates to a metric named `cloudcost_aws_eks_instance_cpu_usd_per_core_hour` which aligns with our metric naming convention. In log world I think that's duplicative, but maybe that's fine.

PR #211, opened 2024-06-25
feat(ec2): Implement generating a pricing map

Pokom:
This is the next step towards implementing #200.
Updates logic in `pricing_map.go` to add an attribute for the total price of the instance.
Largely copies the implementation in `eks.go` to generate the pricing map by listing out OnDemand and Spot prices.
Refactors shared methods from `eks.go` to list on demand prices, spot prices, and instances to `pricing_map.go` and `compute.go` respectively. Create unit tests to ensure `ec2.Collect` has 100% coverage.

One thing that I'm pretty conscious of and don't have a great way of handling is how similar the `Collect` methods are up to this point. Right now I'm okay with duplicated code between the two as the remaining implementation will differ. I think long term there's an opportunity to refactor how we handle the pricing map that is shared between modules, but for now that's out of scope.


pkg/aws/compute/compute_test.go
```diff
@@ -0,0 +1,186 @@
...
+			},
+			want: "cluster-name",
+		},
+	}
+	for name, tt := range tests {
+		t.Run(name, func(t *testing.T) {
+			assert.Equalf(t, tt.want, ClusterNameFromInstance(tt.instance), "ClusterNameFromInstance(%v)", tt.instance)
+		})
+	}
+}
```
Pokom, line 186: This isn't new code, it's removed from `eks.go`.

pkg/aws/compute/compute.go
```diff
@@ -0,0 +1,47 @@
...
+func ClusterNameFromInstance(instance types.Instance) string {
+	for _, tag := range instance.Tags {
+		for _, key := range clusterTags {
+			if *tag.Key == key {
+				return *tag.Value
+			}
+		}
+	}
+	return ""
+}
```
Pokom, line 47: This is pulled from `eks.go` since it's now shared by both ec2 and eks.

pkg/aws/compute/ec2/ec2.go
```diff
@@ -45,7 +57,53 @@ func (c *Collector) CollectMetrics(_ chan<- prometheus.Metric) float64 {
...
+				priceList, err := compute.ListOnDemandPrices(context.TODO(), *region.RegionName, c.pricingService)
+				if err != nil {
+					return fmt.Errorf("%w: %w", compute.ErrListOnDemandPrices, err)
+				}
+
+				if c.ec2RegionClient[*region.RegionName] == nil {
+					return ErrClientNotFound
+				}
+				client := c.ec2RegionClient[*region.RegionName]
+				c.logger.LogAttrs(c.context, slog.LevelDebug, "Getting EC2 spot prices for region", slog.String("region", *region.RegionName))
```
logyball, line 81: NIT - same as above

```suggestion
				c.logger.LogAttrs(c.context, slog.LevelDebug, "Getting spot prices for region", slog.String("region", *region.RegionName))
```

logyball (commented):
LGTM, tested with a few calls, seems to be working correctly!

A few nit comments but nothing blocking. Nice work

logyball (approved):

PR #212, opened 2024-06-25
AKS Pricing Map Population

logyball:
Continuing on to implementing the AKS functionality.

This PR implements a "PriceStore" object used to hold the prices of different machine types based on region, priority, and operating system.  This is based on [Azure's pricing model](https://azure.microsoft.com/en-us/pricing/details/virtual-machines/windows/), where different prices are determined by a combination of those factors.

# Details

### Price Stratification

The PricingMap is built out with the following structure:

```
root -> {
  regionName -> {
    machinePriority -> {
      operatingSystem -> {
        skuName -> information
      }
    }
  }
}
```

That way, in order to uniquely identify a price, we will have to have the following attributes of any VM:
- the region it is deployed into
- it's priority (spot or on-demand)
- the operating system it is running
- it's SKU (e.g. `E8-4as_v4`)

Fortunately, when listing out VMs, these options are available to us.

### Population Strategy

When the PricingStore object is instantiated upon creating the AKS Collector, the price map is populated in the background.  Is is protected by the `lock`, so that if it is read before it is fully populated, the read will block until population is complete.

In addition, there is a region filter that can be used in combination with future work that will cause the price map to only populate for certain regions.  This is useful for the future, as we can only populate prices for locations where VMs are actually present.

### Future Work

- Tests (this should be covered by mocking in my opinion, once the metrics themselves are actually being populated)
- Caching VM IDs -> prices
  - This will help tremendously with lookup.  Once we have the VMSS and VMs listing functionality, we can store a price for individual machines, so they do not need to do an expensive lookup to find their prices 

# Testing

Currently there isn't a great way to test this without the debugger.  However, you can see the population happening in logging lines, but validating the actual numbers will have to come later:

```sh
❯ ./cloudcost-exporter --provider azure --server.address "127.0.0.1:8080" --azure.services "aks" --azure.subscription-id "<redacted>"          
time=2024-06-25T19:15:58.795-04:00 level=INFO msg="Starting cloudcost-exporter" version="(version=v0.1.5-3-g579964b-dirty, branch=logyball/aks-pricing-map-implementation, revision=579964b)" build_context="(go=go1.22.0, platform=darwin/arm64, user=loganballard@Logans-Laptop.local, date=2024-06-25T23:15:46Z, tags=unknown)"
time=2024-06-25T19:15:58.795-04:00 level=INFO msg="populating price map" provider=azure collector=aks subsystem=pricingMap
[...]

PR #213, opened 2024-06-26
AKS Virtual Machine Pricing Metric Population and Caching

logyball:
# AKS Virtual Machine Pricing Metric Population and Caching

This PR implements a working version of an _incomplete_ metric population for Azure AKS.  See the below "Todo" section for follow-up work.

A new method for obtain VM information is introduced.  In addition, a caching mechanism is implemented to allow repeated metric querying without repeated API calls.

## Virtual Machines

A new subsystem of the `aks` package is introduced: `machine_store`.  This is analogous to the already implemented `price_store`.  Rather than pricing information for certain machine types and characteristics, `machine_store` is responsible for:
- obtaining lists of VMs associated with AKS clusters in a given subscription
- obtaining relevant information associated with those VMs, e.g. Machine Type, Region, etc

### Population Strategy

Machines are populated in the following way:
- a list of AKS clusters and the associated resource group(s) of their nodes' Virtual Machines is obtained
-  For each resource group containing nodes associated with the cluster, information about VirtualMachineScaleSets (the backing resource of an AKS node pool) and the VMs that make up that scale set is obtained
- The VMs, along with their relevant information is stored in `MachineMap`

This information is refreshed periodically, [currently every five minutes.](https://github.com/grafana/cloudcost-exporter/blob/81933f82671646b14586a576858e8ce4876c5d8c/pkg/azure/aks/aks.go#L24)

## Caching

Because hitting the Azure API is:
- expensive (time)
- subject to rate-limiting (probably)

We'd like to hit it as few times as possible.  To avoid spamming the API, this PR also introduces a caching method.

### Caching Methodology

The AKS Collector implements a Cache of the following format:

```
{
  machineName -> price
}
```

This is exceedingly simple and SHOULD BE UPDATED TO BE ALL OF THE PROMETHEUS LABELS THAT ARE REQUIRED, @logyball , DUH.

# Putting it all Together

Now that it is established:
- how prices are obtained and saved in a `PriceMap` - [see previous PR for the gist of it](https://github.com/grafana/cloudcost-exporter/pull/212)
- how machines and their relevant details are obtained and saved in a `MachineMap`
- how that information is cached on a per-machine basis

We can combine the three to generate prometheus metrics.

When the AKS collector is first registered, both the PriceMap and the MachineMap are populated.  This can be expensive, as the _first_ time PriceMap is populated, it is un-aware what regions are relevant for the subscription, so it fetches all possible prices.  This is unfortunate.  

After MachineStore is populated, a regionList is stored composed of which regions have an AKS presence in this subscription.   This allows subsequent calls to populate PriceMap to be much more efficient.

After both PriceMap and MachineMap are populated, calls to `Collect()` (happening when the Prometheus `/metrics` endpoint is scraped) follow this pattern:

- Iterate over MachineMap 
  - for each machine in the map, get information from MachineMap
  - get pricing data from cache IF PRESENT
  - else get pricing information from PriceMap and Store in cache

Periodically, the cache is invalidated completely and must be built from scratch.

TODO - get all information from Cache

# Todos

- Grab the Machine "family somehow" (define what this means for Azure)
- Create a Cache that saves all of the relevant information
- Write tests
- If this PR is accepted, update documentation in the AKS module to reflect the design decisions
- Introduce methodology to approximate CPU and RAM cost


pkg/azure/aks/aks.go
```diff
@@ -27,19 +59,34 @@ var (
 
 // Prometheus Metrics
 var (
-// TODO - define Prometheus Metrics
+	InstanceCPUHourlyCostDesc = prometheus.NewDesc(
+		prometheus.BuildFQName(cloudcost_exporter.MetricPrefix, subsystem, "instance_cpu_usd_per_core_hour"),
```
Pokom, line 63: Nothing for your PR, just thinking out loud:

GKE/EKS/AKS all have the same suffix(`instance_cpu_us_per_cour_hour`) and help text. Would it be worthwhile to do either of the options?
1. Define a const `instance(Cpu|Ram)MetricSuffix`that's shared across all providers? 
2. A `GenerateDesc(prefix, subsystem, suffix string) *prometheus.NewDesc` method to keep things aligned?

pkg/azure/aks/aks.go
```diff
@@ -88,10 +132,115 @@ func (c *Collector) CollectMetrics(_ chan<- prometheus.Metric) float64 {
...
+	for vmName, vmInfo := range c.MachineStore.MachineMap {
+		price, err := c.PriceStore.getPriceInfoFromVmInfo(vmInfo)
+		if err != nil {
+			return err
+		}
+
+		c.MachinePricesCache[vmName] = price
+	}
+
+	return nil
```
Pokom, line 152: Suggestion, log out the duration it took to generate the machine map cache

pkg/azure/aks/aks.go
```diff
@@ -88,10 +132,115 @@ func (c *Collector) CollectMetrics(_ chan<- prometheus.Metric) float64 {
...
+}
+
+// TODO - BREAK INTO CPU AND RAM
+func (c *Collector) getMachinePrices(vmName string) (float64, error) {
+	// Cache Hit
+	c.cacheLock.RLock()
+	if price, ok := c.MachinePricesCache[vmName]; ok {
+		c.cacheLock.RUnlock()
```
Pokom, line 160: Any reason not to defer the `c.cacheLock.RUnlock()` call? I know there's a small performance penalty, but seems like it would read a bit cleaner.

pkg/azure/aks/aks.go
```diff
@@ -88,10 +132,115 @@ func (c *Collector) CollectMetrics(_ chan<- prometheus.Metric) float64 {
...
 // Collect satisfies the provider.Collector interface.
 func (c *Collector) Collect(ch chan<- prometheus.Metric) error {
-	// TODO - implement
-	c.logger.LogAttrs(c.context, slog.LevelInfo, "TODO - implement AKS collector Collect method")
+	c.logger.Info("collecting metrics")
+	now := time.Now()
+
+	eg := new(errgroup.Group)
+	if now.After(c.machineStoreNextPopulationTime) {
+		eg.Go(func() error {
```
Pokom, line 185: Suggestion: Log out that the machine store needs to be populated, and then log out the duration it took to populate.

pkg/azure/aks/aks.go
```diff
@@ -88,10 +132,115 @@ func (c *Collector) CollectMetrics(_ chan<- prometheus.Metric) float64 {
...
+			err := c.MachineStore.PopulateMachineStore()
+			if err != nil {
+				return err
+			}
+
+			c.machineStoreNextPopulationTime = time.Now().Add(machineRefreshInterval)
```
Pokom, line 191: Suggestion, log out the next population time stamp __or__ emit this as a metric that we can alert on.

pkg/azure/aks/aks.go
Pokom: Ignore the suggestion, you're doing that in the called method 🤦🏻

pkg/azure/aks/aks.go
```diff
@@ -88,10 +132,115 @@ func (c *Collector) CollectMetrics(_ chan<- prometheus.Metric) float64 {
...
+			}
+
+			c.priceStoreNextPopulationTime = time.Now().Add(priceRefreshInterval)
+			return nil
+		})
+	}
+
+	err := eg.Wait()
+	if err != nil {
+		return err
```
Pokom, line 211: Suggestion, return a custom error created so that we can validate the method returns a specific error during tests.

pkg/azure/aks/aks.go
```diff
@@ -88,10 +132,115 @@ func (c *Collector) CollectMetrics(_ chan<- prometheus.Metric) float64 {
...
+	for vmName, vmInfo := range c.MachineStore.MachineMap {
+		price, err := c.getMachinePrices(vmName)
+		if err != nil {
+			return err
+		}
+
+		labelValues := []string{
+			vmName,
+			vmInfo.Region,
+			"TODO - MACHINE FAMILY?",
```
Pokom, line 234: If this metric doesn't make sense for Azure, then don't worry about it . Our current set of metrics exports an `instance_type` that I believe aligns with `machine_type`. FWIW in GKE/EKS I don't know if there's much value in it. EKS makes it easy as they provide a value directly and it's like "Compute Optimized|Memory Optimized", wheres GKE it's inferred by the first few digits of the machine.

pkg/azure/aks/machine_store.go
```diff
@@ -0,0 +1,273 @@
...
+
+	m.machineMapLock.Lock()
+	defer m.machineMapLock.Unlock()
+
+	clusterList, err := m.getClustersInSubscription()
+	if err != nil {
+		return err
+	}
+
+	go m.setRegionListFromClusterList(clusterList)
```
Pokom, line 226: I don't quite follow why you'd want to run this in a goroutine and update the region within the method. It seems like a bit of a smell and has potential for race conditions. 

Suggestion:

```go
m.RegionList := regionListFromClusterList(clusterList)
```

Or don't run the method in a goroutine.

pkg/azure/aks/machine_store.go
```diff
@@ -0,0 +1,273 @@
...
+
+	// Clear the existing Map
+	m.MachineMap = make(map[string]*VirtualMachineInfo)
```
Pokom, line 229: There's a built in method for this now called [clear](https://tip.golang.org/ref/spec#Clear). I'd use that :)

pkg/azure/aks/aks.go
```diff
@@ -88,10 +132,115 @@ func (c *Collector) CollectMetrics(_ chan<- prometheus.Metric) float64 {
...
+	if err != nil {
+		return err
+	}
+
+	if now.After(c.cacheInvalidationTime) {
+		c.MachinePricesCache = make(map[string]float64)
+		c.cacheInvalidationTime = now.Add(cacheInvalidationInterval)
+		err = c.populateMachinePricesCache()
+		if err != nil {
+			return nil
```
Pokom, line 219: Think you meant to return error 😉 

Suggestion: Create a new error and return that so we can test it later.

pkg/azure/aks/aks.go
```diff
@@ -88,10 +132,115 @@ func (c *Collector) CollectMetrics(_ chan<- prometheus.Metric) float64 {
...
+		})
+	}
+
+	if now.After(c.priceStoreNextPopulationTime) {
+		eg.Go(func() error {
+			regionList := c.MachineStore.getRegionList()
+			err := c.PriceStore.PopulatePriceStore(regionList)
+			if err != nil {
+				return err
```
Pokom, line 201: Suggestion: Create a new error and return that so we can test it later.

pkg/azure/aks/aks.go
Pokom: Ignore this, the bottom level errors should be custom, this should simply return it.

pkg/azure/aks/aks.go
```diff
@@ -88,10 +132,115 @@ func (c *Collector) CollectMetrics(_ chan<- prometheus.Metric) float64 {
...
+			"TODO - MACHINE FAMILY?",
+			vmInfo.MachineTypeSku,
+			vmInfo.OwningCluster,
+			vmInfo.Priority.String(),
+		}
+		ch <- prometheus.MustNewConstMetric(InstanceCPUHourlyCostDesc, prometheus.GaugeValue, price, labelValues...)
+		ch <- prometheus.MustNewConstMetric(InstanceMemoryHourlyCostDesc, prometheus.GaugeValue, price, labelValues...)
+	}
+
+	c.logger.Info("metrics collected")
```
Pokom, line 243: Suggestion, include a duration for the time it took as well.

pkg/azure/aks/price_store.go
```diff
@@ -106,34 +94,17 @@ func (p *PriceStore) buildListOptions(locationList []string) *retailPriceSdk.Ret
...
+	p.logger.Info("populating price store")
 
-	p.lock.Lock()
-	defer p.lock.Unlock()
+	p.regionMapLock.Lock()
+	defer p.regionMapLock.Unlock()
 
-	pager := p.retailPriceClient.NewListPager(p.buildListOptions(locationList))
+	// clear the existing region map
+	p.RegionMap = make(map[string]PriceByPriority)
```
Pokom, line 105: Suggestion: Use the built in [clear](https://tip.golang.org/ref/spec#Clear) method

pkg/azure/aks/price_store.go
```diff
@@ -106,34 +94,17 @@ func (p *PriceStore) buildListOptions(locationList []string) *retailPriceSdk.Ret
...
 
+	pager := p.retailPriceClient.NewListPager((p.buildListOptions(locationList)))
```
Pokom, line 107: Linter is yelling here due to too many parenthesis:

```suggestion
	pager := p.retailPriceClient.NewListPager(p.buildListOptions(locationList))
```

pkg/azure/aks/price_store.go
```diff
@@ -165,23 +136,25 @@ func (p *PriceStore) PopulatePriceStore(locationList []string) error {
...
 
-// TODO - implement ability to lookup a certain VM's
-// Price by it's ID
-func (p *PriceStore) GetVmPrice() {}
-
-// TODO - use to grab regional prices
-// func (p *PriceStore) getPricesByRegion(region string) (*PriceByPriority, error) {
-// 	p.lock.RLock()
-// 	defer p.lock.RUnlock()
+func determineMachineOperatingSystem(sku retailPriceSdk.ResourceSKU) MachineOperatingSystem {
```
Pokom, line 143: nit: `getMachineOperationSystemFromSku`

pkg/azure/aks/price_store.go
```diff
@@ -165,23 +136,25 @@ func (p *PriceStore) PopulatePriceStore(locationList []string) error {
...
+	default:
+		return Linux
+	}
+}
 
-// 	priceByPriority, ok := p.RegionMap[region]
-// 	if !ok {
-// 		return nil, fmt.Errorf("region %s not found", region)
-// 	}
+func determineMachinePriority(sku retailPriceSdk.ResourceSKU) MachinePriority {
```
Pokom, line 152: nit: `getMachinepriorityFromSku`

pkg/azure/aks/aks.go
```diff
@@ -88,10 +132,115 @@ func (c *Collector) CollectMetrics(_ chan<- prometheus.Metric) float64 {
...
+		})
+	}
+
+	err := eg.Wait()
+	if err != nil {
+		return err
+	}
+
+	if now.After(c.cacheInvalidationTime) {
+		c.MachinePricesCache = make(map[string]float64)
```
Pokom, line 215: If the intent here is to clear it, use the built in [clear](https://tip.golang.org/ref/spec#Clear) method.

pkg/azure/aks/price_store.go
```diff
@@ -80,7 +54,21 @@ func NewPricingStore(subId string, priceClient *retailPriceSdk.RetailPricesClien
...
+	defer p.regionMapLock.RUnlock()
+
+	region := vmInfo.Region
+	priority := vmInfo.Priority
+	operatingSystem := vmInfo.OperatingSystem
+	sku := vmInfo.MachineTypeSku
+
+	vmPriceInfo := p.RegionMap[region][priority][operatingSystem][sku]
+
+	return vmPriceInfo.RetailPrice, nil
```
Pokom, line 71: I don't ever see a scenario where an error is returned. However there's quite a few places where a panic could be introduced. Suggestion: Check for the existence of keys, and return error if say:
1. region isn't found
2. priority class isn't found
3. operating system isn't found
4. sku isn't found

pkg/azure/aks/machine_store.go
```diff
@@ -0,0 +1,273 @@
...
+	vmInfo := m.MachineMap[vmName]
+
+	return &VirtualMachineInfo{
+		Name:            vmInfo.Name,
+		Region:          vmInfo.Region,
+		OwningVMSS:      vmInfo.OwningVMSS,
+		MachineTypeSku:  vmInfo.MachineTypeSku,
+		OperatingSystem: vmInfo.OperatingSystem,
+		Priority:        vmInfo.Priority,
+	}
```
Pokom, line 104: Is there any reason not to store this as the object in the `m.MachineMap`? It's fine to create it each time, but seems like you could simply this greatly if it's stored this way.

Pokom, line 104: There's a potential for panics here. You should check to see if `vmInfo` is nil and return an error.

pkg/azure/aks/aks.go
```diff
@@ -88,10 +132,115 @@ func (c *Collector) CollectMetrics(_ chan<- prometheus.Metric) float64 {
...
+	c.cacheLock.RUnlock()
+
+	// Cache Miss
+	c.cacheLock.Lock()
+	defer c.cacheLock.Unlock()
+
+	vmInfo := c.MachineStore.getVmInfoByVmName(vmName)
+	price, err := c.PriceStore.getPriceInfoFromVmInfo(vmInfo)
+	if err != nil {
+		return 0.0, err
```
Pokom, line 172: Suggestion: Return a custom error here

Pokom (commented):
I've left a considerable amount of comments, but that doesn't take away from the great progress on this PR! Very high level speaking, one thing I'd like to see is each method in machine_store and pricing_map that makes api calls _should_ accept a context object as the first method in the signature. This does two things:
1. decouples the Azure api calls from the stores
2. Allows us to pass in contexts with timeouts

We can either sync up next week once we're both in, or if you'd prefer I can submit a PR off of your branch with suggested changes. Otherwise it's too unwieldy for me to make the suggestions all at once

pkg/azure/aks/aks.go
Pokom: @the-it do you have any strong feelings about this since you helped craft the metric documentation?

logyball: https://github.com/grafana/cloudcost-exporter/issues/222 Created this issue to capture this thought

pkg/azure/aks/aks.go
logyball: done here https://github.com/grafana/cloudcost-exporter/blob/e6eaf0d940d828ed0035990628e45d0bce90359e/pkg/azure/aks/aks.go#L170

pkg/azure/aks/aks.go
logyball: removed for Azure

pkg/azure/aks/aks.go
logyball: done here https://github.com/grafana/cloudcost-exporter/blob/e6eaf0d940d828ed0035990628e45d0bce90359e/pkg/azure/aks/aks.go#L54

pkg/azure/aks/aks.go
logyball: done here https://github.com/grafana/cloudcost-exporter/blob/baa2ccf622ea91f1139cce8f462cab43fa542f3b/pkg/azure/aks/aks.go#L199

logyball:
superceded by release v0.3.4

PR #214, opened 2024-06-26
Feat/ec2 instance exporter

Pokom:
This PR ties together the https://github.com/grafana/cloudcost-exporter/pull/211 and lists out instances for a specific AWS profile. The code checks for instances that are not associated with a cluster and then emits a metric for the total hourly cost.

Adds documentation for how the collector is created and provides some basic PromQL queries that _could_ be used.


Pokom:
I'm going to close this in favor of #215. Can re-open if necessary

PR #215, opened 2024-06-27
feat(eks): Export total hourly cost for ec2 + eks instances

Pokom:
This removes the if conditione that filtered out ec2 instances that did not have a cluster name.
The rationale is that after implementing #214, it became apparent that there is no inherent value in having the two implementations split out. Prometheus will understand that an empty label is valid and be treated as if the label does not exist. This has been validated by emitting a set of production data and having it scraped into a Prometheus instance and submitted queries against it.

Adds a metric(`cloudcost_aws_eks_total_hourly_cost`) to emit for total hourly cost for each instance.

The plan for the PR is to will delete the implemented ec2 module and refactor the existing implementation to be called simply ec2. I think this is closer to reality and will allow us to be able to track the costs of ec2 instances associated with eks clusters and those that don't.


pkg/aws/ec2/ec2.go
```diff
@@ -32,30 +32,52 @@ var (
 var (
 	InstanceCPUHourlyCostDesc = prometheus.NewDesc(
 		prometheus.BuildFQName(cloudcostexporter.MetricPrefix, subsystem, "instance_cpu_usd_per_core_hour"),
-		"The cpu cost a compute instance in USD/(core*h)",
-		[]string{"instance", "region", "family", "machine_type", "cluster", "price_tier"},
+		"The cpu cost a ec2 instance in USD/(core*h)",
+		[]string{"instance", "Region", "family", "machine_type", "cluster", "price_tier"},
```
the-it, line 36: I think we must do `exported_cluster` for the cluster label. We use `cluster` otherwise at Grafana

the-it (changes requested):
this needs changes in metric documentation ... but aside that works for me. I don't necessary like the indirection, but I also see the benefit of having just 1 module.

pkg/aws/ec2/ec2.go
```diff
@@ -32,30 +32,52 @@ var (
 var (
 	InstanceCPUHourlyCostDesc = prometheus.NewDesc(
 		prometheus.BuildFQName(cloudcostexporter.MetricPrefix, subsystem, "instance_cpu_usd_per_core_hour"),
-		"The cpu cost a compute instance in USD/(core*h)",
-		[]string{"instance", "region", "family", "machine_type", "cluster", "price_tier"},
+		"The cpu cost a ec2 instance in USD/(core*h)",
```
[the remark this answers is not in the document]
Pokom: Good call, thanks!

pkg/aws/ec2/ec2.go
Pokom: Ahhh, let me align this with GKE and use `cluster_name`: https://github.com/grafana/cloudcost-exporter/blob/0c38ba570a79e04c74a8283012b10b40e88c17ca/pkg/google/gke/gke.go#L45

the-it (approved):
lgtm

PR #225, opened 2024-07-03
feat(gcp): Sloggify our gcp provider

Pokom:
This is a multipart refactor to sloggify the gcp provider and collectors. I'll break each commit up into implementing functionality for GCP as a provider then GCS, GKE, Compute.

### DIFF
--- a/pkg/google/gcp.go
+++ b/pkg/google/gcp.go
@@ -3,7 +3,7 @@ package google
 import (
 	"context"
 	"fmt"
-	"log"
+	"log/slog"
 	"strings"
 	"sync"
 	"time"
@@ -81,6 +81,7 @@ var (
 type GCP struct {
 	config     *Config
 	collectors []provider.Collector
+	logger     *slog.Logger
 }
 
 type Config struct {
@@ -90,6 +91,7 @@ type Config struct {
 	Services        []string
 	ScrapeInterval  time.Duration
 	DefaultDiscount int
+	Logger          *slog.Logger
 }
 
 // New is responsible for parsing out a configuration file and setting up the associated services that could be required.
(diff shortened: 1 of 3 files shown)


pkg/google/gcp.go
```diff
@@ -131,7 +136,9 @@ func New(config *Config) (*GCP, error) {
 				DefaultDiscount: config.DefaultDiscount,
 			}, cloudCatalogClient, regionsClient, storageClient)
 			if err != nil {
-				log.Printf("Error creating GCS collector: %s", err)
+				logger.LogAttrs(ctx, slog.LevelWarn, "Error creating collector",
```
logyball, line 139: Maybe should be:

```suggestion
				logger.LogAttrs(ctx, slog.LevelError, "Error creating collector",
```

?

pkg/google/gcp.go
```diff
@@ -145,15 +152,17 @@ func New(config *Config) (*GCP, error) {
 				ScrapeInterval: config.ScrapeInterval,
 			}, computeService, cloudCatalogClient)
 		default:
-			log.Printf("Unknown service %s", service)
-			// Continue to next service, no need to halt here
+
+			logger.LogAttrs(ctx, slog.LevelWarn, "Error creating service, does not exist",
```
logyball, line 156:
```suggestion
			logger.LogAttrs(ctx, slog.LevelError, "Error creating service, does not exist",
```

see above

logyball (approved):

pkg/google/gcp.go
Pokom: Ah yes, we should. That's definitely an error and not a warning

PR #226, opened 2024-07-05
Azure: adding ability to get CPU, Memory, disk size per machine

logyball:
Two changes:

### Add machine information
- Create a call to the `VirtualMachineSizes` client to get the different sizes of machines that are relevant for a particular subscription
- Add the Number of CPUs to a particular machines information
- Add the MB of RAM for a particular machine's information
- Add the size of the OS Boot Disk (not sure if this will every be relevant)
- Added a Machine Family classification system based on product SKU

### Index by Machine ID

The names of VMs as _nodes_ from a AKS cluster could not be globally unique.  VM IDs are.  Changing the identifier in the map to one that is globally unique.

### DIFF
--- a/pkg/azure/aks/machine_store.go
+++ b/pkg/azure/aks/machine_store.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
+	"strings"
 	"sync"
 	"time"
 
@@ -22,8 +23,26 @@ const (
 )
 
 var (
-	ErrMachineNotFound     = errors.New("machine not found in map")
-	ErrMachineTierNotFound = errors.New("machine tier not found in VMSS object")
+	ErrMachineNotFound       = errors.New("machine not found in map")
+	ErrMachineFamilyNotFound = errors.New("machine family not able to be determined by SKU")
+	ErrMachineTierNotFound   = errors.New("machine tier not found in VMSS object")
+
+	// As annoying as this is, I am unable to find an API call for this
+	// and performance of a map lookup will be quite faster
+	// than maintaining lists of each family
+	//
+	// Based on this logic https://learn.microsoft.com/en-us/azure/virtual-machines/vm-naming-conventions
+	MachineFamilyTypeMap map[byte]string = map[byte]string{
+		'A': "GeneralPurpose",
+		'B': "GeneralPurpose",
+		'D': "GeneralPurpose",
+		'F': "ComputeOptimized",
+		'E': "MemoryOptimized",
+		'M': "MemoryOptimized",
+		'L': "StorageOptimized",
+		'N': "GPUAccelerated",
+		'H': "HighPerformanceCompute",
+	}
 )
 
 type VirtualMachineInfo struct {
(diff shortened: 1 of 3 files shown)


pkg/azure/aks/machine_store.go
```diff
@@ -125,16 +141,44 @@ func (m *MachineStore) getVmInfoFromVmss(ctx context.Context, rgName, vmssName,
...
+					Id:              vmId,
+					Region:          vmRegion,
+					OwningVMSS:      vmssName,
+					OwningCluster:   cluster,
+					MachineTypeSku:  vmSku,
+					Priority:        priority,
+					OperatingSystem: osInfo,
+
+					NumOfCores:     to.Int32(vmSizeInfo.NumberOfCores),
+					MemoryInMB:     to.Int32(vmSizeInfo.MemoryInMB),
```
Pokom, line 174: Do you know if this is MB or MiB? I don't think it impacts this specific PR, but it's important to make sure it aligns with whatever the price is.

pkg/azure/aks/machine_store.go
```diff
@@ -125,16 +141,44 @@ func (m *MachineStore) getVmInfoFromVmss(ctx context.Context, rgName, vmssName,
...
+					OsDiskSizeInMB: to.Int32(vmSizeInfo.OSDiskSizeInMB),
+				}
+				m.logger.LogAttrs(ctx, slog.LevelDebug, "found machine information", slog.String("machineName", vmName))
+				continue
+			}
```
Pokom, line 179: Suggestion, keep the happy path outside of the if statement if possible. It's a bit confusing as I'm reviewing to see all of the other if statements handle error, then this one handles the "ok" path.

Pokom (commented):
Left one suggestion that I think will help improve readability. Otherwise 👍🏻

pkg/azure/aks/machine_store.go
logyball: After looking up the docs, it looks to be in MiB because the docs list computation in GiB?

https://learn.microsoft.com/en-us/azure/virtual-machines/sizes/memory-optimized/e-family

image

Good call out!

[The golang SDK docs, though... ](https://github.com/Azure/azure-sdk-for-go/blob/v12.5.0-beta/arm/compute/models.go#L2267) mention MB 😱 

Updated.

pkg/azure/aks/machine_store.go
logyball: updated! WDYT?

pkg/azure/aks/machine_store.go
```diff
@@ -184,27 +184,27 @@ func (m *MachineStore) getVmInfoFromVmss(ctx context.Context, rgName, vmssName,
...
-
-					NumOfCores:     to.Int32(vmSizeInfo.NumberOfCores),
-					MemoryInMiB:    to.Int32(vmSizeInfo.MemoryInMB),
-					OsDiskSizeInMB: to.Int32(vmSizeInfo.OSDiskSizeInMB),
-				}
-				m.logger.LogAttrs(ctx, slog.LevelDebug, "found machine information", slog.String("machineName", vmName))
-				continue
+			vmSizeInfo, ok := m.MachineSizeMap[vmRegion][vmSku]
+			if !ok {
+				m.logger.LogAttrs(ctx, slog.LevelDebug, "no VM sizing info found", slog.String("machineName", vmName))
```
Pokom, line 189: Missing the `continue` 😱

Pokom (commented):
Once the continue has been added, this is 👍🏻

Pokom (approved):

PR #237, opened 2024-07-09
initial stab into a readiness check

logyball:
Creating a readiness check, as price store population can take _forever_.

# Major Updates

- Added a new method to the Provider and collector interface `CheckReadiness()`
- added a new endpoint `/readyz` that checks the readiness probe for the Provider
- The readiness check essentially tries the read lock for all maps that we store information in, which will succeed if the metrics collectors are able to read the data they need and fail if not
- Updated s3's Mutex to a RWMutex

# Minor Updates

- added a Makefile recipe to generate mocks
- updated contributor docs to reflect azure testing
- updated mocks with new interface definition


cmd/exporter/exporter.go
```diff
@@ -102,8 +102,16 @@ func setupLogger(level string, output string, logtype string) *slog.Logger {
 func runServer(ctx context.Context, cfg *config.Config, csp provider.Provider, log *slog.Logger) error {
 	mux := http.NewServeMux()
 
-	mux.HandleFunc("/", web.HomePageHandler(cfg.Server.Path)) // landing page
-	registryHandler, err := createPromRegistryHandler(csp)    // prom metrics handler
+	mux.HandleFunc("/", web.HomePageHandler(cfg.Server.Path))               // landing page
+	mux.HandleFunc("/ready", func(w http.ResponseWriter, r *http.Request) { // readiness probe
```
Pokom, line 106: Let's align with k8s and use `readyz`: https://kubernetes.io/docs/reference/using-api/health-checks/#api-endpoints-for-health

pkg/aws/aws.go
```diff
@@ -227,6 +227,15 @@ func (a *AWS) Collect(ch chan<- prometheus.Metric) {
 	providerScrapesTotalCounter.WithLabelValues(subsystem).Inc()
 }
 
+func (a *AWS) CheckReadiness() bool {
+	for _, c := range a.collectors {
+		if !c.CheckReadiness() {
+			return false
+		}
```
Pokom, line 234: I wonder if this is better suited to run in parallel, especially given some may take quite some time. If one of them returns false, shortcircuit and return false. Is that something that's possible?

pkg/google/compute/pricing_map.go
```diff
@@ -108,6 +108,11 @@ func NewStructuredPricingMap() *StructuredPricingMap {
 	}
 }
 
+func (spm *StructuredPricingMap) CheckReadiness() bool {
+	// TODO - implement
+	return true
+}
+
```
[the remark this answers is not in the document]
logyball: Yup :/

pkg/aws/aws.go
logyball: I think we should probably keep it serial for now, if we implement more complicated readiness checks that checking a lock, this may be worth the overhead of a goroutine

cmd/exporter/exporter.go
logyball: updated

pkg/google/compute/pricing_map.go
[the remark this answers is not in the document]
Pokom: Let me create an issue and see if I can implement one this afternoon.

PR #240, opened 2024-07-12
Paula/add ebs pricing metrics

paulajulve:


pkg/aws/ec2/ec2.go
```diff
@@ -48,17 +49,25 @@ var (
 		[]string{"instance", "region", "family", "machine_type", "cluster_name", "price_tier"},
 		nil,
 	)
+	persistentVolumeHourlyCostDesc = prometheus.NewDesc(
+		prometheus.BuildFQName(cloudcostexporter.MetricPrefix, subsystem, "persistent_volume_usd_per_hour"),
+		"The cost of an AWS EBS Volume in USD.",
+		[]string{"persistent_volume", "region", "availability_zone", "disk", "type", "size_gib", "state"},
+		nil,
+	)
```
Pokom, line 57: We don't have a way to automate this, when you have a moment can you add an entry to the [ec2 metrics documentation](https://github.com/grafana/cloudcost-exporter/blob/main/docs/metrics/aws/ec2.md)?

pkg/aws/ec2/storage.go
```diff
@@ -0,0 +1,53 @@
...
+		if resp.NextToken == nil || *resp.NextToken == "" {
+			break
+		}
+	}
+
+	return volumes, nil
+}
+
+func NameFromVolume(volume types.Volume) string {
+	pv_name := "kubernetes.io/created-for/pv/name"
```
Pokom, line 44: Suggestion: make `pv_name` a const

pkg/aws/ec2/storage.go
```diff
@@ -0,0 +1,53 @@
...
+
+	for _, tag := range volume.Tags {
+		if *tag.Key == pv_name {
+			return *tag.Value
+		}
+	}
+	
+	return ""
+}
```
Pokom, line 53: This should be a fairly easy method to add tests for. Some scenarios
- volume being nil
- volume set but no matching tag
- volume set with a matching tag

pkg/aws/ec2/pricing_map.go
```diff
@@ -65,21 +66,42 @@ type Prices struct {
...
+		m:       sync.RWMutex{},
+	}
+}
+
+// GenerateComputePricingMap accepts a list of ondemand prices and a list of spot prices.
 // The method needs to
 // 1. Parse out the ondemand prices and generate a productTerm map for each instance type
 // 2. Parse out spot prices and use the productTerm map to generate a spot price map
-func (spm *StructuredPricingMap) GeneratePricingMap(ondemandPrices []string, spotPrices []ec2Types.SpotPrice) error {
+func (spm *ComputeStructuredPricingMap) GenerateComputePricingMap(ondemandPrices []string, spotPrices []ec2Types.SpotPrice) error {
```
Pokom, line 102: My IDE is yelling at me that the receivers(`spm`) for `*ComputeStructuredPricingMap` do not match. Can you update the recievers for the newly named `ComputeStructurePricingMap` to be `cspm`? Looks like it was updated in some but not all methods

pkg/aws/ec2/pricing_map.go
```diff
@@ -326,3 +390,46 @@ func ListSpotPrices(ctx context.Context, client ec2client.EC2) ([]ec2Types.SpotP
...
+		if err != nil {
+			return productOutputs, err
+		}
+
+		if products == nil {
+			break
+		}
+
+		productOutputs = append(productOutputs, products.PriceList...)
+		if products.NextToken == nil {
```
Pokom, line 429: Should we also check to see if the value is an empty string? This bits confused me quite a few times with aws responses, but I've had to also defensively check `if products.NextToken == nil || *products.NextToken == "" {...}`

pkg/aws/ec2/storage.go
paulajulve: New commit has those in place :sisi:

pkg/aws/ec2/storage.go
```diff
@@ -0,0 +1,53 @@
...
+		}
+
+		volumes = append(volumes, resp.Volumes...)
+		if resp.NextToken == nil || *resp.NextToken == "" {
+			break
+		}
+	}
+
+	return volumes, nil
+}
```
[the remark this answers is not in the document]
paulajulve: This one's tested now too.

pkg/aws/ec2/pricing_map.go
paulajulve: sure, but tell me how you got your IDE to yell at you for that, mine is 🤐

pkg/aws/ec2/pricing_map.go
paulajulve: didn't know why you were doing that. I can add it, doesn't hurt to be extra careful.

pkg/aws/ec2/pricing_map.go
Pokom: I'm using GoLand and I don't think I did anything special. When I'm looking at the code there's the squiggly line to inform me of a warning and when I hover over, the IDE provides this affordance: 
image


You can also open up the problems tab and see the warnings/errors for the file: 
image

pkg/aws/ec2/pricing_map.go
Pokom: I'm trying to remember if I actually experienced the error, or if it was when doing tests. I agree it doens't hurt to be extra careful, especially for a loop that makes API requests that can get stuck in an infinite loop 😬

pkg/aws/ec2/pricing_map.go
paulajulve: ah, yeah, JetBrains will do that for you ❤️  I'm on VSCode lately since I wasn't doing that much Go at all. Might need to find a plugin or reconsider my choices in life...

paulajulve: since I'm at it, what do you think about dropping the `structured` bit of this naming altogheter? I find that it makes for a long name and doesn't really add much information. It's a struct. We get that it'll be structured, no?

pkg/aws/ec2/pricing_map.go
paulajulve: Looking at `NextToken`'s type it's a `*string`. Since it's a pointer and not an actual string, it could never be an empty string, no? 🤔

paulajulve: done

pkg/aws/ec2/ec2.go
paulajulve: done. Marking this thread as resolved, but happy to hear feedback on the documentation added itself too.

pkg/aws/ec2/pricing_map.go
```diff
@@ -326,3 +390,46 @@ func ListSpotPrices(ctx context.Context, client ec2client.EC2) ([]ec2Types.SpotP
...
+			break
+		}
+		input.NextToken = products.NextToken
```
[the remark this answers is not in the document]
paulajulve: I haven't directly unit tested `getPricesFromProductsList`, because we get that for free from having both `ListStoragePrices` and `ListOnDemandPrices` tested. I could add a unit test, but I find that it doesn't add that much value, and I'd rather test the public methods instead.

As for adding it to the interface, I only added the other ones to the EC2 interface because I needed them for the compiler to stop screaming at me and to be able to test the new functionality. In this case, these methods are tested without the need to add them to the Pricing interface, so I don't see what the value would be in having them there too 🤔

pkg/aws/ec2/pricing_map.go
Pokom: I'm in favor of renaming. It made sense early on in the project(S3) and I don't think it does anymore. I'm fine to making it just `ComputePricingMap` and `StoragePricingMap` 👍🏻

paulajulve: done

paulajulve:
I've organised the changes into two more meaningful commits to make final review easier. First one is just a no-op refactor of the old code to prepare it to fit the new bits. Second one adds all the new functionality.

I think I've addressed all previous comments. This code is now ready to review 🎉

pkg/aws/ec2/ec2_test.go
```diff
@@ -379,3 +300,208 @@ func TestCollector_Collect(t *testing.T) {
...
+			close(volumesCh)
+		}()
+
+		wg2 := sync.WaitGroup{}
+		wg2.Add(1)
+		go func() {
+			defer wg2.Done()
+			collector.emitMetricsFromVolumesChannel(volumesCh, promCh)
+			wg2.Wait()
+		}()
```
paulajulve, line 498: WaitGroups were the only way I found to make these tests work. I know very little about channels and wait groups, but I was getting stuck on here, waiting for something on the channel, and the tests timed out. Happy to refactor/review with someone who can provide some insight into it.

pkg/aws/ec2/pricing_map.go
```diff
@@ -127,40 +150,71 @@ func (spm *StructuredPricingMap) GeneratePricingMap(ondemandPrices []string, spo
...
-// AddToPricingMap adds a price to the pricing map. The price is weighted based upon the instance type's CPU and RAM.
-func (spm *StructuredPricingMap) AddToPricingMap(price float64, attribute Attributes) error {
-	spm.m.Lock()
-	defer spm.m.Unlock()
-	if spm.Regions[attribute.Region] == nil {
-		spm.Regions[attribute.Region] = &FamilyPricing{}
-		spm.Regions[attribute.Region].Family = make(map[string]*Prices)
+// GenerateStoragePricingMap receives a json with all the prices of the available storage options
+// It iterates over the storage classes and parses the price for each one.
+func (spm *StoragePricingMap) GenerateStoragePricingMap(storagePrices []string) error {
```
logyball, line 155: At some point (probably at the start of this function), this should lock the Mutex on `StoragePricingMap`, unless it's already locked somewhere upstream and I missed that.

pkg/aws/ec2/pricing_map.go
```diff
@@ -127,40 +150,71 @@ func (spm *StructuredPricingMap) GeneratePricingMap(ondemandPrices []string, spo
...
+		if spm.Regions[region] == nil {
+			spm.Regions[region] = &StoragePricing{}
+			spm.Regions[region].Storage = make(map[string]float64)
+		}
+
+		for _, term := range productInfo.Terms.OnDemand {
+			for _, priceDimension := range term.PriceDimensions {
+				price, err := strconv.ParseFloat(priceDimension.PricePerUnit["USD"], 64)
+				if err != nil {
+					log.Printf("error parsing price: %s, skipping", err)
```
logyball, line 173: this is annoying, but my PR review is already going to drive you insane, so might as well throw one more in :)

`StoragePricingMap` should probably define it's own logger so we don't have to use `log.Printf`.  Since this is new code, it makes sense to do it right from the start.  

So something like:

```go 
type StoragePricingMap struct {
	// Regions is a map of region code to StoragePricing
	// key is the region
	// value is a map of storage classes to prices
	Regions map[string]*StoragePricing
	m       sync.RWMutex
	logger *slog.Logger
}

...

func NewStoragePricingMap(l *slog.Logger) *StoragePricingMap {
	return &StoragePricingMap{
		Regions: make(map[string]*StoragePricing),
		m:       sync.RWMutex{},
		logger:  l.With("subsystem", "storagePricing"),
	}
}
```

logyball (changes requested):
Ok I have to sign off for the day, but I think this should be a sizeable chunk... :) 

Main things:
- This is an awesome start.  All the building blocks are there, it just needs a little bit of tweaking
- The biggest thing that I noticed was that we should not be creating as many `context` objects, rather prefer to pass them into functions from their "parent", usually the EC2 or AWS Collector
- Current cost of volumes appears to be in USD/mo rather than USD/hr
- Implement locking when writing to the StorageMap
- Implement a logger on the StorageMap so we can avoid `log.Print`

pkg/aws/ec2/ec2.go
```diff
@@ -91,83 +101,170 @@ func (c *Collector) CollectMetrics(_ chan<- prometheus.Metric) float64 {
...
-				slog.Int("instances", len(reservations)),
-				slog.Duration("duration", time.Since(now)),
-			)
-			instanceCh <- reservations
-		}(region)
+		regionName := *region.RegionName
+		client := c.ec2RegionClients[regionName]
+
+		go c.fetchInstancesData(region, &wgInstances, instanceCh)
+		go c.fetchVolumesData(client, regionName, &wgVolumes, volumeCh)
```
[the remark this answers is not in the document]
paulajulve: I like this better. I assume this means I get rid of the `defer wg.Done()` inside of the functions, right? That's the change I've pushed.

pkg/aws/ec2/pricing_map.go
```diff
@@ -184,27 +238,47 @@ func weightedPriceForInstance(price float64, attributes Attributes) (*Prices, er
...
-	price := spm.Regions[region].Family[instanceType]
+	price := cpm.Regions[region].Family[instanceType]
 	if price == nil {
 		return nil, ErrInstanceTypeNotFound
 	}
-	return spm.Regions[region].Family[instanceType], nil
+	return cpm.Regions[region].Family[instanceType], nil
+}
+
+func (spm *StoragePricingMap) GetPriceForVolumeType(region string, volumeType string, size int32) (float64, error) {
```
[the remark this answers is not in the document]
paulajulve: yup, this was somewhat in the back of my mind when I dropped the `/h` from the docs, but completely forgot to kept that going to the actual metric calculation 😅

I'm thinking it's better to divide and get the hourly cost, than go for a monthly metric like they do, right? hours are the same length every time.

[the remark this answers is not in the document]
paulajulve: Went for hourly cost, since all metrics are hour based, and GKE module does hourly metrics for volumes as well.

And yes, the API gives back the prices with the unit `Gi/Mo`, like the docs, so we need to convert from monthly to hourly, and also yes, they use 30 days months in the docs. So you need to do some math one way or another, because they don't bill on 30 day periods.

pkg/aws/ec2/ec2.go
```diff
@@ -48,17 +49,25 @@ var (
...
-	NextScrape       time.Time
-	ec2RegionClients map[string]ec2client.EC2
-	logger           *slog.Logger
+	Regions                 []ec2Types.Region
+	ScrapeInterval          time.Duration
+	computePricingMap       *ComputePricingMap
+	storagePricingMap       *StoragePricingMap
+	pricingService          pricingClient.Pricing
+	ComputeScrapingInterval time.Time
+	StorageScrapingInterval time.Time
```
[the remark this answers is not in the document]
paulajulve: I like this idea, but it requires a serious overhaul of the existing tests for the compute pricing map, as those depend on these functions being synchronous atm 💩 I'll do that on a separate PR. This one's becoming unwieldy already. In the meantime, I can do the renaming.

pkg/aws/ec2/ec2_test.go
logyball: this one I'm also a bit stumped on...

logyball (changes requested):
Nice progress! I think with just a few touch-ups, we should be good here!

Note: I'm going to continue to noodle on the waitgroup test + channels closing and see if I can figure out something useful there.

pkg/aws/ec2/ec2_test.go
```diff
@@ -379,3 +300,197 @@ func TestCollector_Collect(t *testing.T) {
...
+
+		receivedMsg, ok := <-promCh
+		close(promCh)
+
+		assert.True(t, ok)
+		assert.NotNil(t, receivedMsg)
+		assert.Contains(t, receivedMsg.Desc().String(), "persistent_volume_usd_per_hour")
+	})
```
logyball, line 495: This will also work, not sure how much "better" it is:

```suggestion
		// fill volumes channel with data from the above volume
		go func() {
			volumesCh <- originMsg
			close(volumesCh)
		}()
		go func() {
			defer close(promCh)
			receivedMsg, ok := <-promCh

			assert.True(t, ok)
			assert.NotNil(t, receivedMsg)
			assert.Contains(t, receivedMsg.Desc().String(), "persistent_volume_usd_per_hour")
		}()

		collector.emitMetricsFromVolumesChannel(volumesCh, promCh)
```

logyball: It's not super `idiomatic` go, since the sender should always be closing the channel, and here' were saying that the promCh is receiving, but this is what we've got 🤷 

> Note: Only the sender should close a channel, never the receiver. Sending on a closed channel will cause a panic.

source: https://go.dev/tour/concurrency/4

logyball: Actually, you don't need to close `promCh`.  Updating suggestion.

pkg/aws/ec2/ec2_test.go
```diff
@@ -379,3 +300,197 @@ func TestCollector_Collect(t *testing.T) {
...
+			collector.emitMetricsFromVolumesChannel(volumesCh, promCh)
+			wg2.Wait()
+		}()
+
+		receivedMsg, ok := <-promCh
+		close(promCh)
+
+		assert.True(t, ok)
+		assert.NotNil(t, receivedMsg)
+		assert.Contains(t, receivedMsg.Desc().String(), "persistent_volume_usd_per_hour")
```
logyball, line 494: slightly more idiomatic maybe?

```suggestion
		wg := sync.WaitGroup{}
		wg.Add(2)
		var recMsg prometheus.Metric
		var ok bool

		// fill volumes channel with data from the above volume
		go func() {
			defer wg.Done()
			defer close(volumesCh)
			volumesCh <- originMsg
		}()

		go func() {
			defer wg.Done()
			recMsg, ok = <-promCh
		}()

		go collector.emitMetricsFromVolumesChannel(volumesCh, promCh)

		wg.Wait()

		assert.True(t, ok)
		assert.NotNil(t, recMsg)
		assert.Contains(t, recMsg.Desc().String(), "persistent_volume_usd_per_hour")
```

logyball: Basically, we need these three processes to happen in parallel:

- emitMetrics
- filling the `volumeCh`
- reading from the `promCh`

Essentially, `emitMetrics` blocks until the `volumeCh` has something it can read from, `promCh` is populated by `emitMetrics`, and `volumeCh` needs to _close_ after sending data so that `emitMetrics` knows there's no more values to loop over.

So:
- `emitMetrics` will only return when `volumeCh` is closed
- `recMsg` can only be evaluated after it has been read from `promCh` and _after_ emitMetrics returns

the three things done can actually be written in any order, `go` will magically make them do what we think.  We just need to have a `waitGroup` there to ensure that the thing that initiates the "termination" of the process (closing the `volumeCh`) blocks the `assert`, hence the WaitGroup.

logyball: We also don't _need_ to close the promCh here, but you can.

> Another note: Channels aren't like files; you don't usually need to close them. Closing is only necessary when the receiver must be told there are no more values coming, such as to terminate a range loop.

ref: https://go.dev/tour/concurrency/4

logyball: You can even do something like:

```go
                originMsgs := []ec2Types.Volume{
			{
				AvailabilityZone: aws.String(fmt.Sprintf("%sa", regionName)),
				VolumeId:         aws.String("vol-111111111"),
				VolumeType:       ec2Types.VolumeType(volumeType),
				Size:             aws.Int32(100),
			},
			{
				AvailabilityZone: aws.String(fmt.Sprintf("%sa", regionName)),
				VolumeId:         aws.String("vol-111111111"),
				VolumeType:       ec2Types.VolumeType(volumeType),
				Size:             aws.Int32(100),
			},
		}

		wg := sync.WaitGroup{}
		wg.Add(1)

		// fill volumes channel with data from the above volume
		go func() {
			defer wg.Done()
			defer close(volumesCh)
			volumesCh <- originMsgs
		}()
		go collector.emitMetricsFromVolumesChannel(volumesCh, promCh)

		wg.Wait()

		for i := 0; i < len(originMsgs); i++ {
			m, ok := <-promCh

			assert.True(t, ok)
			assert.NotNil(t, m)
			assert.Contains(t, m.Desc().String(), "persistent_volume_usd_per_hour")
		}
```

Note that if you loop over the promCh _the amount of expected times_, this will work.  But if you change it to looping over the channel itself:

```go
		for m := range promCh {
			// m, ok := <-promCh

			// assert.True(t, ok)
			assert.NotNil(t, m)
			assert.Contains(t, m.Desc().String(), "persistent_volume_usd_per_hour")
		}
```

you'll get into trouble, because that channel will never be closed, and the test will hang.

logyball: And now that I look at it even more, you can even just do this:

```go
		collector.storagePricingMap = &StoragePricingMap{
			Regions: map[string]*StoragePricing{
				regionName: {
					Storage: map[string]float64{
						volumeType: 0.1047,
					},
				},
			},
		}

		originMsgs := []ec2Types.Volume{
			{
				AvailabilityZone: aws.String(fmt.Sprintf("%sa", regionName)),
				VolumeId:         aws.String("vol-111111111"),
				VolumeType:       ec2Types.VolumeType(volumeType),
				Size:             aws.Int32(100),
			},
		}

		go collector.emitMetricsFromVolumesChannel(volumesCh, promCh)
		volumesCh <- originMsgs
		close(volumesCh)

		m, ok := <-promCh
		assert.True(t, ok)
		assert.NotNil(t, m)
		assert.Contains(t, m.Desc().String(), "persistent_volume_usd_per_hour")
```

All that's required is that the `emitMetrics` function is in the background such that it can receive on the `volumesCh`, and emit on the `promCh`.

`promCh` will block on receiving,  and closing the `volumesCh` indicates to `emitMetrics` that there's nothing more to loop over.

In fact, if you don't close the `volumesCh`, the test will still pass, but `emitMetrics` will have never returned (which I guess we don't care about for the purposes of this test)

pkg/aws/ec2/ec2.go
```diff
@@ -91,93 +102,190 @@ func (c *Collector) CollectMetrics(_ chan<- prometheus.Metric) float64 {
...
 		close(instanceCh)
 	}()
-	c.emitMetricsFromChannel(instanceCh, ch)
+	go func() {
+		wgVolumes.Wait()
+		close(volumeCh)
+	}()
+	c.emitMetricsFromReservationsChannel(instanceCh, ch)
+	c.emitMetricsFromVolumesChannel(volumeCh, ch)
 	c.logger.LogAttrs(context.TODO(), slog.LevelInfo, "Finished collect", slog.Duration("duration", time.Since(start)))
```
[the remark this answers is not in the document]
paulajulve: dammit! I'd looked for all the `context.Background()` ones, didn't know there was this option 😠

pkg/aws/ec2/pricing_map.go
```diff
@@ -327,3 +411,46 @@ func ListSpotPrices(ctx context.Context, client ec2client.EC2) ([]ec2Types.SpotP
...
 	return spotPrices, nil
 }
+
+func ListStoragePrices(ctx context.Context, region string, client pricingClient.Pricing) ([]string, error) {
+	input := &pricing.GetProductsInput{
+		ServiceCode: aws.String("AmazonEC2"),
+		Filters: []types.Filter{
+			{
+				Field: aws.String("regionCode"),
+				Type:  "TERM_MATCH",
```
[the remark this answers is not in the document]
paulajulve: ohhh... that'll solve [this](https://github.com/grafana/cloudcost-exporter/blob/5f9e474dbe535e11fe8a22f5386e72badfed9824/pkg/aws/ec2/pricing_map.go#L242) 😄 🎉

pkg/aws/ec2/pricing_map_test.go
```diff
@@ -337,6 +427,92 @@ func Test_weightedPriceForInstance(t *testing.T) {
...
+			ctx:    context.Background(),
+			region: "us-east-1",
+			err:    nil,
+			want: []string{
+				"This is definitely an accurate test",
+			},
+			GetProducts: func(ctx context.Context, input *pricing.GetProductsInput, optFns ...func(*pricing.Options)) (*pricing.GetProductsOutput, error) {
+				return &pricing.GetProductsOutput{
+					PriceList: []string{
+						"This is definitely an accurate test",
```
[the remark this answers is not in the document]
paulajulve: I know the commit log will show my name but this is not my doing! it was there when I got here! 😂

pkg/aws/ec2/ec2_test.go
```diff
@@ -26,92 +28,6 @@ var (
 	logger = slog.New(slog.NewTextHandler(os.Stdout, nil))
 )
 
-func TestCollector_ListOnDemandPrices(t *testing.T) {
```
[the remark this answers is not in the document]
paulajulve: I just moved it to the `pricing_map_test.go` file instead. That's a more appropriate place for it, since the function it's testing lives in `pricing_map.go`. The test is now called `Test_ListOnDemandPrices`, but has the exact same content.

pkg/aws/ec2/ec2_test.go
```diff
@@ -171,15 +87,29 @@ func TestCollector_Collect(t *testing.T) {
 	})
 
 	t.Run("Collect should return an error if ListOnDemandPrices returns an error", func(t *testing.T) {
```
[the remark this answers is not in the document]
paulajulve: Nope, we _are_ testing `ListOnDemandPrices`. These tests aren't super obvious. I had to add the spot prices block here because I changed the order in the function. So if I don't mock it, spots prices fails, and code never makes it into OnDemand prices. The error we're testing is the one on line 107. `ListOnDemandPrices` is supposed to get that error and bubble it up.

pkg/aws/ec2/ec2_test.go
```diff
@@ -271,7 +187,7 @@ func TestCollector_Collect(t *testing.T) {
 		defer close(ch)
 		assert.ErrorIs(t, collector.Collect(ch), ErrGeneratePricingMap)
 	})
-	t.Run("Collect should return an error if GeneratePricingMap returns an error", func(t *testing.T) {
+	t.Run("Collect should return an error if GenerateComputePricingMap returns an error", func(t *testing.T) {
```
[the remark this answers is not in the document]
paulajulve: ah, this seems to be the happy path. This tests already existed and I didn't read them, I just renamed the function. But this explains why there were two tests with the exact same name. Must've been a copy-paste gone wild in the past. I'll rename.

pkg/aws/ec2/ec2_test.go
```diff
@@ -188,13 +118,6 @@ func TestCollector_Collect(t *testing.T) {
 	})
 	t.Run("Collect should return a ClientNotFound Error if the ec2 client is nil", func(t *testing.T) {
 		ps := mockpricing.NewPricing(t)
-		ps.EXPECT().GetProducts(mock.Anything, mock.Anything, mock.Anything).
```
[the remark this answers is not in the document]
paulajulve: It's just that the test was failing because this expects the function to be called once, and now we're returning the error before the function is called. Basically, it yelled at me XD

paulajulve:
staged the new changes on image `v0.3.5-27-g2414266-dirty`, currently running on `dev-us-east-0`. Looks to be doing ok.

pkg/aws/ec2/ec2_test.go
```diff
@@ -379,3 +300,187 @@ func TestCollector_Collect(t *testing.T) {
...
+				AvailabilityZone: aws.String(fmt.Sprintf("%sa", regionName)),
+				VolumeId:         aws.String("vol-111111111"),
+				VolumeType:       ec2Types.VolumeType(volumeType),
+				Size:             aws.Int32(100),
+			},
+		}
+
+		go func() {
+			collector.emitMetricsFromVolumesChannel(volumesCh, promCh)
+		}()
```
logyball, line 473: last NIT and then I SWEAR I'll go away:

```suggestion
		go collector.emitMetricsFromVolumesChannel(volumesCh, promCh)
```

No need for an anonymous function here

logyball (approved):

pkg/aws/ec2/ec2_test.go
paulajulve: changed locally, push forced... forgot to commit 🤦‍♀️

PR #243, opened 2024-07-12
Azure: Update locking mechanism

logyball:
In some initial testing, we've discovered that the combination of the readiness probe, machine store population, and prometheus scrape can cause a deadlocking situation.  This PR does a couple things:

- creates a temporary band-aid fix that moves the locking responsibility from the AKS collector to the individual machine and price stores
- removes the TryLock method of readiness probes from all other providers, as it cannot be trusted at the moment.

### DIFF
--- a/pkg/azure/aks/aks.go
+++ b/pkg/azure/aks/aks.go
@@ -18,9 +18,6 @@ import (
 const (
 	subsystem             = "azure_aks"
 	AZ_API_VERSION string = "2023-01-01-preview" // using latest API Version https://learn.microsoft.com/en-us/rest/api/cost-management/retail-prices/azure-retail-prices
-
-	priceRefreshInterval   = 24 * time.Hour
-	machineRefreshInterval = 5 * time.Minute
 )
 
 type MachineOperatingSystem int
@@ -85,11 +82,9 @@ type Collector struct {
 	context context.Context
 	logger  *slog.Logger
 
-	PriceStore                   *PriceStore
-	priceStoreNextPopulationTime time.Time
+	PriceStore *PriceStore
 
-	MachineStore                   *MachineStore
-	machineStoreNextPopulationTime time.Time
+	MachineStore *MachineStore
 }
 
 type Config struct {
@@ -101,8 +96,6 @@ type Config struct {
 
 func New(ctx context.Context, cfg *Config) (*Collector, error) {
 	logger := cfg.Logger.With("collector", "aks")
-	now := time.Now()
-
 	priceStore, err := NewPricingStore(ctx, logger, cfg.SubscriptionId)
 	if err != nil {
 		return nil, err
@@ -117,11 +110,8 @@ func New(ctx context.Context, cfg *Config) (*Collector, error) {
 		context: ctx,
 		logger:  logger,
 
-		PriceStore:                   priceStore,
-		priceStoreNextPopulationTime: now.Add(priceRefreshInterval),
-
-		MachineStore:                   machineStore,
-		machineStoreNextPopulationTime: now.Add(machineRefreshInterval),
+		PriceStore:   priceStore,
+		MachineStore: machineStore,
 	}, nil
 }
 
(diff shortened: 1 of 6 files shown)


paulajulve (approved):

pkg/azure/aks/price_store.go
```diff
@@ -192,6 +199,17 @@ func (p *PriceStore) validateMachinePriceIsRelevantFromSku(ctx context.Context,
...
+		p.logger.LogAttrs(p.context, slog.LevelDebug, "not time to populate yet", slog.Time("nextPopulationTime", p.nextPopulationTime))
+		return nil
+	}
+
+	if p.currentlyPopulating {
+		p.logger.Info("another process is currently populating, aborting")
+		return nil
+	}
+
+	p.currentlyPopulating = true
```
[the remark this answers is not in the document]
logyball: Yup - I'm ditching this idea entirely.  I'll follow up with a more cohesive explination.

pkg/azure/aks/machine_store.go
```diff
@@ -362,7 +383,9 @@ func (m *MachineStore) PopulateMachineStore(ctx context.Context) error {
 	}
 
 	m.MachineMap = vmInfoMap
-	m.logger.LogAttrs(m.context, slog.LevelInfo, "machine store populated", slog.Duration("duration", time.Since(startTime)))
+	m.nextPopulationTime = time.Now().Add(machineRefreshInterval)
+	m.currentlyPopulating = false
```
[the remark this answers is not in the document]
logyball: Yeah, this approach is not fully thought out.  Replacing with something much more sensible.

pkg/azure/aks/aks.go
```diff
@@ -113,15 +107,25 @@ func New(ctx context.Context, cfg *Config) (*Collector, error) {
...
+	_, err = populationCron.AddFunc(fmt.Sprintf("@every %s", priceRefreshInterval), priceStore.PopulatePriceStore)
+	if err != nil {
+		return nil, err
+	}
+	_, err = populationCron.AddFunc(fmt.Sprintf("@every %s", machineRefreshInterval), machineStore.PopulateMachineStore)
+	if err != nil {
+		return nil, err
+	}
+
+	populationCron.Start()
```
Pokom, line 120: This isn't a _hard_ no, but I don't like the idea of adding a library for something that we can effectively handle with a ticker. See [go by example](https://gobyexample.com/tickers), I don't think it'd be that hard to emulate that with just tickers.

pkg/azure/aks/machine_store.go
```diff
@@ -358,12 +368,12 @@ func (m *MachineStore) PopulateMachineStore(ctx context.Context) error {
 
 	err = eg.Wait()
 	if err != nil {
-		return err
+		m.logger.Error("BIG ERROR YO")
```
Pokom, line 371: heh, love the intent, but let's update this error message with what the problem is.

Pokom (commented):
Push back if you feel strongly, but I'd prefer to not introduce a cron library when time.Ticker would handle the scenario just as well. Also check your IDE, there's a few unused variables(I can follow up with a golint config update to fail for those)

pkg/azure/aks/machine_store.go
logyball: Updated. The shame 🤦

pkg/azure/aks/aks.go
logyball: updated to just use tickers

logyball:
> Push back if you feel strongly, but I'd prefer to not introduce a cron library when time.Ticker would handle the scenario just as well. Also check your IDE, there's a few unused variables(I can follow up with a golint config update to fail for those)

Also updated the `golangci-lint` config file to check for unused variables.  WDYT?

pkg/azure/aks/machine_store.go
```diff
@@ -269,16 +265,31 @@ func (m *MachineStore) getMachineTypesByLocation(ctx context.Context, location s
...
+	for _, vmInfo := range m.MachineMap {
+		vmi = append(vmi, vmInfo)
+	}
+
+	return vmi, nil
+}
+
+func (m *MachineStore) PopulateMachineStore() {
 	startTime := time.Now()
+	ctx := context.TODO()
```
Pokom, line 286: nit/suggestion: You should pass this context into the method.

pkg/azure/aks/price_store.go
```diff
@@ -190,15 +186,16 @@ func (p *PriceStore) validateMachinePriceIsRelevantFromSku(ctx context.Context,
 	return true
 }
 
-func (p *PriceStore) PopulatePriceStore(ctx context.Context) error {
+func (p *PriceStore) PopulatePriceStore() {
 	startTime := time.Now()
-	p.logger.Info("populating price store")
+	ctx := context.TODO()
```
Pokom, line 191: Same nit as above, this context should still be passed in.

logyball: yup, good idea - done!

pkg/azure/aks/machine_store.go
logyball: yup, good idea - done!

Issue #201, opened 2024-06-24
Failing GCP Test - `TestGCP_CollectMetrics/two_collectors_with_no_errors`


logyball:
Test is failing on the main branch:

```shell
--- FAIL: TestGCP_CollectMetrics (0.00s)
    --- FAIL: TestGCP_CollectMetrics/two_collectors_with_no_errors (0.00s)
        gcp_test.go:206: 
                Error Trace:    /Users/loganballard/projects/cloudcost-exporter/pkg/google/gcp_test.go:206
                Error:          Not equal: 
                                expected: &utils.MetricResult{FqName:"cloudcost_exporter_collector_last_scrape_duration_seconds", Labels:utils.LabelMap{"collector":"test", "provider":"gcp"}, Value:0, MetricType:2}
                                actual  : &utils.MetricResult{FqName:"cloudcost_exporter_collector_last_scrape_error", Labels:utils.LabelMap{"collector":"test", "provider":"gcp"}, Value:0, MetricType:2}
                            
                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -1,3 +1,3 @@
                                 (*utils.MetricResult)({
                                - FqName: (string) (len=57) "cloudcost_exporter_collector_last_scrape_duration_seconds",
                                + FqName: (string) (len=46) "cloudcost_exporter_collector_last_scrape_error",
                                  Labels: (utils.LabelMap) (len=2) {
                Test:           TestGCP_CollectMetrics/two_collectors_with_no_errors
FAIL
FAIL    github.com/grafana/cloudcost-exporter/pkg/google        0.304s
```

@Pokom left a comment here

```go
				// TODO: @pokom need to figure out why _sometimes_ this fails if we set it to *.Times(tt.numCollectors)

```

PR #255, opened 2024-07-22
AZ: Add main interface tests 

logyball:
Adding some interface tests for the Azure collector using `gomock`.

Ninja add: updating `GetListOfVmsForSubscription` that never returns an error to simply return a machine list.

### DIFF
--- a/pkg/azure/aks/machine_store_test.go
+++ b/pkg/azure/aks/machine_store_test.go
@@ -15,6 +15,40 @@ func TestMachineStoreMapCreation(t *testing.T) {
 	t.Skip()
 }
 
+func TestGetListOfVmsForSubscription(t *testing.T) {
+	fakeMachineStore := &MachineStore{
+		MachineMap:     make(map[string]*VirtualMachineInfo),
+		machineMapLock: &sync.RWMutex{},
+	}
+	fakeMachineStore.MachineMap["vm1"] = &VirtualMachineInfo{
+		Name: "vm1",
+	}
+	fakeMachineStore.MachineMap["vm2"] = &VirtualMachineInfo{
+		Name: "vm2",
+	}
+
+	testTable := map[string]struct {
+		expectedNames []string
+	}{
+		"base_case": {
+			expectedNames: []string{"vm1", "vm2"},
+		},
+	}
+
+	for name, tc := range testTable {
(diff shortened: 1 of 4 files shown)


pkg/azure/azure_test.go
```diff
@@ -39,3 +43,90 @@ func Test_New(t *testing.T) {
...
+			assert.Equal(t, err, tc.expectedErr)
+		})
+	}
+}
+
+func Test_CollectMetrics(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+
+	fakePromCh := make(chan prometheus.Metric)
```
Pokom, line 97: Nit, don't really need to call this fake since it's an actual channel that's used for tested :)

pkg/azure/azure_test.go
```diff
@@ -39,3 +43,90 @@ func Test_New(t *testing.T) {
...
+			},
+		},
+	}
+
+	for name, tc := range testCases {
+		t.Run(name, func(t *testing.T) {
+			go func() {
+				for range fakePromCh {
+					// no metrics are generated here, so loop through to avoid
+					// process hang waiting on metrics that will never come
```
Pokom, line 114: Very small nit, put the comment before the for loop so we can get a nice tidy line:

`for range fakePromCh{}`

pkg/azure/aks/machine_store_test.go
```diff
@@ -15,6 +15,40 @@ func TestMachineStoreMapCreation(t *testing.T) {
...
+	for name, tc := range testTable {
+		t.Run(name, func(t *testing.T) {
+			vmList := fakeMachineStore.GetListOfVmsForSubscription()
+			var vmNameList []string
+
+			for _, v := range vmList {
+				vmNameList = append(vmNameList, v.Name)
+			}
+
+			assert.ElementsMatch(t, tc.expectedNames, vmNameList)
```
Pokom, line 47: TIL! This might help us with some flaky tests in #201

Pokom (approved):
Few nits but nothing blocking. 👍🏻

pkg/azure/azure_test.go
logyball: updated to just ch :D

PR #261, opened 2024-07-22
Experiment: Refactoring Azure SDK Client into Wrapper

logyball:
When attempting to write out larger-reaching integration tests for the Azure Module of `cloudcost-exporter`, I was running into a bit of a wall with how the mocks currently work.  This PR attempts to address that.

# What

Move our internal use of the Azure SDK into its own package.  The client(s) used will live in a Wrapper object that will be attached to the Azure object, which can be passed to individual Collectors for when they need to reach the Azure API.

# Why

By decoupling the logic of reaching out to the Azure API, we can _much_ more easily test the actual logic of the collectors. 

# Notes for reviewers

The git diff is a bit much.  One way to look at this is through several components

### Component One: Azure Client Wrapper

These changes are isolated to `pkg/azure/azureClientWrapper/azureClientWrapper.go`.  They are fairly straightforward and are lifting the logic from the individual components (e.g. MachineStore-> getVmInfoFromVmss becomes AzureClient->ListVirtualMachineScaleSetsOwnedVms).  

### Component Two: Machine Store

Machine store has taken the logic that reaches out to the Azure API and deferred that responsibility to AzureClient.  Notable places:

- `getVmInfoFromVmss` now uses `azClientWrapper.ListVirtualMachineScaleSetsOwnedVms`
- `getVMSSInfoFromResourceGroup ` now uses `azClientWrapper.ListVirtualMachineScaleSetsFromResourceGroup`
- `getClustersInSubscription` now uses `azClientWrapper.ListClustersInSubscription`

### Component Three: Price Store

- `PopulatePriceStore` now uses `p.azureClientWrapper.ListPrices`

### Component Four: Testing

Testing can be seen fully implemented for a single function (as a PoC) in `price_store_test.go -> TestPopulatePriceStore`.   

# Downsides (and counterpoints)

- This is a slightly different approach than the other two collectors, but I think it makes it slightly easier to think about from a testing perspective
- There will be a need to implement more stubs to the client wrapper as the functionality grows
  - This should not be challenging.  This functionality will already have to be implemented _somewhere_, might as well be the wrapper
- the Azure object must instantiate clients that it may not end up using, depending on which services are required
  - This is only a downside if the credentialing becomes a problem.  I do not think this will occur, as credentials are checked server-side upon making a _request_, so unused clients are more of an annoyance than a true problem.


paulajulve (commented):
LGTM, broadly. I don't love the naming of wrapper, I'd call that a service, but that's just a style choice. Before I :rubberstamp: tho, since the previously existing tests are not super extensive, have you staged these changes?

paulajulve (approved):
Currently staged and looks ok. Will defer to you to monitor until you feel confident merging. It's low risk anyway, since nobody is depending on these metrics yet.

Issue #262, opened 2024-07-22
compute: support using `node` instead/above of `instance` label


jjo:
## What

Either change compute modules' `instance` to `node`, or support e.g. `--compute.vm-instance-label=LABEL_NAME`.

## Why

Couple main reasons:

1)  When using `cloudcost-exporter` metrics in Kubernetes, to be able to join against `kube_.*` metrics (like `kube_node_labels`, `kube_node_annotations`, `kube_node_status_capacity`, `kube_node_status_allocatable`, ... see full list below), you then are forced to relabel `instance` to `node` with a construct like

```
  label_join( <CSP>_instance_cpu_usd_per_core_hour{}, "node", "", "exported_instance")
```
The above needs to be done _for every_ `<CSP>` you may run (especially the case if you have a centralized TSDB), note also that you need to visit _every_ timeseries to perform this relabelling (under the context of TSDB pressure).

2) Wisely, `cloudcost-exporter` already chose `cluster_name` instead of `cluster` (this being a popular _label_name_ choice, then avoiding the scrape-time rename to `exported_cluster`), then why keeping `instance` that _by design_ scraping agents will always rename to `exported_instance` (thus the example relabelling query above).

logyball:
I'm hesitant about replacing the `instance` label with `node`, as this couples our definition of "machine" with "machine provisioned by kubernetes".  Since we want to make this more agnostic of _how_ the machine was provisioned, I think that `node` wouldn't be the right choice here.

That being said, this _does_ make sense as an optional config argument, as it can then be used for convenience for a big (and possibly majority) use-case: managed Kubernetes.

I think this brings up an interesting point, since the `instance` label will often be replaced by Prometheus Scrapers and rewritten as `exported_instance` (as the instance of the _job_ running cloudcost-exporter is already using the `instance` label).  Thoughts on:

- config value for the label meaning "name of machine"?
- renaming the `instance` label to something agnostic of how it was provisioned, like `machine_name`?

jjo:
> I'm hesitant about replacing the `instance` label with `node`, as this couples our definition of "machine" with "machine provisioned by kubernetes". Since we want to make this more agnostic of _how_ the machine was provisioned, I think that `node` wouldn't be the right choice here.
> 
> That being said, this _does_ make sense as an optional config argument, as it can then be used for convenience for a big (and possibly majority) use-case: managed Kubernetes.
> 
[...]

PR #264, opened 2024-07-22
compute: use "node" instead of "instance" label name

jjo:
Ref: #262.

See #262 for the rationale, this PR implements straightfwd
`s/instance/node/`, alternatively we could consider this to be
optionally behind a flag, although increasing complexity for main to
CSP modules coupling and test cases.


jjo:
Closing as per discussion at #262.

Issue #263, opened 2024-07-22
compute: support labelling spot instance as boolean


jjo:
## What

Support e.g. `--compute.spot-as-boolean-label=LABEL_NAME` (e.g. with `_LABEL_NAME_=is_spot`), to set such _additional_ label.

## Why

As a _Spot_ instance is driven by "upper" level workloads' choice (when setting the nodeSelectors/taints/labels to trigger CA to allocate such specific nodes to them), with workloads explicitly knowing the rewards/risk aspects of spots. This is quite different from ondemand vs RIs (granted the latter are the only ones guaranteed to be available).

In the case of GrafanaLabs, this has grown in our recording rules to be used as `spot={true|false}`, which is now forcing us to use below relabeling trick (actually _add-labelling_) to set it:

```
(
  label_replace(<CSP>_instance_cpu_usd_per_core_hour{price_tier="spot"}, "spot", "true", "", "")
  or
  label_replace(<CSP>_instance_cpu_usd_per_core_hour{price_tier!="spot"}, "spot", "false", "", "")
)
```
The above needs to be done for every <CSP> you may run (especially the case if you have a centralized TSDB), note also that you need to visit every timeseries to perform this relabelling (under the context of TSDB pressure), create the _two_ TSs sets, to then `OR` them.

logyball:
This issue feels like a fine one to add to me.  It does expand the set of labels for compute instances by one per metric, but doesn't add cardinality, and will probably have conveniences down the line.  I don't even think we need this as a config, to be honest, we can just straight-up add a spot boolean where:

- ondemand and reserved instances set it to false
- spot instances set it to true

Pokom:
@jjo and I synced up on this yesterday and came to the agreement that the best thing to do here is to simply add a new label called `spot` that is a boolean value. This label is more of a relic of how we internally compute our TCO metrics, but there's potential value for others as well. 

The rationale for a new label is
1. Minimal overhead added to each time series
2. Reduced complexity by avoiding an operational toggle

jjo:
Closing after syncing with @the-it on fully embracing `price_tier` instead.

PR #266, opened 2024-07-23
compute: allow "instance" label name to be set (behind a flag)

### LINKED ISSUE #262: compute: support using `node` instead/above of `instance` label

jjo:
Fixes #262.

## What

Allow `instance` label name to be set via new `-compute.instance-label` flag.

## Why

See #262 for the rationale and discussion.

## How

* for each compute `Collector`, create new `collectorMetrics` type
  containing compute metrics (moved from _global var_), to let each
  `Collector` own an allocated instance obtained via
  `newCollectorMetrics(instanceLabel)`

* changing metrics from _global var_ to dynamically allocated object is
  needed for:
  * setting its label at runtime
  * allowing tests to be self-contained, else the 1st test to run would "own" setting `instanceLabel`
  * pave the way to also fixing #263, where we'd also need to (optionally) add a `spot={true,false}` label

* there's a new `config.CommonConfig` type, currently only having a
  single field `ComputeInstanceLabel`, which is passed down from `cmd/`
  to each instantiated _Provider_, then to each compute _Collector_

## Testing

Adapted testing with a couple `ComputeInstanceLabel="node"` changes,
although it should be improved to test changes to the label name.

--

### DIFF
--- a/cmd/exporter/config/config.go
+++ b/cmd/exporter/config/config.go
@@ -5,10 +5,15 @@ import (
 	"time"
 )
 
+type CommonConfig struct {
+	ComputeInstanceLabel string
+}
+
 type Config struct {
-	Provider  string
-	ProjectID string
-	Providers struct {
+	Provider     string
+	ProjectID    string
+	CommonConfig CommonConfig
+	Providers    struct {
 		AWS struct {
 			Profile  string
 			Region   string
--- a/cmd/exporter/exporter.go
+++ b/cmd/exporter/exporter.go
@@ -31,6 +31,7 @@ import (
 
 func main() {
 	var cfg config.Config
+	commonFlags(&cfg)
 	providerFlags(flag.CommandLine, &cfg)
 	operationalFlags(&cfg)
 	flag.Parse()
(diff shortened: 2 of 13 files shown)


jjo:
BTW also manually tested it locally with:

```
$ ./cloudcost-exporter  --aws.profile $DEV_PROFILE --aws.region us-east-1 --aws.services ec2 --compute.instance-label node
[...]

$ curl -s http://localhost:8080/metrics | egrep ^cloudcost_aws
[...]
cloudcost_aws_ec2_instance_total_usd_per_hour{cluster_name="<EDITED>",family="General purpose",machine_type="m6g.xlarge",node="ip-10-60-5-156.us-east-2.compute.internal",price_tier="ondemand",region="us-east-2"} 0.154
cloudcost_aws_ec2_instance_total_usd_per_hour{cluster_name="<EDITED>",family="General purpose",machine_type="m7g.2xlarge",node="ip-10-60-11-138.us-east-2.compute.internal",price_tier="spot",region="us-east-2a"} 0.1292
cloudcost_aws_ec2_instance_total_usd_per_hour{cluster_name="<EDITED>",family="General purpose",machine_type="m7g.2xlarge",node="ip-10-60-3-71.us-east-2.compute.internal",price_tier="spot",region="us-east-2a"} 0.1292
[...]

jjo:
Closing as this was intended as a PoC / exploration to sense the feasibility and complexity on adding such flag, then we also agreed on not moving fwd with this approach.

PR #270, opened 2024-07-30
AZ PoC Retry mechanism, making logging more uniform in Client Wrapper

logyball:
# Retry Mechaism

We've talked a bit about a retry mechanism for failed API calls.  I've implemented a PoC here from our very own @matryer 's  [try package](https://github.com/matryer/try/tree/master) to do just that.

# Improved logging

Previously the error logs and returned errors in the AZ Client wrapper were a bit of a mess, this works to clean them up.

# Price Map Clearing order

Changed the "clearing" of the price map to _after_ we've listed the API results.  This ensures that if we suffer an AZ outage (like we did today) where we are unable to list the prices, we will not wipe out the old data.  Since pricing data doesn't change often, this should be fairly safe to have "stale" data.


pkg/azure/azureClientWrapper/azureClientWrapper.go
```diff
@@ -70,9 +73,9 @@ func NewAzureClientWrapper(logger *slog.Logger, subscriptionId string, credentia
 	}, nil
 }
 
-var ErrPageAdvanceFailure = errors.New("failed to advance page")
-
 func (a *AzClientWrapper) ListVirtualMachineScaleSetsOwnedVms(ctx context.Context, rgName, vmssName string) ([]*armcompute.VirtualMachineScaleSetVM, error) {
+	listLogger := a.logger.With("pager", "listVirtualMachineScaleSetsOwnedVms")
```
Pokom, line 77: Just a nit, what's your reasoning for `listLogger`? I think `logger := ...` would be clear enough.

Pokom (approved):
I'm curious to see it @\matryer has any concerns around us using the package 😅 I'd be curious if he has a recommendation for another one, but I'm okay with it for now.

pkg/azure/azureClientWrapper/azureClientWrapper.go
logyball: no reason... just (as jjo says) E_NAMING_IS_HARD :)

I'll change to `logger`, it makes more sense.

PR #287, opened 2024-08-27
Standardise label as persistentvolume

paulajulve:
I've realised `cloudcost_aws_ec2_persistent_volume_usd_per_hour` is the only metric using this label as `persistent_volume` instead of `persistentvolume`.


Pokom (approved):
👍🏻 Nice catch!

PR #317, opened 2024-10-08
chore(GCP): Delete compute module

Pokom:
When initially building out the compute module, we had an idea that we wanted to keep the GKE/Compute module completely seperate. The primary motivation was because the labels would be different and thus we would want distinct metric names. Since then, Grafana has not moved the compute module into production and we have since implemented a similar feature for AWS. The main decision we made was to have the `cluster_name` label empty, which Prometheus will then discard.

This change removes the compute module, but does _not_ implement parity. We can follow up with that once there is a demand for us to develop a module specifically for GCE instances.


pkg/google/gke/gke.go
```diff
@@ -259,3 +263,31 @@ func ListDisks(project string, zone string, service *compute.Service) ([]*comput
 	}
 	return disks, nil
 }
+
+// ListInstancesInZone will list all instances in a given zone and return a slice of MachineSpecs
+func ListInstancesInZone(projectID, zone string, c *compute.Service) ([]*MachineSpec, error) {
+	var allInstances []*MachineSpec
+	var nextPageToken string
+	log.Printf("Listing instances for project %s in zone %s", projectID, zone)
```
logyball, line 271: I'd say either remove this logging statement or convert to `slog` to keep consistency

pkg/google/gke/gke.go
```diff
@@ -259,3 +263,31 @@ func ListDisks(project string, zone string, service *compute.Service) ([]*comput
...
+	now := time.Now()
+
+	for {
+		instances, err := c.Instances.List(projectID, zone).
+			PageToken(nextPageToken).
+			Do()
+		if err != nil {
+			log.Printf("Error listing instances in zone %s: %s", zone, err)
```
logyball, line 279: same as above

pkg/google/gke/gke.go
```diff
@@ -259,3 +263,31 @@ func ListDisks(project string, zone string, service *compute.Service) ([]*comput
...
+		}
+		for _, instance := range instances.Items {
+			allInstances = append(allInstances, NewMachineSpec(instance))
+		}
+		nextPageToken = instances.NextPageToken
+		if nextPageToken == "" {
+			break
+		}
+	}
+	log.Printf("Finished listing instances in zone %s in %s", zone, time.Since(now))
```
logyball, line 290: same as above

logyball (approved):
Approved w/ logging comments

PR #322, opened 2024-10-09
Add use_status label to GKE PVs

paulajulve:
Adding a `use_status` label to the `cloudcost_gcp_gke_persistent_volume_usd_per_hour`. 

It's an abstraction and a completely opinionated. Instead of adding just the information of who's using the disk, which is what GKE's API offers, what we want to be able to do is surface disks that are just sitting idle, not being used by anyone. We don't much care about what node/instance is actually doing the using, just whether it's being used or not.

I'm using the label name `use_status` because the API offers a `status` field which refers to a different thing. According to the API internal docs:
```// Status: [Output Only] The status of disk creation. - CREATING: Disk is
	// provisioning. - RESTORING: Source data is being copied into the disk. -
	// FAILED: Disk creation failed. - READY: Disk is ready for use. - DELETING:
	// Disk is deleting.
	//
	// Possible values:
	//   "CREATING" - Disk is provisioning
	//   "DELETING" - Disk is deleting.
	//   "FAILED" - Disk creation failed.
	//   "READY" - Disk is ready for use.
	//   "RESTORING" - Source data is being copied into the disk.
	//   "UNAVAILABLE" - Disk is currently unavailable and cannot be accessed,
	// attached or detached.
```

That's not what we're looking for here, but just to make sure we don't run into having overwritten this property in the future, I'm calling this label `use_status` instead of just `status`.


pkg/google/gke/disk.go
```diff
@@ -124,3 +126,11 @@ func (d Disk) DiskType() string {
 	}
 	return "persistent_volume"
 }
+
+func (d Disk) UseStatus() string {
+	if len(d.users) == 0 {
+		return "idle"
```
Pokom, line 132: Suggestion: turn `idle` and `in-use` into const variables that we can reference. This should help keep the tests and the code inline without much toil and refactors are simpler.

Pokom (commented):
From the code, everything looks good. Let's settle on the label name and then decide how to move forward. Great stuff!

Pokom:
> I'm using the label name use_status because the API offers a status field which refers to a different thing. According to the API internal docs:

For AWS we're using `state` as a label to communicate whether it's in use or not. I see four possibilities
1. align both with `state`
2. align both with `use_status`
3. let them not be aligned
4. find the one that 

@paulajulve do you have a strong preference?

paulajulve:
For AWS I used `state` because that's aligned with AWS's API. AWS calls this concept `state`. GCP's API on the other hand returns a list of users that can be empty, but does not return any field that would hold this idea of being in use vs not.

I briefly considered using `state` here too, since it wouldn't overwrite `status`, but I find it difficult to parse for future users why a metric might end up having both a `status` and a `state` labels 🤔

Pokom (approved):
@paulajulve and I discussed synchronously about the label name to use and agreed that the current one makes sense for _google_. We don't need the labels to align directly between CSP's as these can be changed after the fact with PromQL functions. 

The one thing that looks left to be done is to push the documentation around the states for a volume within GCP into the codebase. That wouldn't block, but would be nice to have.

PR #336, opened 2024-10-22
fix(gke): Update collect method to gracefully fail

### LINKED ISSUE #335: GKE stops exporting metrics if machine family isn't found in pricing map
If the machine family is not found when finding the price of a GKE instance, the program returns the error and does not process the remaining instances that are found: https://github.com/grafana/cloudcost-exporter/blob/c3031fdd0b27ff21244c24b852fbc502ae06d964/pkg/google/gke/gke.go#L159-L165

The expected behavior in this scenario is that machines without pricing should at least be skipped and we process the rest of the machines. There's a few possibilities on how to handle missing prices, but I believe that's best discussed in a separate issue as there are multiple options with pro's and cons. For now, let's stem the bleeding and ensure that metrics for existing prices are exported.

Pokom:
There is a condition where a failure to find a machine type in the price map was causing the `Collect` method to return instead of continuing. Instead, this changes the logic to log out the message within `Collect` and continue operating.
Adds a unit test to reproduce the error and then validates it is resolved.

A future improvement here would be to
1. Export a metric
2. Set a sane default value(if provided?)

- closes #335

### DIFF
--- a/pkg/google/gke/gke.go
+++ b/pkg/google/gke/gke.go
@@ -145,6 +145,13 @@ func (c *Collector) Collect(ch chan<- prometheus.Metric) error {
 				clusterName := instance.GetClusterName()
 				// We skip instances that do not have a clusterName because they are not associated with an GKE cluster
 				if clusterName == "" {
+					c.logger.LogAttrs(ctx,
+						slog.LevelDebug,
+						"instance does not have a clustername",
+						slog.String("region", instance.Region),
+						slog.String("machine_type", instance.MachineType),
+						slog.String("project", project),
+					)
 					continue
 				}
 				labelValues := []string{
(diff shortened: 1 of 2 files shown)


pkg/google/gke/gke.go
```diff
@@ -198,7 +214,15 @@ func (c *Collector) Collect(ch chan<- prometheus.Metric) error {
...
-					fmt.Printf("%s error getting cost of storage: %v\n", disk.Name, err)
+					c.logger.LogAttrs(ctx,
+						slog.LevelError,
+						err.Error(),
+						slog.String("disk_name", disk.Name),
+						slog.String("project", project),
+						slog.String("region", d.Region()),
+						slog.String("cluster_name", d.Cluster),
+						slog.String("storage_class", d.StorageClass()),
+					)
```
Pokom, line 225: This wasn't part of the issue linked in the PR, but it's good cleanup to convert the section to using slog.

jjo (approved):
LGTM!, thanks for the quick one :shipit:

PR #344, opened 2024-11-01
feat(gke): Handle hyperdisk balanced volumes

Pokom:
First attempt at handling hyperdisk balanced volumes. The primary goal was to update the pricing map logic in such a way that would parse of hyperdisks  balanced and include only the cost of capacity. Hyperdisk Balanced pricing is done in such a way that if you don't configure IOPS and Throughput, you use a free tier.

In the future IOPS, Throughput, and High Availability costs need to be saved and calculated as well.

### DIFF
--- a/pkg/google/gke/pricing_map.go
+++ b/pkg/google/gke/pricing_map.go
@@ -188,10 +188,11 @@ func (m PricingMap) GetCostOfStorage(region, storageClass string) (float64, erro
 
 var (
 	storageClasses = map[string]string{
-		"Storage PD Capacity":    "pd-standard",
-		"SSD backed PD Capacity": "pd-ssd",
-		"Balanced PD Capacity":   "pd-balanced",
-		"Extreme PD Capacity":    "pd-extreme",
+		"Storage PD Capacity":         "pd-standard",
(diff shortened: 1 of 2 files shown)


pkg/google/gke/pricing_map.go
```diff
@@ -272,6 +277,15 @@ func (pm *PricingMap) Populate(ctx context.Context, billingService *billingv1.Cl
 					log.Printf("Storage class %s already exists in region %s", storageClass, data.Region)
 					continue
 				}
+				// Switch statement must go here to handle hyperdisk cases, otherwise what's happening is
+				// The four dimenions get ignored. There is a sku for:
```
jjo, line 281:
```suggestion
				// The four dimensions get ignored. There is a sku for:
```

jjo (approved):
LGTM, suggest adding a _CAVEATS_ or _Experimental_ section to the main `README.md`, to highlight current state of hyperdisk support.

pkg/google/gke/pricing_map.go
Pokom: lol, I learned today that `golangci-lint` fails on spelling errors, thank you!

PR #350, opened 2024-11-04
chore(gke): Remove deprecated method

Pokom:
In #344, deprecated code was discovered which led to some confusion. It wasn't clear exactly how to remove the deprecated code and ensure tests were still functioning.
1. Rename `GeneratePricingMap` to `ParseSkus`
2. Update `ParseSkus` to be a receiver method on PricingMaps
3. Update `pricingMap.Populate` to return the result of ParseSkus
4. Update pricing map tests to call ParseSkus instead

### DIFF
--- a/pkg/google/gke/pricing_map.go
+++ b/pkg/google/gke/pricing_map.go
@@ -196,6 +196,8 @@ var (
 	}
 )
 
+// Populate is responsible for collecting skus related to Compute Engine, parsing out the response, and then populating the pricing map
+// with relevant skus.
 func (pm *PricingMap) Populate(ctx context.Context, billingService *billingv1.CloudCatalogClient) error {
 	serviceName, err := billing.GetServiceName(ctx, billingService, "Compute Engine")
 	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/google/gke/pricing_map.go
```diff
@@ -265,14 +272,14 @@ func (pm *PricingMap) Populate(ctx context.Context, billingService *billingv1.Cl
...
-					continue
-				}
 				if storageClass == "" {
 					log.Printf("Storage class not found for %s. Skipping", data.Description)
 					continue
 				}
+				if strings.Contains(data.Description, "Confidential") {
+					log.Printf("Storage class contains Confidential: %s\n%s\n", storageClass, data.Description)
+					continue
+				}
```
Pokom, line 282: I'm moving this after the first `storageClass` check since that one should take priority.

pkg/google/gke/pricing_map_test.go
```diff
@@ -644,13 +640,17 @@ func TestGeneratePricingMap(t *testing.T) {
 		},
 	} {
 		t.Run(tc.name, func(t *testing.T) {
-			pm, err := GeneratePricingMap(tc.skus)
+			pricingMap := &PricingMap{
+				Compute: make(map[string]*FamilyPricing),
+				Storage: make(map[string]*StoragePricing),
+			}
```
Pokom, line 646: Specifically not calling `NewPricingMap(...)` because that method will also call `populate`. I'm specifically trying to avoid having to mock out the billing client.

cindy (approved):
This change looks good to me, but I would be open to having someone else review it as well.

the-it (approved):
lgtm

PR #351, opened 2024-11-06
feat(gke): Implement StoragePricing to handle HyperDIsk Dimensions

Pokom:
First part of refactoring the GKE disks to account for HyperDisk pricing dimensions. Extends the storage pricing map to have a `StoragePrice` struct that has the costs for
- ProvisionedSpace
- IOPs
- Throughput

The first of many changes that's meant to incrementally add support.

Next steps
- Return the `StoragePrice` struct to pass down calculating cost to `disk.go`

### DIFF
--- a/pkg/google/gke/pricing_map.go
+++ b/pkg/google/gke/pricing_map.go
@@ -124,14 +124,6 @@ func NewPricingMap(ctx context.Context, billingService *billingv1.CloudCatalogCl
 	return pm, nil
 }
 
-// NewComputePricingMap returns a new PricingMap in a way that can be used afterwards.
-func NewComputePricingMap() *PricingMap {
-	return &PricingMap{
-		Compute: map[string]*FamilyPricing{},
-		Storage: map[string]*StoragePricing{},
-	}
-}
-
 // FamilyPricing is a map where the key is the family and the value is the price tiers
 type FamilyPricing struct {
 	Family map[string]*PriceTiers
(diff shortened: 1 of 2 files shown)


pkg/google/gke/pricing_map.go
```diff
@@ -124,14 +124,6 @@ func NewPricingMap(ctx context.Context, billingService *billingv1.CloudCatalogCl
...
 }
 
-// NewComputePricingMap returns a new PricingMap in a way that can be used afterwards.
-func NewComputePricingMap() *PricingMap {
-	return &PricingMap{
-		Compute: map[string]*FamilyPricing{},
-		Storage: map[string]*StoragePricing{},
-	}
-}
-
```
Pokom, line 134: This was no longer used after removing GCP compute module in https://github.com/grafana/cloudcost-exporter/pull/317

pkg/google/gke/pricing_map.go
```diff
@@ -143,14 +135,20 @@ func NewMachineTypePricing() *FamilyPricing {
 	}
 }
 
+type StoragePrices struct {
+	ProvisionedSpaceGiB float64
+	Throughput          float64
+	IOops               float64
```
Pokom, line 141: The names are subject to change once I investigate what the returned metadata is for disks. I'm also partial to `IOops` as it leans in on the "oops" bit

spinillos (approved):

pkg/google/gke/pricing_map.go
jjo: +1 I'm used to `IOps` in general, not fused on any of them fwiw.

jjo (approved):

Issue #18, opened 2023-11-30
Define the Versioning Strategy


Pokom:
While 0 is the new 1.0, we should have a well defined strategy on how we version and release `cloudcost-exporter`.

Define a release workflow in github actions.

nikimanoledaki:
Out of curiosity, what is the current release process?

Is it manually done through --> https://github.com/grafana/cloudcost-exporter/releases/new where a new tag is also created manually?

As for the cadence, it looks like it's weekly to bi-monthly. Could you share more on the reason for the patches, such as `0.4.x`, versus new minor versions?

Pokom:
@nikimanoledaki see https://github.com/grafana/cloudcost-exporter/blob/main/docs/contribute/releases.md for a very brief guide on the release process. For versioning, we try to stick with [semver](https://semver.org/) as best as we can, but it's always a judgement call. Fixes/dependencies end up being patches, whereas new features is a minor version update. Since we're still in alpha, we make no gaurantees of backwards compatibility so we've stuck with `0.x.x` until we're ready . 

> As for the cadence, it looks like it's weekly to bi-monthly.

Cadence is so far as we need it, but mostly aligning with PR's merged in or dependency updates.

nikimanoledaki:
Thank you, hadn't seen the release docs on this!

Happy to add a [goreleaser workflow](https://github.com/goreleaser/goreleaser-action) if you agree to this approach!

Pokom:
Great work @nikimanoledaki! From your perspective, is there anything else we should do on this issue? Or is having release notes automatically generated good enough for now?

nikimanoledaki:
We can close this issue and revisit the release process with more specific changes we'd like to make.

Ad hoc releases are fine for now IMO rather than a cadence - we don't need to introduce unnecessary processes but that could change in the next months.

We could introduce some stylistic changes to the changelog if we want. For example, the Terraform provider doesn't use the goreleaser's changelog but rather uses GitHub's own release note generation features: https://github.com/grafana/terraform-provider-grafana/releases/tag/v3.16.0 I think what we have is good for now - we can change this later if we want!

Pokom:
Closing with #359, and we can always iterate in the future. Thanks again @nikimanoledaki

PR #359, opened 2024-11-13
Add release GA workflow with goreleaser

nikimanoledaki:
Addresses https://github.com/grafana/cloudcost-exporter/issues/18

* Initialize goreleaser and its configuration
* Add a release GA workflow to create new releases when there are new tags
* Update docs about the release process

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,28 @@
+name: Release
+on:
+  push:
+    tags:
+      - v*
+
+permissions:
+  contents: read
+
+jobs:
+  goreleaser:
+    name: Release
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: Set up Go
+        uses: actions/setup-go@v5
+      - name: Run GoReleaser
+        uses: goreleaser/goreleaser-action@v6
+        with:
+          distribution: goreleaser
+          versions: '~> v2'
+          args: release --clean
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -13,13 +13,13 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
           fetch-depth: 0
       - name: Set up Go
-        uses: actions/setup-go@v5
+        uses: actions/setup-go@41dfa10bad2bb2ae585af6ee5bb4d7d973ad74ed # 5.1.0
       - name: Run GoReleaser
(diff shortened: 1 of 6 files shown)


.goreleaser.yaml
```diff
@@ -0,0 +1,27 @@
...
+    name_template: >-
+      {{ .ProjectName }}_
+      {{- title .Os }}_
+      {{- if eq .Arch "amd64" }}x86_64
+      {{- else }}{{ .Arch }}{{ end }}
+      {{- if .Arm }}v{{ .Arm }}{{ end }}
+checksum:
+  name_template: 'checksums.txt'
+changelog:
+  sort: asc
```
nikimanoledaki, line 27: @Pokom we can add some rules here to exclude `chore` or `docs` if you want. My thinking was that we could leave all PR titles for now and iterate later if we want to make the release notes leaner.

.github/workflows/release.yml
```diff
@@ -0,0 +1,28 @@
...
+permissions:
+  contents: read
+
+jobs:
+  goreleaser:
+    name: Release
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
```
spinillos, line 16: For security purposes, we should set the SHA instead the version: [docs](https://enghub.grafana-ops.net/docs/default/component/deployment-tools/platform/continuous-integration/#pin-actions-to-a-full-length-commit-sha). Its something that we need to update in other repos 😅. Here you are an [example](https://github.com/grafana/deployment_tools/blob/master/.github/actions/kube-manifests-exporter/conftest-full/action.yml#L34).

Pokom: TIL! Thanks for sharing that @spinillos. I agree, let's pin to a specific version here and we can clean up other instances in another PR

.goreleaser.yaml
Pokom: Agreed on `docs` and think that would be fine to add here, assuming it's a simple check. `chore` may be related to code changes, so I would hold off on excluding `chore` PR's

.github/workflows/release.yml
nikimanoledaki: I didn't know that, thank you! Changed it for the new actions, did the same for other actions that we use, and updated the docs :)

.goreleaser.yaml
nikimanoledaki: added exclusions for docs and specifically chore(deps) :)

docs/contribute/releases.md
```diff
@@ -1,19 +1,45 @@
...
 To cut a release, clone `cloudcost-exporter` locally and pull latest from main.
-Determine if the next release is a major, minor, or hotfix. 
+Determine if the next release is a major, minor, or hotfix.
 Then increment the relevant version label.
 
 For instance, let's say we're on `v0.2.2` and determined the next release is a minor change.
-The next version would then be `v0.3.0`. 
-Execute the follow command and generate the tag and push it:
+The next version would then be `v0.3.0`.
+Execute the follow command to generate the tag and push it:
```
Pokom, line 17:
```suggestion
Execute the following command to generate the tag and push it:
```

Pokom (approved):
One very minor suggestion, otherwise 👍🏻

PR #367, opened 2024-11-20
Add a backstage catalog info file

Pokom:
Adds an entrypoint for Backstage so that our internal instance can render the project.

### DIFF
--- a/catalog-info.yaml
+++ b/catalog-info.yaml
@@ -11,6 +11,6 @@ metadata:
   description: |
     Prometheus Exporter for Cloud Provider agnostic cost metrics
 spec:
-  type: tool
+  type: service
   owner: group:default/platform-monitoring
   lifecycle: production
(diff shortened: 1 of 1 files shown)


catalog-info.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: backstage.io/v1alpha1
+kind: Component
+metadata:
+  name: cloudcost-exporter
+  title: Cloudcost Exporter
+  annotations:
+    backstage.io/techdocs-ref: dir:./docs
```
K-Phoen, line 7: I think this should be `backstage.io/techdocs-ref: dir:.`, with a `mkdocs.yml` file at the root to configure the docs generation (like https://github.com/grafana/cog/blob/main/mkdocs.yml)

Also: I think there should be an index.md file in every folder under `docs/` for the docs to be generated properly (or just for the navigation?)

You should be able to preview the docs locally with `npx @techdocs/cli serve -v`

K-Phoen (approved):

PR #371, opened 2024-11-27
Fix build step for release workflow

nikimanoledaki:
2nd error from the failed [run](https://github.com/grafana/cloudcost-exporter/actions/runs/11934973233/job/33591750094):
```
error=build for cloudcost-exporter does not contain a main function
```

The path to `main()` is set, but it may have been trying to create multiple builds due to the list syntax. Let's try this way.

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -1,10 +1,11 @@
 # Make sure to check the documentation at https://goreleaser.com
 
+project_name: cloudcost-explorer
 builds:
   - id: cloudcost-explorer
     binary: cloudcost-explorer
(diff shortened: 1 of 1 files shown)


nikimanoledaki:
Tried this locally to validate that it works :)

```
goreleaser release --snapshot --clean 
...
  • release succeeded after 1m46s
  • thanks for using goreleaser!
```

K-Phoen (approved):
https://github.com/grafana/cog/blob/e8e8591e88684737926e94885bee2835fb8cb0ce/.goreleaser.yaml#L16

Makes sense!

PR #372, opened 2024-11-27
Update release workflow's permissions

nikimanoledaki:
- Increase permissions of release workflow from read to write
- Specify version 2 in goreleaser workflow to get rid of warning


K-Phoen (approved):
:rocket:

PR #414, opened 2025-01-28
Generate helm chart configuration docs

nikimanoledaki:
I was browsing the KEDA Helm Chart and realised they use [helm-docs](https://github.com/norwoodj/helm-docs) to generate their Helm Chart config doc README. Adding this here ✨


nikimanoledaki:
I made a mistake with the rebase - cherry picking the last commit to a new branch

PR #415, opened 2025-01-28
Generate Helm Chart docs using helm-docs

nikimanoledaki:
I was browsing the KEDA Helm Chart and realised they use [helm-docs](https://github.com/norwoodj/helm-docs) to generate their Helm Chart config doc README. Adding this here ✨


Pokom (commented):
This is great! Can you add a new target in the `Makefile` to generate the docs? 

Not necessarily needed for this PR, but I'm thinking it's about time we start adding a helm action. For this one, it would be good if we failed the build if there's a diff after running helm-docs.

nikimanoledaki:
@Pokom that's a very good point, we should definitely have a `make` command to validate the helm chart + generate the helm chart docs. On it!

Pokom (approved):
👍🏻 Validated locally. Thanks for adding the `make helm` command!

PR #417, opened 2025-01-29
Define Helm GitHub Actions

Pokom:
This is the first attempt at adding helm github actions. The primary goal is to focus first on tests, then in a later PR we can add releases.

The following tests are run
- checking for doc changes

### DIFF
--- a/.github/workflows/helm.yml
+++ b/.github/workflows/helm.yml
@@ -1,0 +1,28 @@
+---
+name: Helm Tests
+on:
+  push:
+    paths:
+      - 'deploy/helm/**'
+  pull_request:
+    paths:
+      - 'deploy/helm/**'
+  # Allows you to run this workflow manually
+  workflow_dispatch:
+jobs:
+  check-for-doc-changes:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 #v4.2.2
+      - name: Install make
+        run: sudo apt-get update && sudo apt-get install -y make
+        shell: bash
+      - name: Regenerate Helm Docs
+        run: |
+          make helm > /dev/null # we don't actually need to output, just the results
+          if ! git diff --exit-code; then
+            echo "Helm docs are out of date. Please run 'make -C deploy/helm docs' and commit the changes."
+            exit 1
+          fi
+        shell: bash
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -13,18 +13,34 @@ jobs:
   build-lint-test:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
 
-    - uses: ./.github/actions/setup-goversion
+      - uses: ./.github/actions/setup-goversion
 
-    - name: Build
-      run: go build -v ./...
+      - name: Build
+        run: go build -v ./...
 
-    - name: Lint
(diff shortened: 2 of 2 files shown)


.github/workflows/helm.yml
```diff
@@ -0,0 +1,28 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 #v4.2.2
+      - name: Regenerate Helm Docs
+        run: |
+          docker run --rm \
+             -v "$(pwd):/helm-docs" \
+             -u "$(id -u)" \
+             jnorwood/helm-docs:latest # This should probably be a specific version
```
Pokom, line 23: I initially used `make helm` but the action failed due to `make` not being installed. I'll follow up and see how to do that so that we share code between the Makefile and worfkflow.

Pokom:
@nikimanoledaki can you think of any other tests I can add? A nice one we could probably do, but idk the complexity of it, would be to also try a dry run of the outputted charts. Locally I do

`helm template my-release ./deploy/helm/cloudcost-exporter \
  --set replicaCount=3 \
  --set image.tag=latest \
  --set serviceAccount.annotations.eks\\.amazonaws\\.com/role-arn=arn:aws:iam::123456789012:role/CloudCostExporterRole | kubectl apply --dry-run=client -f -
`

nikimanoledaki:
> ```shell
> helm template my-release ./deploy/helm/cloudcost-exporter \
>   --set replicaCount=3 \
>   --set image.tag=latest \
>   --set serviceAccount.annotations.eks\\.amazonaws\\.com/role-arn=arn:aws:iam::123456789012:role/CloudCostExporterRole \
>   | kubectl apply --dry-run=client -f -
> ```

We would need to authenticate with a cluster (could be a kind cluster) -- is that your suggestion? Do have similar integration tests already in cloudcost-exporter? 

I see a similar kind of test suggested in the Helm docs here that uses `helm test ...`: https://helm.sh/docs/topics/chart_tests/#example-test

Actually there is a file in [`deploy/helm/cloudcost-exporter/charts/templates/tests/test-connection.yaml`](https://github.com/grafana/cloudcost-exporter/blob/main/deploy/helm/cloudcost-exporter/templates/tests/test-connection.yaml). It's part of what was generated initially. It's meant to be used as a test e.g. to check that all pods become active etc 👀 

It's worth looking into an integration testing strategy that involves deploying the helm chart etc!

nikimanoledaki (approved):
LGTM for now -- I'd open a separate issue if we want to add integration tests that involve deploying the Helm chart to a kubernetes cluster!

Pokom:
> > ```shell
> > helm template my-release ./deploy/helm/cloudcost-exporter \
> >   --set replicaCount=3 \
> >   --set image.tag=latest \
> >   --set serviceAccount.annotations.eks\\.amazonaws\\.com/role-arn=arn:aws:iam::123456789012:role/CloudCostExporterRole \
> >   | kubectl apply --dry-run=client -f -
> > ```
> 
> We would need to authenticate with a cluster (could be a kind cluster) -- is that your suggestion? Do we have integration tests where cloudcost-exporter is deployed to a cluster?
> 
> I see a similar kind of test suggested in the Helm docs here that uses `helm test ...`: https://helm.sh/docs/topics/chart_tests/#example-test
> 
> Actually there is a file in [`deploy/helm/cloudcost-exporter/charts/templates/tests/test-connection.yaml`](https://github.com/grafana/cloudcost-exporter/blob/main/deploy/helm/cloudcost-exporter/templates/tests/test-connection.yaml). It's part of what was generated initially. It's meant to be used as a test e.g. to check that all pods become active etc 👀
> 
> It's worth devising a strategy for how to do integration tests that involve deploying cloudcost-exporter through the helm chart etc!

The command I proposed does a dry run, so no actual cluster would be targeted. Basically the point would be to
1. Confirm that helm charts gets rendered
2. Confirm that the manifests are valid

It's more static analysis then a full integration suite. But I agree, let's follow up on that in a separate PR 👍🏻

PR #449, opened 2025-03-11
Add GitHub Action to release Helm chart

nikimanoledaki:


.github/workflows/release-helm.yml
```diff
@@ -0,0 +1,116 @@
+---
+name: Release Helm Chart
+
+on:
+  workflow_dispatch: {}
+  push:
+    branches:
+      - main
+    paths:
+        - 'deploy/helm/Chart.yaml'
```
Pokom, line 10: Do we want the workflow to be pinned to releases(IE, tags) similarly to what we do for the docker container?

Pokom: Though, I guess the intent here is to only run this on man with the Chart.yaml file changes. Is that somewhat coupled to the image promotion?

nikimanoledaki: Yeah, we'll need to update the `Chart.yaml` with the new release manually when we create a tag, which will then release the Helm Chart!

Pokom (commented):
More informational questions then anything else. Great start on this @nikimanoledaki!

.github/workflows/release-helm.yml
```diff
@@ -0,0 +1,116 @@
...
+          cd source
+          "${CR_TOOL_PATH}/cr" package "${{ steps.parse-chart.outputs.chartpath }}" --config "${CR_CONFIGFILE}" --package-path "${CR_PACKAGE_PATH}"
+          echo "Result of chart package:"
+          ls -l "${CR_PACKAGE_PATH}"
+
+      - name: Create tag "helm-chart/<version>"
+        run: |
+          cd source
+          echo "Making tag ${{ steps.parse-chart.outputs.tagname }}"
+          git tag "${{ steps.parse-chart.outputs.tagname }}"
```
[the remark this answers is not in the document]
nikimanoledaki: That's a good catch, I knew I had left something inconsistent in the workflow - thank you for finding it. I think this was copied over from the k8s monitoring Helm chart where they create a local release _and_ a release in grafana/helm-charts. In our case, we only need a release in grafana/helm-charts. I should remove this.

.github/workflows/release-helm.yml
nikimanoledaki: I went back and forth with the Helm release cadence. Some other projects trigger the Helm chart's release GHA with any change to `helm/*`. I limited it to `Chart.yaml` so that we only release a new chart when `Chart.yaml`'s `appVersion` (a new release) has been created, that way we keep version parity. What do you think?

nikimanoledaki: Update: I don't think that we need to (or should!) pin the Helm chart version and image version to each other. Here are other examples of `Chart.yaml` from other Grafana Labs-owned charts where these are **not coupled**:
* https://github.com/grafana/helm-charts/blob/main/charts/agent-operator/Chart.yaml
* https://github.com/grafana/loki/blob/main/production/helm/loki/Chart.yaml
* https://github.com/grafana/helm-charts/blob/main/charts/snyk-exporter/Chart.yaml
* https://github.com/grafana/helm-charts/blob/main/charts/grafana/Chart.yaml

I'll update things accordingly!

nikimanoledaki:
Changed our approach here, see:
* https://github.com/grafana/deployment_tools/issues/235950
* https://github.com/grafana/helm-charts/pull/3607

PR #464, opened 2025-03-26
Add credentials service to assume role using STS

nikimanoledaki:
We need to add a new AWS service for credentials that assume a role if prompted through a flag. That might hopefully ™️ unblock us to query a separate account using role chaining.


nikimanoledaki:
Got this to work! 🥳

PR #467, opened 2025-04-03
renamed GCP projects in providerFlags

ohyesgocool:
**Rename gcp.bucket-projects to gcp.projects for clarity**

This merge request improves the user experience when configuring GCP projects by introducing a more intuitive flag name. The existing` gcp.bucket-projects` flag is currently used for all GCP metrics (not just bucket metrics), which is confusing and counterintuitive.

Changes made: 

1. Added a new `gcp.projects` flag in exporter.go that serves the same purpose as the legacy flag
2. Updated the developer guide to reflect the new recommended flag usage

### DIFF
--- a/cmd/exporter/config/config.go
+++ b/cmd/exporter/config/config.go
@@ -16,10 +16,11 @@ type Config struct {
 			RoleARN  string
 		}
 		GCP struct {
-			DefaultGCSDiscount int
-			Projects           StringSliceFlag
-			Region             string
-			Services           StringSliceFlag
+			DefaultGCSDiscount        int
+			Projects                  StringSliceFlag
+			BucketProjectsDeprecated  bool 
+			Region                    string
+			Services                  StringSliceFlag
 		}
 		Azure struct {
 			Services       StringSliceFlag
--- a/cmd/exporter/config/config.go
+++ b/cmd/exporter/config/config.go
@@ -16,11 +16,11 @@ type Config struct {
 			RoleARN  string
 		}
 		GCP struct {
-			DefaultGCSDiscount        int
-			Projects                  StringSliceFlag
-			BucketProjectsDeprecated  bool 
-			Region                    string
-			Services                  StringSliceFlag
+			DefaultGCSDiscount       int
+			Projects                 StringSliceFlag
+			BucketProjectsDeprecated bool
+			Region                   string
+			Services                 StringSliceFlag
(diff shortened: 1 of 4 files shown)


ohyesgocool:
@Pokom should we also add a log warn for the existing users ? Something like - _The flag `'gcp.bucket-projects'` is deprecated, instead use `gcp.projects`_ 😃

Pokom:
@nikimanoledaki when you have a moment, can you take a peak at this? I believe we've discussed this internally a few times, but might be a good opportunity for @leonorfmartins to review as well.

cindy:
> @Pokom should we also add a log warn for the existing users ? Something like - _The flag `'gcp.bucket-projects'` is deprecated, instead use `gcp.projects`_ 😃

Hello there! Sorry this has taken so long to get back to you. I wonder if it would be a good idea to  deprecate this flag and add the new one and also print out a warning that the old one is deprecated and a warning like "'gcp.bucket-projects' is deprecated and will be removed in a future version. Use '--gcp.projects' instead."

nikimanoledaki:
Hi @ohyesgocool 👋 

@cindy shared some great feedback. Do you have capacity to implement the changes they recommended, please?
- Deprecate the `gcp.bucket-projects` flag
- Add `gcp.projects` as a new flag
- Print a warning that the old flag is deprecated: `'gcp.bucket-projects' is deprecated and will be removed in a future version. Use '--gcp.projects' instead.`

nikimanoledaki (commented):
```
go run cmd/exporter/exporter.go -provider gcp -project-id=$GCP_PROJECT_ID -gcp.bucket-projects=$GPC_PROJECT_ID -gcp.bucket-projects=$GPC_PROJECT_ID
time=2026-02-17T10:53:21.271+01:00 level=INFO msg="Starting cloudcost-exporter" version="(version=, branch=, revision=unknown)" build_context="(go=go1.24.6, platform=darwin/amd64, user=, date=, tags=unknown)"
time=2026-02-17T10:53:21.272+01:00 level=WARN msg="'gcp.bucket-projects' is deprecated and will be removed in a future version. Use '--gcp.projects' instead."
```

Validated the log warning ✅

nikimanoledaki (approved):

nikimanoledaki:
I fixed that issue in main so if you rebase with main it should be good to go!

PR #477, opened 2025-04-15
Assume AWS IAM role through EC2 client

nikimanoledaki:
This PR:
* Refactors the func to assume a role into a separate func
* Refactors the creation of a new EC2 client to include previous options and only change the region to override previous region configurations and not lose other options as these scale

Note to reviewer: this has been tested in dev cluster ✅

### DIFF
--- a/pkg/aws/aws.go
+++ b/pkg/aws/aws.go
@@ -147,7 +147,7 @@ func New(ctx context.Context, config *Config) (*AWS, error) {
 			}
 			regionClientMap := make(map[string]ec2client.EC2)
 			for _, r := range regions.Regions {
-				client, err := newEc2Client(*r.RegionName, config.Profile, config.RoleARN)
+				client, err := newEc2Client(*r.RegionName, ac)
 				if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/aws/aws.go
```diff
@@ -248,18 +237,35 @@ func (a *AWS) Collect(ch chan<- prometheus.Metric) {
...
+	if err != nil {
+		return nil, err
 	}
-	// Set max retries to 10. Throttling is possible after fetching the pricing data, so setting it to 10 ensures the next scrape will be successful.
-	options = append(options, awsconfig.WithRetryMaxAttempts(maxRetryAttempts))
+
+	return ec2.NewFromConfig(ac), nil
+}
+
+func assumeRole(roleARN string, options []func(*awsconfig.LoadOptions) error) ([]func(*awsconfig.LoadOptions) error, error) {
```
nikimanoledaki, line 252: This is not a full refactor but I'm splitting this as a helper func since the plan is to refactor `New` at some point...

Pokom (approved):

PR #481, opened 2025-04-25
Validate Role ARN before assuming role

nikimanoledaki:
Adding validation for the Role ARN, which I forgot to add while refactoring 😰

### DIFF
--- a/pkg/aws/aws.go
+++ b/pkg/aws/aws.go
@@ -246,10 +246,12 @@ func newEc2Client(region, profile, roleARN string) (*ec2.Client, error) {
 	// Set max retries to 10. Throttling is possible after fetching the pricing data, so setting it to 10 ensures the next scrape will be successful.
 	options = append(options, awsconfig.WithRetryMaxAttempts(maxRetryAttempts))
 
-	var err error
-	options, err = assumeRole(roleARN, options)
-	if err != nil {
-		return nil, err
+	if roleARN != "" {
+		var err error
+		options, err = assumeRole(roleARN, options)
+		if err != nil {
+			return nil, err
+		}
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/aws/aws.go
```diff
@@ -246,10 +246,12 @@ func newEc2Client(region, profile, roleARN string) (*ec2.Client, error) {
...
-	var err error
-	options, err = assumeRole(roleARN, options)
-	if err != nil {
-		return nil, err
+	if roleARN != "" {
+		var err error
+		options, err = assumeRole(roleARN, options)
+		if err != nil {
+			return nil, err
+		}
```
Pokom, line 254: Is the intent here to fail the start of the module if you can not assume the role? IE, would this have prevented a `v0.8.1` image from starting?

Pokom (approved):

pkg/aws/aws.go
nikimanoledaki: This is a great question. I will try to answer here, then move things to the docs!

There are two roles that cloudcost-exporter can use: the first one is the IRSA role. This is used as authentication between the cluster to the CSP via OIDC. The second role that can be passed to cloudcost-exporter is a role that it should assume (for example, to gain access to the resources in a separate account). The first role is required while the second role is optional.

> Is the intent here to fail the start of the module if you can not assume the role?

Assuming a role is optional. It should only attempt to assume a new role if a Role ARN has been passed through the flag: `--set '--aws.roleARN=<role-arn>'`. Maybe this flag should be named differently, like `assumedRoleARN`? 🤔

nikimanoledaki: merging this PR and adding docs here https://github.com/grafana/cloudcost-exporter/pull/487

PR #500, opened 2025-06-05
cloudcost-exporter: AWS ELB metrics support

jvrplmlmn:
This PR introduces AWS ELB (Elastic Load Balancing) support. When enabled (`--provider=aws --aws.services=elb (...)`), it will expose 2 new types of metrics: 
* `cloudcost_aws_elb_loadbalancer_usage_total_usd_per_hour`
* `cloudcost_aws_elb_loadbalancer_capacity_units_total_usd_per_hour`

These metrics are added for the AWS ELB service, for Network LBs and Application LBs, which is added as a label:
```
cloudcost_aws_elb_loadbalancer_usage_total_usd_per_hour{name="<redacted>",region="<redacted>",type="network"} 0.0225
cloudcost_aws_elb_loadbalancer_usage_total_usd_per_hour{name="<redacted>",region="<redacted>",type="application"} 0.0225 

cloudcost_aws_elb_loadbalancer_capacity_units_total_usd_per_hour{name="<redacted>",region="<redacted>",type="network"} 0.006
cloudcost_aws_elb_loadbalancer_capacity_units_total_usd_per_hour{name="<redacted>",region="<redacted>",type="application"} 0.008
```

Classic LBs are not supported since they were created via the deprecated AWS SDK and are being phased out in general.


pkg/aws/aws.go
```diff
@@ -160,6 +163,28 @@ func New(ctx context.Context, config *Config) (*AWS, error) {
 				ScrapeInterval: config.ScrapeInterval,
 			}, pricingService)
 			collectors = append(collectors, collector)
+		case "ELB":
+			pricingService := pricing.NewFromConfig(ac)
+			computeService := ec2.NewFromConfig(ac)
```
nikimanoledaki, line 168: This should initialise the ELB collector

pkg/aws/elb/elb.go
```diff
@@ -0,0 +1,350 @@
...
+			mu.Unlock()
+
+			return nil
+		})
+	}
+
+	return eg.Wait()
+}
+
+func (c *Collector) fetchRegionPricing(region string) (*RegionPricing, error) {
```
nikimanoledaki: Please move pricing-related logic to `pricing_map.go`. For an example on how to do this, checkout the AKS collector:  https://github.com/grafana/cloudcost-exporter/blob/main/pkg/azure/aks/price_store.go#L184

pkg/aws/elb/elb.go
```diff
@@ -0,0 +1,350 @@
...
+func (c *Collector) collectLoadBalancers() ([]LoadBalancerInfo, error) {
+	var allLoadBalancers []LoadBalancerInfo
+	var mu sync.Mutex
+
+	eg := errgroup.Group{}
+
+	for regionName, client := range c.elbRegionClients {
+		regionName := regionName
+		client := client
+
```
nikimanoledaki: I'm wondering whether we should add a `sync.WaitGroup` and add `wg.Add(foo)` where `foo` is the number of regions, before initialising the Go routines 🤔 WDYT?

This is likely something we'll have to backfill for all collectors, so no worries if this is out of the scope of this PR. I can open a separate issue for it. Would love to hear what you think from a technical perspective.

pkg/aws/elb/elb.go
```diff
@@ -0,0 +1,350 @@
...
+	// Set default rates if not found (fallback values)
+	if len(regionPricing.ALBHourlyRate) == 0 {
+		regionPricing.ALBHourlyRate["default"] = 0.0225 // Default ALB rate
+	}
+	if len(regionPricing.NLBHourlyRate) == 0 {
+		regionPricing.NLBHourlyRate["default"] = 0.0225 // Default NLB rate
+	}
+	if len(regionPricing.CLBHourlyRate) == 0 {
+		regionPricing.CLBHourlyRate["default"] = 0.025 // Default CLB rate
+	}
```
nikimanoledaki: These default rates should be set as constants (and moved to `pricing_map.go` along with the rest of the func)

nikimanoledaki (commented):
Thank you @jvrplmlmn! 👏 This is really neat. I've added a couple of review comments for the pricing logic to maintain consistency with the AKS package since Mark had mentioned that one was a good example to follow. We'll need to find ways to abstract away some of the common init logic if we find time to refactor the cloudcost-exporter.

pkg/aws/elb/elb.go
nikimanoledaki: This is non-blocking for this PR :)

cindy:
I'll take a stab at this one.

pkg/aws/elb/elb.go
cindy: @nikimanoledaki I think I need clarification here. Are you suggesting to simply move the code as a pricing map or to implement a price store (like the one in AKS)?

nikimanoledaki: Ideally both, since we should keep pricing logic separate from `pkg/aws/elb/elb.go` and follow the AKS implementation as an example / best practice

nikimanoledaki: Resolving since the current PR is good for now - let's refactor in a follow-up PR once https://github.com/grafana/cloudcost-exporter/pull/569/files is merged

cindy:
```bash
$ go run cmd/exporter/exporter.go -provider aws -aws.profile $AWS_PROFILE --aws.services=elb --aws.region=us-east-1
time=2025-08-19T22:53:54.468Z level=INFO msg="Starting cloudcost-exporter" version="(version=, branch=, revision=unknown)" build_context="(go=go1.24.5, platform=darwin/arm64, user=, date=, tags=unknown)"
time=2025-08-19T22:53:58.093Z level=INFO msg="registering collectors" provider=aws count=1
time=2025-08-19T22:53:58.094Z level=INFO msg="Starting server" address=:8080 path=/metrics
time=2025-08-19T22:54:09.670Z level=INFO msg="Starting ELB collection" provider=aws
time=2025-08-19T22:54:09.670Z level=INFO msg="Refreshing ELB pricing data" provider=aws
time=2025-08-19T22:54:12.496Z level=INFO msg="Completed ELB collection" provider=aws load_balancers=112
```

```bash
$ curl localhost:8080/metrics |grep ^cloudcost_aws_elb_loadbalancer | wc -l
     224 # 112 LBs x 2 metrics 
```

PR #507, opened 2025-06-20
Add missing GCP success metric. Fix success metric counter

leonorfmartins:
Fixes https://github.com/grafana/deployment_tools/issues/284190 and modifies the success metric to be = total metric - amount of errors during each scraping interval.


pkg/aws/aws.go
```diff
@@ -44,7 +45,7 @@ var (
 	collectorSuccessDesc = prometheus.NewDesc(
 		prometheus.BuildFQName(cloudcost_exporter.ExporterName, subsystem, "collector_success"),
 		"Was the last scrape of the AWS metrics successful.",
```
jjo: Can you clarify the description?
"Was the last scrape of the AWS metrics successful." sounds like a boolean that will flip between 0,1 (Gauge?), then shouldn't it be better for this to be a Counter instead?

pkg/azure/azure.go
```diff
@@ -53,9 +55,9 @@ var (
 		[]string{"provider", "collector"},
 	)
 	collectorSuccessDesc = prometheus.NewDesc(
-		prometheus.BuildFQName(cloudcost_exporter.ExporterName, subsystem, "collector_success"),
-		"Was the last scrape of the Azure metrics successful.",
-		[]string{"collector"},
+		prometheus.BuildFQName(cloudcost_exporter.ExporterName, "collector", "success"),
+		"Number of successful scrapes for the collector.",
```
jjo: Here the description doesn't match the previous one (see my previous comment), in this case it sounds like a Counter (strongly preferred indeed).

leonorfmartins: hm, that's a good point 🤔 I took the same type as `collectorLastScrapeErrorDesc`, which is also a gauge, since successful = total - errors. Total scrapes is a counter, so it seems like the errors one should also be a counter. I'll change both the error one and the success scrape one 😄

PR #519, opened 2025-07-18
Remove unused metrics

leonorfmartins:
After defining SLOs for this application, there is some cleanup we can in regards to metrics, namely, remove `cloudcost_collector_success` and `cloudcost_collector_scrapes_total`

They weren't reliable metrics: total scrapes is discouraged to be collected according to [Prometheus collector best practices](https://prometheus.io/docs/instrumenting/writing_exporters/#metrics-about-the-scrape-itself) and successful scrapes was based on the total scrapes, so there's also no point on leaving this around.

Relates to https://github.com/grafana/deployment_tools/issues/275831

### DIFF
--- a/pkg/aws/aws.go
+++ b/pkg/aws/aws.go
@@ -16,7 +16,6 @@ import (
 	"github.com/aws/aws-sdk-go-v2/service/pricing"
 	"github.com/aws/aws-sdk-go-v2/service/sts"
 	"github.com/prometheus/client_golang/prometheus"
-	dto "github.com/prometheus/client_model/go"
 
 	ec2Collector "github.com/grafana/cloudcost-exporter/pkg/aws/ec2"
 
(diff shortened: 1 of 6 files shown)


spinillos (approved):

nikimanoledaki (commented):
Do you think that we should remove `“last_scrape_duration_seconds”` as well, given [the following advice](https://prometheus.io/docs/instrumenting/writing_exporters/#metrics-about-the-scrape-itself)?
> Other scrape “meta” metrics should be avoided. For example, a counter for the number of scrapes, or a histogram of the scrape duration. 

WDYT about `"last_scrape_time"` & `"last_scrape_error"`?

jjo:
> Do you think that we should remove `“last_scrape_duration_seconds”` as well, given [that it's also explicitly advised against in the docs](https://prometheus.io/docs/instrumenting/writing_exporters/#metrics-about-the-scrape-itself)?
> 

IMO it'd be relevant keeping it if there's some "sync" reason behind: i.e. the scrape time depends for example on the exporter itself blocking while consuming some provider API / alike, that we could then use to track it, not only for SLO purposes but also e.g. when adding new observed resources to assess their impact. All this previous sentence is N/A if the scrape is _async_, i.e. scraping only visits internal memory objects.

nikimanoledaki (approved):
We can merge this and iterate over the collector metrics that may be duplicates ☺️ I don't want to block this work. We'll be instrumentalizing our `/metrics` endpoint next so we're developing our metrics/instrumentalization process for cloudcost-exporter well 👍

PR #523, opened 2025-07-23
Add duration and request errors on the metrics endpoint

leonorfmartins:
Instrument prometheus using InstrumentHandlerDuration and InstrumentHandlerCounter so we get added the following metrics:
- `cloudcost_exporter_request_duration_seconds_bucket`
- `cloudcost_exporter_request_duration_seconds_sum`
- `cloudcost_exporter_request_duration_seconds_count`
- `cloudcost_exporter_requests_total`

Screenshot 2025-07-23 at 13 17 32

It will help to define SLOs in https://github.com/grafana/deployment_tools/pull/302655

### DIFF
--- a/cmd/exporter/exporter.go
+++ b/cmd/exporter/exporter.go
@@ -139,6 +139,8 @@ func runServer(ctx context.Context, cfg *config.Config, csp provider.Provider, l
 			return fmt.Errorf("error shutting down server: %w", err)
 		}
 	case err := <-errChan:
+		// FIXME: This is not correct, we need to get the status code from the response
+		// and it doesn't fit how histogram works as they expect a value distribution and not a counter.
 		requestErrors.WithLabelValues(cfg.Server.Path, "GET").Observe(1)
 		if !errors.Is(err, http.ErrServerClosed) {
 			return fmt.Errorf("error running server: %w", err)
--- a/cmd/exporter/exporter.go
+++ b/cmd/exporter/exporter.go
@@ -141,10 +141,11 @@ func runServer(ctx context.Context, cfg *config.Config, csp provider.Provider, l
 }
 
 func createPromRegistryHandler(csp provider.Provider) (http.Handler, error) {
+	var subsystem = "metrics_handler"
 	requestDuration := prometheus.NewHistogramVec(
 		prometheus.HistogramOpts{
-			Name:    "cloudcost_exporter_request_duration_seconds",
-			Help:    "Duration of HTTP requests in seconds",
+			Name:    prometheus.BuildFQName(cloudcost_exporter.ExporterName, subsystem, "request_duration_seconds"),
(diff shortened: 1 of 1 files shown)


cmd/exporter/exporter.go
```diff
@@ -141,22 +141,48 @@ func runServer(ctx context.Context, cfg *config.Config, csp provider.Provider, l
 }
 
 func createPromRegistryHandler(csp provider.Provider) (http.Handler, error) {
+	requestDuration := prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Name:    "cloudcost_exporter_request_duration_seconds",
```
nikimanoledaki: The metric name should be set [like this](https://github.com/grafana/cloudcost-exporter/blob/dcd6dfbc51509754c56d357606f08b2b8146044b/pkg/aws/s3/s3.go#L92C11-L92C120)

I would also recommend naming the subsystem of each metric to `metric_handler_duration_seconds` & `metric_handler_requests_total` since they are used exclusively for the `/metrics` path

cmd/exporter/exporter.go
```diff
@@ -141,22 +141,48 @@ func runServer(ctx context.Context, cfg *config.Config, csp provider.Provider, l
...
+			Help:    "Duration of HTTP requests in seconds",
```
nikimanoledaki: Since this is only used for the metrics endpoint:
```suggestion
			Help:    "Duration of HTTP requests in seconds for the metrics endpoint",
```

cmd/exporter/exporter.go
```diff
@@ -141,22 +141,48 @@ func runServer(ctx context.Context, cfg *config.Config, csp provider.Provider, l
...
+		},
+		[]string{"method"},
+	)
+
+	requestCounter := prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "cloudcost_exporter_requests_total",
+			Help: "Total number of HTTP requests",
+		},
+		[]string{"code", "method"},
```
nikimanoledaki, line 159: Should `code` be `status` for parity with other similar metrics? Or is this what the default helper func returns?

nikimanoledaki (commented):
A couple of change requests but otherwise we're so close to shipping this! 😍

cmd/exporter/exporter.go
leonorfmartins: It actually panics if it's not `code`, as it expects this label to be there (fails with `panic: metric partitioned with non-supported labels`)

I followed the example set defined in the promhttp package [here](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus/promhttp#example-InstrumentHandlerDuration)

(looking at other random counterVec metrics, it seems `code` is also used?)
Screenshot 2025-07-23 at 13 43 03

nikimanoledaki (approved):
LGTM 🥳

PR #534, opened 2025-07-28
Abstract aws client so it can be reused for RDS

leonorfmartins:


pkg/aws/aws.go
```diff
@@ -108,26 +110,28 @@ func New(ctx context.Context, config *Config) (*AWS, error) {
...
-			}
 			regionClientMap := make(map[string]ec2client.EC2)
 			for _, r := range regions.Regions {
-				client, err := newEc2Client(*r.RegionName, config.Profile, config.RoleARN)
-				if err != nil {
-					return nil, fmt.Errorf("error creating ec2 client: %w", err)
-				}
-				regionClientMap[*r.RegionName] = client
+				ec2Client := ec2.NewFromConfig(ac)
+				regionClientMap[*r.RegionName] = ec2Client
```
spinillos:
```suggestion
				regionClientMap[*r.RegionName] = ec2.NewFromConfig(ac)
```

pkg/aws/aws.go
```diff
@@ -108,26 +110,28 @@ func New(ctx context.Context, config *Config) (*AWS, error) {
...
+			if err != nil {
+				return nil, fmt.Errorf("error getting regions: %w", err)
+			}
+		}
 		switch strings.ToUpper(service) {
 		case "S3":
 			client := costexplorer.NewFromConfig(ac)
 			collector := s3.New(config.ScrapeInterval, client)
 			collectors = append(collectors, collector)
 		case "EC2":
```
spinillos, line 130: I'm thinking that it could be `case "EC2", "RDS"`, initialise both, but only add the colector if the service is the expected one. Mainly because you need to initialise ec2 in both cases, and they have common configuration.

It leaves only the switch removing the service check at the beginning to get the regions.

leonorfmartins: I'm not sure what you mean, I do have some questions:
- how would we handle the mapping region - client? We seem to need to iterate over regions to assign the particular ec2 or rds client. Are you suggesting we could abstract this as well?
- the plan is to add an RDS collector as well, but right now it's very much WIP as we first need to figure out out to fetch and parse pricings for each instance :)

cmd/exporter/exporter.go
```diff
@@ -85,7 +85,7 @@ func operationalFlags(cfg *config.Config) {
 	flag.DurationVar(&cfg.Collector.ScrapeInterval, "scrape-interval", 1*time.Hour, "Scrape interval")
 	flag.DurationVar(&cfg.Collector.Timeout, "collector-interval", 1*time.Minute, "Context timeout for collectors")
 	flag.DurationVar(&cfg.Server.Timeout, "server-timeout", 30*time.Second, "Server timeout")
-	flag.StringVar(&cfg.Server.Address, "server.address", ":8080", "Default address for the server to listen on.")
+	flag.StringVar(&cfg.Server.Address, "server.address", ":8000", "Default address for the server to listen on.")
```
cindy: Was this port number change intended?

leonorfmartins: no 🙈 it collides with my local Docker, so I changed it to test the app locally, but we shouldn't change it. Thanks!

PR #535, opened 2025-07-28
Refactor GCS and GKE from Google provider

spinillos:
This PR seems a bit large but its mainly moving some pieces of the code to a different place. Why? Because the providers are working totally "different" in terms of how the code was implemented. Some of them have all logic inside `CollectMetrics`, others isolates part of the code in a different file, configuration of each service saves information that doesn't belong to the proper collector, etc...

Changes:
* A client that embed all the GCP services, an interface of the necessary calls and a mock of this interface to use it in tests.
  * Includes billing, storage, regions and compute services.
* BucketsCache is using directly in buckets service, removing it from the collector. It also has an interface and and dummy mock for testing.
* Metrics are moved into `metrics` folder to avoid dependency cycles (I think that it can be improved but in other PR)
* Most of the tests from `gcs_test` are now inside the client's test since most of them belong to the API.

Things that aren't improved in this PR:
* The new client exposes GCP models that should be mapped to a custom one to avoid to to expose it outside the client.
* Some client calls accepts "metrics" as argument because it was the easiest way to do it. Metrics must be called only inside collector to have everything in the same place.


pkg/google/client/gcp_client.go
```diff
@@ -0,0 +1,87 @@
...
+	regions *Region
+	bucket  *Bucket
+}
+
+type Config struct {
+	ProjectId string
+	Discount  int
+}
+
+func NewGPCClient(ctx context.Context, cfg Config) (*GPCClient, error) {
```
jjo: typo `s/GPC/GCP/`

pkg/google/client/gcp_client.go
```diff
@@ -0,0 +1,87 @@
...
+	billingv1 "cloud.google.com/go/billing/apiv1"
+	"cloud.google.com/go/billing/apiv1/billingpb"
+	computeapiv1 "cloud.google.com/go/compute/apiv1"
+	"cloud.google.com/go/storage"
+	"github.com/grafana/cloudcost-exporter/pkg/google/client/cache"
+	"github.com/grafana/cloudcost-exporter/pkg/google/metrics"
+	computev1 "google.golang.org/api/compute/v1"
+)
+
+type GPCClient struct {
```
jjo: typo s/GPC/GCP/

jjo (commented):
1st pass, found a few more `gpcclient` cases where it should be `gcpclient`

jjo (approved):
LGTM, thanks for this!

did a smoke-test verification with
```
$ make build-binary
$ ./cloudcost-exporter --provider=gcp --project-id=grafanalabs-global --gcp.bucket-projects=grafanalabs-global,ops-tools-1203 --gcp.services=gcs,gke &

$ curl -s http://localhost:8080/metrics | less
```
then some manual sampling for some expected timeseries

PR #548, opened 2025-07-30
Refactor azure and generate all mocks as the same way.

spinillos:
It adds a small cleanup in azure provider, but this one already had a client for azure SDK so it isn't too much work.

And I saw that some 2 mocks were generating by a make command, others with `go:generate` comment. Apart of this, the mocks were mixing `gomock` from Uber with `mock` from testify library. I leave everything using the Uber's one.

To have all the mocks as the same way, I used the comment way. If you are using an IDE, it adds an option on the left side to create/update the mock (at least in Goland).
image


rooneyshuman (approved):
LGTM! 🚀

PR #561, opened 2025-08-11
Create a client per region

spinillos:
EC2 needs a client per region instead the same for all clients. It restores the previous code adapted to the latest refactor to have everything like before.

A test was deleted because it doesn't test anything, and [this one](https://github.com/grafana/cloudcost-exporter/compare/aws/client-per-region?expand=1#diff-30e0dcbef190e9ab58c39cd1943eda8b2e2da8fb35408aa1e4514fb510b51a43R79) was fixed.

### DIFF
--- a/pkg/aws/client/aws_client.go
+++ b/pkg/aws/client/aws_client.go
@@ -12,8 +12,6 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 )
 
-const maxRetryAttempts = 10
-
 type Config struct {
 	PricingService p.Pricing
 	EC2Service     e.EC2


pkg/aws/aws.go
```diff
@@ -105,7 +117,8 @@ func New(ctx context.Context, config *Config) (*AWS, error) {
 				Regions:        regions,
 				Logger:         logger,
 				ScrapeInterval: config.ScrapeInterval,
-			}, awsClient)
+				RegionMap:      awsClientPerRegion,
```
nikimanoledaki, line 125: Since the logic in L99-110 only applies to `case serviceEC2`, could we move this out of `New` and back to a helper func?

nikimanoledaki (approved):
this is great. thank you! 🌟

PR #564, opened 2025-08-12
feat(cloudcost-exporter): add initial natGW support for AWS and Google

jjo:
Ref: #311440.

--


pkg/aws/aws.go
```diff
@@ -115,6 +119,24 @@ func New(ctx context.Context, config *Config) (*AWS, error) {
...
+			if config.Region != "" {
+				loadOpts = append(loadOpts, awsconfig.WithRegion(config.Region))
+			}
+			acfg, err := awsconfig.LoadDefaultConfig(ctx, loadOpts...)
+			if err != nil {
+				return nil, fmt.Errorf("failed to load AWS config for NAT Gateway: %w", err)
+			}
+			ceClient := awscostexplorer.NewFromConfig(acfg)
+			gwCollector := awsgwnat.New(config.ScrapeInterval, ceClient)
+			collectors = append(collectors, gwCollector)
```
nikimanoledaki: Please update to the new client initialization method - it should abstract a lot of this away 😄  for an example, [see `case serviceEC2` on `main`](https://github.com/grafana/cloudcost-exporter/blob/main/pkg/aws/aws.go#L120-L127)

jjo: for now I've just rebase-d it at 4261dc4  (essentially to make it work again and void losing functionality), will see how to use new multi-cli stuff

pkg/aws/common/common.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+	return strings.Join(split[1:], "-")
+}
+
+// isBillExpirationError determines whether the error returned by Cost Explorer
+// indicates that the bill changed during pagination and the request should be retried.
+func IsBillExpirationError(err error) bool {
+	var apiErr smithy.APIError
+	if errors.As(err, &apiErr) {
+		if apiErr.ErrorCode() == "BillExpirationException" {
```
nikimanoledaki: Try this:
```suggestion
		if apiErr.ErrorCode() == awscostexplorer.ErrCodeBillExpirationException {
```
Some of the errors are defined in the SDK at https://pkg.go.dev/github.com/archsaber/aws-sdk-go-v2/service/costexplorer#pkg-constants :)

jjo: :( 
```
pkg/aws/common/common.go:70:44: undefined: awscostexplorer.ErrCodeBillExpirationException
```

jjo: there seems to be a diff between these two ->
```
      awscostexplorer "github.com/aws/aws-sdk-go-v2/service/costexplorer"
+     "github.com/aws/aws-sdk-go/service/costexplorer"
```
the latter does have that symbol, but
```
go: module github.com/aws/aws-sdk-go is deprecated: aws-sdk-go is deprecated. Use aws-sdk-go-v2.
```

jjo: but then, even after `go get ` blah/etc forcing _also_ this older sdk, ends up being the same string ->
```
    // The requested report expired. Update the date interval and try again.
    ErrCodeBillExpirationException = "BillExpirationException"
```

nikimanoledaki: Ah they had it in SDK v1 https://github.com/aws/aws-sdk-go/blob/v1.55.7/service/costexplorer/errors.go#L22 😞

pkg/aws/aws.go
jjo: hopef00ly addressed at c7bc4ce

pkg/aws/common/common.go
nikimanoledaki: Here's what it looks like in SDK v2: https://github.com/aws/aws-sdk-go-v2/blob/main/service/costexplorer/types/errors.go#L65 

It should be possible to use that type to check the error :) (edit: hopefully 🤞 )

pkg/aws/aws.go
```diff
@@ -133,6 +135,10 @@ func New(ctx context.Context, config *Config) (*AWS, error) {
 			}, awsClient)
 			// TODO: append new aws rds collectors next
 			// collectors = append(collectors, collector)
+		case serviceNATGW:
+			ceClient := costexplorer.NewFromConfig(ac)
+			gwCollector := awsgwnat.New(config.ScrapeInterval, ceClient)
```
nikimanoledaki: The Cost Explorer client is created above as `BillingService: costexplorer.NewFromConfig(ac),`

This includes all the config params included in the AWS client (region, profile, role ARN)

Then set the AWS client here:
```suggestion
			gwCollector := awsgwnat.New(config.ScrapeInterval, awsClient)
```
I think that's the pattern 😄 Will revisit tomorrow with fresh eyes!

nikimanoledaki:
Note: I'm refactoring some of this work & testing the authentication with different profiles ⚠️

pkg/google/natgateway/natgateway.go
```diff
@@ -0,0 +1,178 @@
...
+		// service regions are like: [us-central1, europe-west1, ...]
+		for _, region := range sku.ServiceRegions {
+			for _, project := range projects {
+				skuID := sku.GetName()
+				skuName := sku.Description
+				if isDataProcessing(desc) {
+					c.metrics.DataProcessingGauge.WithLabelValues(region, project, skuID, skuName).Set(priceUSD)
+				} else {
+					c.metrics.HourlyRateGauge.WithLabelValues(region, project, skuID, skuName).Set(priceUSD)
+				}
```
nikimanoledaki: Are we including the SKU `Networking Cloud NAT IP Usage` here? It's a [different charge type](https://cloud.google.com/nat/pricing):
> Price per hour for a static or an ephemeral external IP address used by a NAT gateway**
> $0.005

nikimanoledaki: I see that this is specified in the metric via the label `sku_name` - perfect!

jjo: BTW removed google support, moving it to a followup PR.

nikimanoledaki (approved):

Issue #575, opened 2025-08-19
Add metrics for RDS (AWS) & Cloud SQL (GCP)


nikimanoledaki:
### Cloud Providers

AWS, GCP

### Feature Request

Add metrics for pricing per region for RDS (AWS) & Cloud SQL (GCP):

* Docs on how to add a new module: https://github.com/grafana/cloudcost-exporter/blob/main/docs/contribute/creating-a-new-module.md
* Example PR: https://github.com/grafana/cloudcost-exporter/pull/569
* Metric should be something like `aws_rds_instance_running_hours`

## AWS

RDS instances are EC2 instances under the hood. That means that we should find what type of EC2 instance an RDS uses, find the EC2 instance pricing for it, and join that information.

In the AWS console's Cost Explorer, the billing-level/actual RDS costs can be found with these filters:
* Service: `Relational Database Service (RDS)`
* Usage Type Group: `RDS: Running Hours`
* Usage Type: The string should include `Multi-AZUsage`
* Group by the Dimension `Instance type` to find which EC2 instances are used e.g. `db.m5.large`

Get the DB instance info
* The [DescribeDBInstances](https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_DescribeDBInstances.html_) API call returns the [DB's instance class](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Concepts.DBInstanceClass.html ) (`DBInstanceClass`) e.g. `db.m4.xlarge`. We should be able to deduce the type of EC2 instance used from this. 

Pricing information
* https://aws.amazon.com/rds/pricing/
* https://aws.amazon.com/blogs/aws/aws-price-list-api-update-regional-price-lists/ <-- check this to see the schema of the pricing API

## GCP

[SKU: `Cloud SQL`](https://cloud.google.com/skus?hl=en&filter=cloud%20sql&currency=USD)

TBD - follow a similar logic to find the GCP pricing per compute.

leonorfmartins:
For CloudSQL:
We can use this package to list instances: https://pkg.go.dev/google.golang.org/api@v0.253.0/sqladmin/v1beta4#InstancesListCall.Pages

PR #593, opened 2025-09-08
Get pricelist for AWS RDS

leonorfmartins:
Get price rates for all AWS RDS instances.

This PR only instantiates the RDS client and fetches the price list. Work for populating this into metrics and exporting them will come in a follow-up PR.

Part of https://github.com/grafana/cloudcost-exporter/issues/575

### DIFF
--- a/pkg/aws/aws.go
+++ b/pkg/aws/aws.go
@@ -93,6 +93,7 @@ func New(ctx context.Context, config *Config) (*AWS, error) {
 	if err != nil {
 		return nil, err
 	}
+
 	awsClient := client.NewAWSClient(client.Config{
 		PricingService: awsPricing.NewFromConfig(ac),
 		EC2Service:     ec2.NewFromConfig(ac),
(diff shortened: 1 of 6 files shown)


pkg/aws/aws.go
```diff
@@ -130,14 +129,30 @@ func New(ctx context.Context, config *Config) (*AWS, error) {
...
 		case serviceRDS:
-			_ = rds.New(&rds.Config{
+			// pricing API for RDS client needs to use always the same region
+			// as for RDS , the pricing data is only available in the us-east-1
+			pricingConfig, err := createAWSConfig(ctx, "us-east-1", config.Profile, config.RoleARN)
+			if err != nil {
+				return nil, err
+			}
+			awsRDSClient := client.NewAWSClient(client.Config{
+				PricingService: awsPricing.NewFromConfig(pricingConfig),
```
jjo, line 136: Why do we need to (also) initialize EC2, ELB for this RDS client?

leonorfmartins: Good catch, you are totally right! The only thing we are actually needing implicitly was the ec2 client because we were fetching regions again. But, since `describeregions` [using the ec2 client seems to be the only way the AWS API allows to get regions](https://stackoverflow.com/questions/65843812/how-to-get-a-list-of-aws-regions-programmatically-in-go-sdk-v2?utm_source=chatgpt.com) for any service, I'm removing the if statement (which was covering all services anyways) and fetching the regions before creating the client. Updated that accordingly :)

PR #604, opened 2025-09-15
Collect AWS RDS hourly cost metrics

leonorfmartins:
Process to collect RDS hourly cost metrics:
- fetch all existing RDS instances
- identify the 5 dimensions needed to uniquely get an hourly rate for running a certain instance: `tier`, `region`, `engine`,  `single or multi availability` and `location type`.
- if there is already an hourly price for a certain DB instance in memory, use that one
- otherwise, use the pricing API to fetch the hourly price.

Chosen metric labels are db `name`, `tier` and `region`. We can later join that with `aws_rds_info{}` on db `name` to get the DB's namespace.

NOTE:
Refreshing metrics takes ~1m and I'll work on parallelising this process next. However, considering that the scraping interval happens every 60m, I don't think it's a blocker to get this merged. 

Example of exported metrics:
```
# TYPE cloudcost_aws_rds_hourly_rate_usd_per_hour gauge
cloudcost_aws_rds_hourly_rate_usd_per_hour{name="asserts-prod-ap-northeast-0",region="ap-northeast-1",tier="db.t4g.small"} 0.101
cloudcost_aws_rds_hourly_rate_usd_per_hour{name="asserts-prod-ap-south-1",region="ap-south-1",tier="db.t4g.small"} 0.084
cloudcost_aws_rds_hourly_rate_usd_per_hour{name="asserts-prod-ap-southeast-1",region="ap-southeast-1",tier="db.t4g.small"} 0.102
cloudcost_aws_rds_hourly_rate_usd_per_hour{name="asserts-prod-ap-southeast-2",region="ap-southeast-3",tier="db.t4g.small"} 0.102
cloudcost_aws_rds_hourly_rate_usd_per_hour{name="asserts-prod-au-southeast-1",region="ap-southeast-2",tier="db.t4g.small"} 0.102
cloudcost_aws_rds_hourly_rate_usd_per_hour{name="asserts-prod-ca-east-0",region="ca-central-1",tier="db.t4g.small"} 0.07
cloudcost_aws_rds_hourly_rate_usd_per_hour{name="asserts-prod-eu-central-0",region="eu-central-2",tier="db.t4g.small"} 0.082
cloudcost_aws_rds_hourly_rate_usd_per_hour{name="asserts-prod-eu-north-0",region="eu-north-1",tier="db.t4g.small"} 0.065
```

### DIFF
--- a/pkg/aws/aws.go
+++ b/pkg/aws/aws.go
@@ -134,8 +134,7 @@ func New(ctx context.Context, config *Config) (*AWS, error) {
 			}
 			awsRDSClient := client.NewAWSClient(client.Config{
 				PricingService: awsPricing.NewFromConfig(pricingConfig),
-				// EC2Service:     ec2.NewFromConfig(ac), //we also need to pass in the ec2 client in order to describe regions, as they are shared between RDS and EC2
-				RDSService: rds2.NewFromConfig(ac),
+				RDSService:     rds2.NewFromConfig(ac),
 			})
 			collector := rds.New(ctx, &rds.Config{
 				ScrapeInterval: config.ScrapeInterval,
--- a/pkg/aws/aws.go
+++ b/pkg/aws/aws.go
@@ -137,10 +137,8 @@ func New(ctx context.Context, config *Config) (*AWS, error) {
 			}
 			awsRDSClient := client.NewAWSClient(client.Config{
 				PricingService: awsPricing.NewFromConfig(pricingConfig),
-				EC2Service:     ec2.NewFromConfig(ac),
-				BillingService: costexplorer.NewFromConfig(ac),
+				EC2Service:     ec2.NewFromConfig(ac), //we also need to pass in the ec2 client in order to describe regions, as they are shared between RDS and EC2
 				RDSService:     rds2.NewFromConfig(ac),
-				ELBService:     elbv2.NewFromConfig(ac),
 			})
 			regions, err = awsRDSClient.DescribeRegions(ctx, false)
 			if err != nil {
--- a/pkg/aws/aws.go
+++ b/pkg/aws/aws.go
@@ -18,9 +18,9 @@ import (
 	awsPricing "github.com/aws/aws-sdk-go-v2/service/pricing"
 	rds2 "github.com/aws/aws-sdk-go-v2/service/rds"
 	"github.com/aws/aws-sdk-go-v2/service/sts"
-	"github.com/grafana/cloudcost-exporter/pkg/aws/client"
 	"github.com/prometheus/client_golang/prometheus"
 
+	"github.com/grafana/cloudcost-exporter/pkg/aws/client"
 	ec2Collector "github.com/grafana/cloudcost-exporter/pkg/aws/ec2"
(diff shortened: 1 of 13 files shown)


pkg/aws/client/pricing_test.go
```diff
@@ -302,3 +303,91 @@ func TestListStoragePrices(t *testing.T) {
 		})
 	}
 }
+
+func Test_GetRDSUnitData(t *testing.T) {
+	tests := []struct {
+		name string
+		// priceList awsPricing.GetProductsOutput
```
hairyhenderson: should this line be deleted?

leonorfmartins: yes 🙈 thanks!

hairyhenderson (approved):
LGTM, though I'm not super-familiar with this code so you'll probably want a second review too!

jjo (approved):
LGTM!, tho I'd like a final review/check from @nikimanoledaki 

Smoke-tested:
* running for >1hr w/no crashes
* numbers matching:
  * AWS console
  * `promql-cli query -c ".scrape http://localhost:8080/metrics ^cl; .pinat now"` (from [jjo/promql-cli](https://github.com/jjo/promql-cli))

pkg/aws/client/aws_client.go
```diff
@@ -27,7 +28,7 @@ type AWSClient struct {
 	priceService   *pricing
 	computeService *compute
 	billing        *billing
-	rdsClient      *rds.Client
+	rdsClient      *rdsClient
```
nikimanoledaki: We should rename this to maintain consistency with the other services, like this:
```suggestion
	rdsService      *rds
```

pkg/aws/aws.go
```diff
@@ -130,14 +126,23 @@ func New(ctx context.Context, config *Config) (*AWS, error) {
 			})
 			collectors = append(collectors, collector)
 		case serviceRDS:
-			_ = rds.New(&rds.Config{
+			// pricing API for RDS client needs to use always the same region
+			// as for RDS , the pricing data is only available in the us-east-1
+			pricingConfig, err := createAWSConfig(ctx, "us-east-1", config.Profile, config.RoleARN)
```
nikimanoledaki, line 131: Note - should limiting to `us-east-1` be done here in the AWS config? Or would it be better to set the region that is passed to the `rdsCollector.Config`? 👀 

I think the region in the AWS config might be the region with which cloudcost-exporter authenticates with the AWS API, tied to the authentication method of the cloudcost-exporter. I haven't had time to verify this, but wanted to leave a note that this should be double checked 👀

nikimanoledaki (commented):
Left a note for something that should be checked before we merge this 👀

pkg/aws/aws.go
leonorfmartins: from what I understand, the `pricingAPI` only works with one region but the RDS / EC2 API needs a client / region mapping to be able to list all resources in a region. That's why this part of the code is more convoluted as we can't reuse existing regions on the pricing API :/ 

Not sure if this answers your question!

PR #640, opened 2025-10-10
feat: Add VPC cost metrics for AWS

simaarfania:
Part of https://github.com/grafana/deployment_tools/issues/319890. 

This PR adds pricing metrics for AWS VPC networking components by integrating with the AWS Pricing API. The VPC collector introduces a custom pricing map that efficiently handles the complexities of VPC pricing, including different endpoint types that have significantly different cost structures. The implementation includes proper region-specific pricing with explicit us-east-1 configuration for pricing API calls (similar to RDS).

Metrics added are:
  - `cloudcost_aws_vpc_endpoint_hourly_rate_usd_per_hour{endpoint_type="standard",region} `- Standard VPC interface endpoints
  - `cloudcost_aws_vpc_endpoint_service_hourly_rate_usd_per_hour{region} `- Service-specific VPC endpoints 
  - `cloudcost_aws_vpc_transit_gateway_hourly_rate_usd_per_hour{region}` - Transit Gateway hourly costs 
  - `cloudcost_aws_vpc_elastic_ip_in_use_hourly_rate_usd_per_hour{region} `- In-use Elastic IP addresses 
  - `cloudcost_aws_vpc_elastic_ip_idle_hourly_rate_usd_per_hour{region}` - Idle Elastic IP addresses

### DIFF
--- a/pkg/aws/vpc/vpc.go
+++ b/pkg/aws/vpc/vpc.go
@@ -20,7 +20,6 @@ const (
 	serviceName = "vpc"
 )
 
-// PriceRefreshInterval defines how often to refresh pricing data from AWS Pricing API
 const PriceRefreshInterval = 24 * time.Hour
 
 var (


pkg/aws/vpc/pricing_map.go
```diff
@@ -0,0 +1,311 @@
...
+	}
+	return pricing, nil
+}
+
+// Refresh fetches and updates pricing data for all regions
+func (pm *VPCPricingMap) Refresh(ctx context.Context, regions []ec2Types.Region, clients map[string]client.Client) error {
+	pm.logger.Info("Refreshing VPC pricing data")
+
+	eg := errgroup.Group{}
+	var mu sync.Mutex
```
rooneyshuman: This lock may be redundant. `SetRegionPricing()` already has a lock, and each `goroutine` updates a different region so there's no contention between them anyway

pkg/aws/vpc/pricing_map.go
```diff
@@ -0,0 +1,311 @@
...
+		"region", region,
+		"vpcEndpoints", len(regionPricing.VPCEndpointRates),
+		"transitGateways", len(regionPricing.TransitGatewayRates),
+		"elasticIPs", len(regionPricing.ElasticIPRates))
+
+	return regionPricing, nil
+}
+
+// GetVPCEndpointHourlyRate returns the hourly rate for standard VPC endpoints in a region
+func (pm *VPCPricingMap) GetVPCEndpointHourlyRate(region string) float64 {
```
rooneyshuman: the functions to `Get` the rates seem to repeat a similar pattern:

1. Get pricing data with error handling
2. Search for specific usage types with string matching
3. Return default value if not found
4. Log warnings for failures

We may be able to deduplicate some of this code and make it DRY-er

pkg/aws/vpc/pricing_map.go
```diff
@@ -0,0 +1,311 @@
...
+	pricing, err := pm.GetRegionPricing(region)
+	if err != nil {
+		pm.logger.Warn("Failed to get VPC endpoint pricing", "region", region, "error", err)
+		return VPCEndpointHourlyRateDefault
+	}
+
+	// Look for standard VPC endpoint hourly rates (VpcEndpoint-Hours)
+	for usageType, rate := range pricing.VPCEndpointRates {
+		if strings.Contains(usageType, "VpcEndpoint-Hours") && !strings.Contains(usageType, "Service") {
```
rooneyshuman: to avoid hardcoded strings in pattern matching, can we define these as constants up top?

eg:
```
const (
	VpcEndpointHours       = "VpcEndpoint-Hours"
)
```

pkg/aws/vpc/vpc.go
```diff
@@ -0,0 +1,199 @@
...
+
+func New(ctx context.Context, config *Config) *Collector {
+	logger := config.Logger.With("logger", serviceName)
+	pricingMap := NewVPCPricingMap(logger)
+
+	// Use the dedicated client with us-east-1 pricing service for all regions
+	clientMap := make(map[string]awsclient.Client)
+	for _, region := range config.Regions {
+		clientMap[*region.RegionName] = config.Client
+	}
```
rooneyshuman: if we're using a dedicated client, do we need to construct a `map`?

pkg/aws/vpc/vpc.go
```diff
@@ -0,0 +1,199 @@
...
+
+	// Initial pricing data load
+	if err := pricingMap.Refresh(ctx, config.Regions, clientMap); err != nil {
```
rooneyshuman: following up on the above, we could rework `Refresh()` to take a single client instead of a client map to perform the pricing check

pkg/aws/vpc/pricing_map.go
```diff
@@ -0,0 +1,311 @@
...
+	eg := errgroup.Group{}
+	var mu sync.Mutex
+
+	for _, region := range regions {
+		regionName := *region.RegionName
+		regionClient, ok := clients[regionName]
+		if !ok {
+			pm.logger.Warn("No client found for region", "region", regionName)
+			continue
+		}
```
rooneyshuman: with the suggestion of passing in the single client since all regions use the same dedicated client, this lookup would not be needed

pkg/aws/vpc/vpc.go
```diff
@@ -0,0 +1,199 @@
...
+	ch <- VPCEndpointHourlyGaugeDesc
+	ch <- VPCEndpointServiceHourlyGaugeDesc
+	ch <- TransitGatewayHourlyGaugeDesc
+	ch <- ElasticIPInUseGaugeDesc
+	ch <- ElasticIPIdleGaugeDesc
+	return nil
+}
+
+// Collect satisfies the provider.Collector interface.
+func (c *Collector) Collect(ch chan<- prometheus.Metric) error {
```
rooneyshuman, line 129: we should add context awareness here - if the scrape is canceled mid-way (e.g., shutdown), it’ll still iterate all regions

pkg/aws/vpc/pricing_map.go
```diff
@@ -0,0 +1,311 @@
...
+
+	ec2Types "github.com/aws/aws-sdk-go-v2/service/ec2/types"
+	pricingTypes "github.com/aws/aws-sdk-go-v2/service/pricing/types"
+	"github.com/grafana/cloudcost-exporter/pkg/aws/client"
+	"golang.org/x/sync/errgroup"
+)
+
+const (
+	// Default hourly rates (fallback values)
+	VPCEndpointHourlyRateDefault    = 0.01  // $0.01 per VPC endpoint per hour
```
leonorfmartins: where do these values come from?

I wonder if it would be better not tos et any default values: in case the API fails, I think it would be better to throw an error (like we do for other services) instead of populating with maybe outdated values. In this case it will be harder for us to know there's an issue with the API and fix it right away. Wdyt?

pkg/aws/vpc/pricing_map.go
```diff
@@ -0,0 +1,311 @@
...
+				}
+
+				price, err := strconv.ParseFloat(priceStr, 64)
+				if err != nil {
+					pm.logger.Warn(pricingErr, "error", "failed to parse price", "price", priceStr, "parse_error", err)
+					continue
+				}
+
+				// Skip zero prices
+				if price == 0 {
```
leonorfmartins: I think cases where price is 0 will be rare, but there might be legit values still, no? Why are we skipping them?

pkg/aws/vpc/pricing_map.go
```diff
@@ -0,0 +1,316 @@
...
+	pricingErr := fmt.Sprintf("error fetching VPC pricing for region %s", region)
+
+	for _, product := range prices {
+		var productInfo vpcProduct
+		if err := json.Unmarshal([]byte(product), &productInfo); err != nil {
+			pm.logger.Warn(pricingErr, "error", "failed to unmarshal pricing product", "unmarshal_error", err)
+			continue
+		}
+
+		usageType := productInfo.Product.Attributes.UsageType
```
leonorfmartins: These subfields can still be unpopulated, maybe it's better to check if they exist so we dont cause a nil pointer deference?

pkg/aws/vpc/pricing_map.go
```diff
@@ -0,0 +1,316 @@
...
+				}
+
+				price, err := strconv.ParseFloat(priceStr, 64)
+				if err != nil {
+					pm.logger.Warn(pricingErr, "error", "failed to parse price", "price", priceStr, "parse_error", err)
+					continue
+				}
+
+				// Skip negative prices
+				if price < 0 {
```
leonorfmartins: I guess this _could_ happen, but sounds weird to have negative prices!... do you know in which situations it could happen?
I'm not saying this should be removed, I'm just curious about the use case

pkg/aws/vpc/pricing_map.go
```diff
@@ -0,0 +1,316 @@
...
+					pm.logger.Warn(pricingErr, "error", "negative price detected", "price", price, "usageType", usageType, "region", region)
+					continue
+				}
+
+				// Categorize by usage type patterns
```
leonorfmartins: nit: I find that a switch statement on `usageType` would be more readable, but it's just a suggestion

pkg/aws/vpc/pricing_map.go
```diff
@@ -0,0 +1,316 @@
...
+				}
+				if !excluded {
+					pm.logger.Info("Selected "+serviceType+" rate", "region", region, "usageType", usageType, "rate", rate, "pattern", pattern)
+					return rate, nil
+				}
+			}
+		}
+	}
+
+	// Try fallback patterns
```
leonorfmartins: I'm not sure if I understand what this fallback pattern is, but I'm guessing they could be the default values that got deleted in [here](https://github.com/grafana/cloudcost-exporter/pull/640/commits/210d804e1e18d22a69fb98c966b6de98e7bb821a) ? If so, this part of the code can go as well, right?

pkg/aws/vpc/pricing_map.go
```diff
@@ -0,0 +1,316 @@
...
+				}
+			}
+		}
+	}
+
+	return 0, fmt.Errorf("no %s rate found for region %s", serviceType, region)
+}
+
+// getRateWithFallback is a generic helper for getting pricing rates
+func (pm *VPCPricingMap) getRateWithFallback(region string, serviceType string, rateMapGetter func(*VPCRegionPricing) map[string]float64, matcher usageTypeMatcher) (float64, error) {
```
leonorfmartins: If I'm assuming correctly on the comment above, `withFallback` sufixes on these functions can also be dropped 👍

pkg/aws/vpc/vpc.go
```diff
@@ -0,0 +1,221 @@
...
+	pricingMap := NewVPCPricingMap(logger)
+
+	// Initial pricing data load using the dedicated us-east-1 pricing client
+	if err := pricingMap.Refresh(ctx, config.Regions, config.Client); err != nil {
+		logger.Error("Failed to load initial VPC pricing data", "error", err)
+	}
+
+	// Set up periodic pricing refresh
+	go func() {
+		ticker := time.NewTicker(24 * time.Hour) // Refresh daily
```
rooneyshuman: let's define the refresh interval as a const like we do for other services ([example](https://github.com/grafana/cloudcost-exporter/blob/d6344abd721789920f0cefb04288d3161f3ad02f/pkg/google/gke/pricing_map.go#L61))

rooneyshuman (approved):
🚀

PR #663, opened 2025-10-21
Add latency log and change RDS metric label

leonorfmartins:
In order to do proper namespace and cluster attribution, we need n unique label which is common with one of the env metrics we have. Name was not 100% reliable, as the DB name varies from the crossplane resource name in most of the cases.

### DIFF
--- a/pkg/aws/rds/rds.go
+++ b/pkg/aws/rds/rds.go
@@ -24,7 +24,7 @@ var (
 		cloudcost_exporter.MetricPrefix,
 		subsystem,
 		"hourly_rate_usd_per_hour",
-		"Hourly cost of NAT Gateway by region. Cost represented in USD/hour",
+		"Hourly cost of AWS RDS instances by region, tier and id. Cost represented in USD/hour",
 		[]string{"region", "tier", "id"},
 	)
 )


pkg/aws/rds/rds.go
```diff
@@ -25,7 +25,7 @@ var (
 		subsystem,
 		"hourly_rate_usd_per_hour",
 		"Hourly cost of NAT Gateway by region. Cost represented in USD/hour",
```
nikimanoledaki:
```suggestion
		"Hourly cost of RDS DBs by region. Cost represented in USD/hour",
```

pkg/aws/rds/rds.go
```diff
@@ -120,9 +121,10 @@ func (c *Collector) Collect(ch chan<- prometheus.Metric) error {
 			hourlyPrice,
 			region,
 			*instance.DBInstanceClass,
-			*instance.DBInstanceIdentifier,
+			*instance.DbiResourceId,
 		)
 	}
+	logger.Info("Finished collect", "subsystem", subsystem, "duration", time.Since(start))
```
nikimanoledaki: This looks like a duplicate of `cloudcost_exporter_metrics_handler_request_duration_seconds_bucket` which we use for the Scrape Latency SLO 👀

nikimanoledaki: We use this in other services (VPC, NAT Gateway, EC2, AKS) as well though 🤔

leonorfmartins: yeah, I was about to comment about it... I can remove this log for now and we can revisit these logs in an upcoming PR, since they are present in more services?

nikimanoledaki: Let's do that 👍

pkg/aws/rds/rds.go
```diff
@@ -24,8 +24,8 @@ var (
 		cloudcost_exporter.MetricPrefix,
 		subsystem,
 		"hourly_rate_usd_per_hour",
-		"Hourly cost of NAT Gateway by region. Cost represented in USD/hour",
-		[]string{"region", "tier", "name"},
+		"Hourly cost of AWS RDS instances by region, tier and id. Cost represented in USD/hour",
+		[]string{"region", "tier", "id"},
```
nikimanoledaki, line 28: `resource_id` would help so that it's not confused with `instance_id` - WDYT?

```suggestion
		"Hourly cost of AWS RDS instances by region, tier and DBI resource ID. Cost represented in USD/hour",
		[]string{"region", "tier", "resource_id"},
```

nikimanoledaki: Unless we want to keep `id` to be the same as the crossplane label, in which case we can just update the description:
```suggestion
		"Hourly cost of AWS RDS instances by region, tier and DBI resource ID. Cost represented in USD/hour",
		[]string{"region", "tier", "id"},
```

nikimanoledaki:
It was just a nit documentation thing - we can keep moving 👍

Issue #274, opened 2024-08-13
Use Tickers to populate EC2 pricing maps


paulajulve:
Follow up to https://github.com/grafana/cloudcost-exporter/issues/235

Currently these are populated within the main routine in the Collect method on the `ec2` package. It means that every time the exporter goes to collect the metrics, it stops to first check if the compute one is outdated, if it is, populate it, and then the same for the storage one, before moving on to actually collecting the metrics that we want to expose.

This responsibility should not belong to the Collect method. Instead, we could follow [the same pattern as the `aks` package](https://github.com/grafana/cloudcost-exporter/blob/983ce7758230709cfb70ccdf3134395b86edb257/pkg/azure/aks/aks.go#L105-L127):
- create the maps within the `New` function instead of the `Collect` one
- create tickers for each
- delegate population to the methods of the map structs themselves
- spin up a background go routine for them to re-populate in the background whenever they expire, independent of each other and of the `Collect` function

suntala:
closed by https://github.com/grafana/cloudcost-exporter/pull/664

PR #664, opened 2025-10-22
EC2: Add tickers to populate pricing maps

suntala:
For https://github.com/grafana/cloudcost-exporter/issues/274

Where do the collected metrics get consumed? Would be great to check that everything continues to work as expected after the inclusion of tickers. Found this: [docs/contribute/developer-guide.md](https://github.com/grafana/cloudcost-exporter/blob/983ce7758230709cfb70ccdf3134395b86edb257/docs/contribute/developer-guide.md)

Notes:
* [Synchronous pricing map initialization](https://github.com/grafana/cloudcost-exporter/pull/664/files#diff-e31bc93bf13d48b81b2da61a67eaec7bdd0630e52887f523dc034eef95f85815R91-R96) takes care of the initialisation that was taking place [in Collect before](https://github.com/grafana/cloudcost-exporter/pull/664/files#diff-e31bc93bf13d48b81b2da61a67eaec7bdd0630e52887f523dc034eef95f85815L111).
* We now [clear the map before generating it](https://github.com/grafana/cloudcost-exporter/pull/664/files#diff-28c56763f93edc53a3a1b90d2e72d51e6d23da1fb631cf6a7b14311b1774053bR117-R122) because previously the [maps were getting created anew each time](https://github.com/grafana/cloudcost-exporter/pull/664/files).

Testing changes:
*  TestNew was added
* [Checking client presence](https://github.com/grafana/cloudcost-exporter/pull/664/files#diff-30e0dcbef190e9ab58c39cd1943eda8b2e2da8fb35408aa1e4514fb510b51a43L79) is now happening in TestNew. Other test cases which no longer apply have been removed or moved ("Collect should return an error if GenerateComputePricingMap returns an error", "Unparsable String into json")
* Test_PopulateStoragePricingMap was removed and instead the test cases were added to TestStoragePricingMap_GenerateStoragePricingMap
* mock_client_test.go was added to provide a mock with explicit behaviour.
* In TestComputePricingMap_GenerateComputePricingMap, the ["Just spot prices as input" test case was added](https://github.com/grafana/cloudcost-exporter/pull/664/files#diff-065393c179ccf17a639da31c2308ce8e50a1af364b59f4d212e432eaaa88aa7bR116) along with "Returns error when ListSpotPrices fails" and "Returns error when ListOnDemandPrices fails"
* Two new test cases [TestStoragePricingMap_GenerateStoragePricingMap](https://github.com/grafana/cloudcost-exporter/pull/664/files#diff-065393c179ccf17a639da31c2308ce8e50a1af364b59f4d212e432eaaa88aa7bR251-R262): "Returns error when parsing invalid JSON" and "Returns error when ListStoragePrices fails"
* TestStoragePricingMap_GenerateStoragePricingMap_Refresh and TestComputePricingMap_GenerateComputePricingMap_Refresh are added to ensure maps are refreshed.

Metrics comparison between main branch vs this branch: https://raintank-corp.slack.com/archives/C07SUF6SESF/p1763633139580359?thread_ts=1760716400.800529&cid=C07SUF6SESF
The metrics captured here regarding ec2 are identical in both branches.

I ran 

```
go run cmd/exporter/exporter.go \
  -provider aws \
  -aws.profile $AWS_PROFILE \
  -aws.region us-east-1 \
  -aws.services ec2
  ``` 
  with management profile: https://raintank-corp.slack.com/archives/C07SUF6SESF/p1763633302475329?thread_ts=1760716400.800529&cid=C07SUF6SESF
  
  Then did 
  ```
  curl http://localhost:8080/metrics | grep ec2
```

### DIFF
--- a/pkg/aws/aws.go
+++ b/pkg/aws/aws.go
@@ -138,7 +138,10 @@ func newWithDependencies(ctx context.Context, config *Config, awsClient client.C
 				RegionMap:      regionClients,
 			})
 			if err != nil {
-				return nil, err
+				logger.LogAttrs(ctx, slog.LevelError, "Error creating collector",
+					slog.String("service", service),
+					slog.String("message", err.Error()))
+				continue
 			}
 			collectors = append(collectors, collector)
 		case serviceRDS:


pkg/aws/ec2/ec2.go
```diff
@@ -82,17 +79,55 @@ type Config struct {
...
+
+	// Initial population so that Collect can use the maps
+	if err := computeMap.GenerateComputePricingMap(ctx); err != nil {
+		return nil, fmt.Errorf("failed initial compute pricing: %w", err)
+	}
+	if err := storageMap.GenerateStoragePricingMap(ctx); err != nil {
+		return nil, fmt.Errorf("failed initial storage pricing: %w", err)
+	}
+
+	go func(ctx context.Context) {
```
leonorfmartins: I wonder if these 2 go funcs could be merged into one and just have one more case? I guess the ctx.Done will be the same signal for both storage and compute, right?

suntala: Thanks for taking a look! The ticket description mentions "create tickers for each", which makes sense since they should be "independent of each other and of the Collect function".

pkg/aws/ec2/pricing_map.go
```diff
@@ -187,6 +212,79 @@ func (spm *StoragePricingMap) GenerateStoragePricingMap(storagePrices []string)
...
+
+func (spm *StoragePricingMap) fetchPricing(ctx context.Context) ([]string, error) {
+	var storagePrices []string
+	eg, ctx := errgroup.WithContext(ctx)
+	eg.SetLimit(errGroupLimit)
+	m := sync.Mutex{}
+	for _, region := range spm.cfgRegions {
+		region := region
+		eg.Go(func() error {
+			regionClient, ok := spm.regionMap[*region.RegionName]
```
leonorfmartins, line 277: I'm aware this code already existed, but do you know why is it needed to fetch the region client in a go routine as well? It looks like it could still be done outside of it and then we wouldn't have to worry about concurrently accessing the region.

suntala: Good question. I'm not sure about why it's included in the go routine. I'd like to keep as much of the old logic as possible for this PR though to keep it focused on the tickers refactor. It could be a good improvement to add separately.

pkg/aws/ec2/ec2.go
leonorfmartins: right, but this go func doesn't create a ticker in itself, it waits for a ticker signal and reacts to it (or waits for a ctx.Done). So, we could combine both in the same go routine to wait for either ticker instead of repeating the function?

suntala: The pattern referenced in the ticket has them in separate go routines: https://github.com/grafana/cloudcost-exporter/blob/983ce7758230709cfb70ccdf3134395b86edb257/pkg/azure/aks/aks.go#L105-L127

I think keeping them separate promotes them being fully independent of each other. Was there a reason for combining them besides not repeating the function?

pkg/aws/ec2/ec2.go
```diff
@@ -82,17 +79,55 @@ type Config struct {
...
 
 // New creates an ec2 collector
-func New(config *Config) *Collector {
+func New(ctx context.Context, config *Config) (*Collector, error) {
 	logger := config.Logger.With("logger", "ec2")
+	computeMap := NewComputePricingMap(logger, config)
+	storageMap := NewStoragePricingMap(logger, config)
+
+	priceTicker := time.NewTicker(config.ScrapeInterval)
+	machineTicker := time.NewTicker(config.ScrapeInterval)
```
rooneyshuman: these are both pricing tickers for two different resources, compute and storage. Let's rename them for clarity
```suggestion
	computeTicker := time.NewTicker(config.ScrapeInterval)
	storageTicker := time.NewTicker(config.ScrapeInterval)
```

pkg/aws/ec2/ec2.go
rooneyshuman: I think I agree with keeping the 2 goroutines as written. We get fault isolation this way - if `GenerateComputePricingMap` blocks or takes longer, `GenerateStoragePricingMap` still runs on schedule

rooneyshuman: nit: we don't need to explicitly pass `ctx` as a parameter. It is implicitly available to the goroutine from the outer scope - this can be applied to both goroutines
```suggestion
	go func() {
```

pkg/aws/ec2/pricing_map.go
```diff
@@ -65,42 +70,60 @@ type StoragePricingMap struct {
...
+// GenerateComputePricingMap fetches spot and ondemand prices from AWS using the
+// configured region clients. It then needs to:
 // 1. Parse out the ondemand prices and generate a productTerm map for each instance type
 // 2. Parse out spot prices and use the productTerm map to generate a spot price map
-func (cpm *ComputePricingMap) GenerateComputePricingMap(ondemandPrices []string, spotPrices []ec2Types.SpotPrice) error {
+func (cpm *ComputePricingMap) GenerateComputePricingMap(ctx context.Context) error {
+	cpm.logger.LogAttrs(ctx, slog.LevelInfo, "Refreshing compute pricing map")
+	ondemandPrices, spotPrices, err := cpm.fetchPricing(ctx)
+	if err != nil {
+		return err
```
rooneyshuman, line 115: can we log the error here?

pkg/aws/ec2/pricing_map.go
rooneyshuman: we should wrap this `InstanceDetails` access in an RLock before reading from it
```suggestion
		cpm.m.RLock()
		_, ok := cpm.InstanceDetails[instanceType]
		cpm.m.RUnlock()
		if !ok {
```

pkg/aws/ec2/ec2.go
suntala: Good catch. Forgot to update these names. Thanks for the review!

pkg/aws/ec2/pricing_map.go
suntala: Implemented in [3597469](https://github.com/grafana/cloudcost-exporter/pull/664/commits/3597469a6ea16d63f3bb9cebbb3a6bc7cc1148c1)

pkg/aws/ec2/ec2.go
suntala: Makes sense. Implemented here: [f22ce31](https://github.com/grafana/cloudcost-exporter/pull/664/commits/f22ce31eda44be9c077bef9f594bd26ca0043cb0)

pkg/aws/ec2/pricing_map.go
suntala: Thanks! Implemented in [96782ca](https://github.com/grafana/cloudcost-exporter/pull/664/commits/96782ca387df6bce9594221d24f29007741c767d)

pkg/aws/aws.go
```diff
@@ -131,12 +131,15 @@ func newWithDependencies(ctx context.Context, config *Config, awsClient client.C
...
-			collector := ec2Collector.New(&ec2Collector.Config{
+			collector, err := ec2Collector.New(ctx, &ec2Collector.Config{
 				Regions:        regions,
 				Logger:         logger,
 				ScrapeInterval: config.ScrapeInterval,
 				RegionMap:      regionClients,
 			})
+			if err != nil {
+				return nil, err
+			}
```
nikimanoledaki, line 145: We shouldn't return an error for the provider if one of the collectors returns an error. This is by design so that one collector's error does not take down other collectors. 

The error can be logged instead, which is what we do for [GCP collectors](https://github.com/grafana/cloudcost-exporter/blob/9d1b44202dfe91daaa1ea5ebd77c4ef1f884273f/pkg/google/gcp.go#L115C1-L120C5):
```suggestion
			if err != nil {
				logger.LogAttrs(ctx, slog.LevelError, "Error creating collector",
					slog.String("service", service),
					slog.String("message", err.Error()))
				continue
			}
```

In a follow-up PR, we should add this for the rest of the AWS collectors as well.

suntala: Thanks for the review. I added an issue so we don't lose track of this: https://github.com/grafana/cloudcost-exporter/issues/716

nikimanoledaki: thank you @suntala!!

rooneyshuman (approved):

PR #673, opened 2025-10-28
Collect GCP cloudSQL hourly cost metrics

leonorfmartins:
## Description
Price calculation logic:
1. If the instance tier matches the custom format (db-custom-{vCPU}-{RAM}), use custom pricing logic:
- Find CPU and RAM component SKUs in the region (usage_unit "h" for CPU, "GiBy" for RAM)
- Calculate total price: (vCPU count × CPU price per hour) + (RAM in GB × RAM price per GB per hour)
2. If the instance is a standard instance (any other tier format), use standard pricing logic:  
- Find a SKU matching the region and tier type:
   * For micro/small tiers (db-f1-micro, db-g1-small): match by tier name in SKU description
   * For standard tiers (db-n1-standard-*, etc.): match by region only (first matching SKU)
 - Return the price from the matched SKU

## Test
Metrics will look like this:
```
# TYPE cloudcost_gcp_cloudsql_cost_usd_per_hour gauge
cloudcost_gcp_cloudsql_cost_usd_per_hour{instance="grafanalabs-dev:europe-west1:dev-eu-west-3-grafana-com-dev-replica-1",region="europe-west1",sku="custom"} 0.16625
cloudcost_gcp_cloudsql_cost_usd_per_hour{instance="grafanalabs-dev:europe-west1:dev-eu-west-3-grafana-com-shared-microservices-dev-1-replica-2",region="europe-west1",sku="custom"} 0.16625
cloudcost_gcp_cloudsql_cost_usd_per_hour{instance="grafanalabs-dev:us-central1:asserts-dev-us-central-0",region="us-central1",sku="158E-8B8E-6C76"} 0.035
cloudcost_gcp_cloudsql_cost_usd_per_hour{instance="grafanalabs-dev:us-central1:bq-sync-mysql",region="us-central1",sku="custom"} 0.784
cloudcost_gcp_cloudsql_cost_usd_per_hour{instance="grafanalabs-dev:us-central1:codecov-dev-us-central-0",region="us-central1",sku="custom"} 0.6125
cloudcost_gcp_cloudsql_cost_usd_per_hour{instance="grafanalabs-dev:us-central1:cortex-dev-01-adaptive-metrics-api",region="us-central1",sku="158E-8B8E-6C76"} 0.035
```

## Attribution
We can join with [this metric](https://ops.grafana-ops.net/goto/cf434nqpgay9sd?orgId=1) to get cluster / namespace attribution. It will work for DBs provisioned in TF and crossplane.

Fixes https://github.com/grafana/deployment_tools/issues/398889

### DIFF
--- a/docs/metrics/gcp/cloudsql.md
+++ b/docs/metrics/gcp/cloudsql.md
@@ -1,0 +1,6 @@
+# GCP CloudSQL Metrics
+
+| Metric name                                    | Metric type | Description                                                                                 | Labels                                                                                                                                                                                                                                                                    |
+|------------------------------------------------|-------------|---------------------------------------------------------------------------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| cloudcost_gcp_cloudsql_cost_usd_per_hour      | Gauge       | Hourly cost of GCP CloudSQL instances by instance name, region and sku. Cost represented in USD/hour | `instance`=&lt;CloudSQL instance connection name&gt; <br/> `region`=&lt;GCP region where the instance is located&gt; <br/> `sku`=&lt;GCP billing SKU identifier or "custom" for custom machine types&gt; |
+


pkg/google/client/sqladmin.go
```diff
@@ -0,0 +1,28 @@
+package client
+
+import (
+	"context"
+
+	sqladmin "google.golang.org/api/sqladmin/v1beta4"
```
hairyhenderson, line 6: I'm curious why `v1beta4` and not `v1`?

pkg/google/gcp.go
```diff
@@ -128,6 +129,17 @@ func New(config *Config) (*GCP, error) {
 					slog.String("message", err.Error()))
 				continue
 			}
+		case "SQL":
+			collector, err = cloudsql.New(&cloudsql.Config{
+				Projects:       config.Projects,
+				ScrapeInterval: config.ScrapeInterval,
+			}, gcpClient)
```
nikimanoledaki, line 139: It looks like `Logger:         config.Logger,` is missing here (for consistency)

pkg/google/client/billing.go
```diff
@@ -127,8 +127,8 @@ func getPriceFromSku(sku *billingpb.Sku) (float64, error) {
 		return 0.0, fmt.Errorf("found sku without TieredRates: %+v", sku)
 	}
 	tierRate := priceInfo.PricingExpression.TieredRates[tierRatesLen-1]
-
-	return 1e-9 * float64(tierRate.UnitPrice.Nanos), nil // Convert NanoUSD to USD when return
+	// The cost of the SKU is units + nanos
+	return float64(tierRate.UnitPrice.Units) + 1e-9*float64(tierRate.UnitPrice.Nanos), nil // Convert NanoUSD to USD when return
```
nikimanoledaki, line 131: Question here - does this affect the cost of other services as well? Were they wrong before this change, and has the cost of the SKU been validated?

pkg/google/cloudsql/cloudsql.go
```diff
@@ -0,0 +1,149 @@
...
+	projects   []string
+	gcpClient  client.Client
+	config     *Config
+	pricingMap *pricingMap
+	logger     *slog.Logger
+}
+
+const (
+	subsystem           = "gcp_cloudsql"
+	CostRefreshInterval = 24 * 30 * time.Hour
```
nikimanoledaki: We refresh `24 * time.Hour` in other services - wondering why this is different here?

pkg/google/cloudsql/cloudsql.go
```diff
@@ -0,0 +1,149 @@
...
+func (c *Collector) CollectMetrics(ch chan<- prometheus.Metric) float64 {
+	return 0
+}
+
+func (c *Collector) getAllCloudSQL(ctx context.Context) ([]*sqladmin.DatabaseInstance, error) {
+	var allCloudSQLInfo = []*sqladmin.DatabaseInstance{}
+	seenInstances := make(map[string]bool)
+
+	for _, project := range c.projects {
+		cloudSQLInstances, err := c.gcpClient.ListSQLInstances(ctx, project)
```
nikimanoledaki, line 122: Should we be processing by region here in a similar way to what we do for forwarding rules [here](https://github.com/grafana/cloudcost-exporter/blob/06161bbcbfcd3d8b6371c75dc888a646820f247b/pkg/google/networking/forwarding_rule.go#L172)? 

The `instance.Region` contains the region, so this would be needed if (or when, as we scale) the scrape time takes too long for this collector and times out, which would lead to data loss. It could take longer especially since we're processing projects sequentially. Potentially not needed yet if the scrape time is under 10s!

pkg/google/cloudsql/cloudsql.go
```diff
@@ -0,0 +1,149 @@
...
+		if err != nil {
+			c.logger.Error("error listing sql instances for project", "project", project, "error", err)
+			return nil, err
+		}
+		for _, instance := range cloudSQLInstances {
+			if instance.ConnectionName != "" {
```
nikimanoledaki, line 128: Could there be a case where `instance.ConnectionName == ""` that we need to handle? We could move the `append` with this conditional to be sure that it exists. 

Additionally, do we need to validate that `instance.Region` exists, before we add it as a label? Maybe this is always present but good to double check :)

pkg/google/cloudsql/cloudsql_test.go
```diff
@@ -0,0 +1,284 @@
...
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+		},
+		// {
+		// 	name: "custom pricing",
```
nikimanoledaki: This test case is commented out 👀

pkg/google/cloudsql/pricingmap.go
```diff
@@ -0,0 +1,433 @@
...
+func priceForSKU(sku *billingpb.Sku) (float64, bool) {
+	if sku == nil {
+		return 0, false
+	}
+	if len(sku.PricingInfo) == 0 {
+		return 0, false
+	}
+	if len(sku.PricingInfo[0].PricingExpression.TieredRates) == 0 {
+		return 0, false
+	}
```
nikimanoledaki: Is the price for the SKU really 0 or was it not found so we should be returning an error? 🤔

pkg/google/client/billing.go
leonorfmartins: I validated the cost of the sku for cloudSQL and I believe this was wrong in the past: the unit price is of type money (in docs here: https://docs.cloud.google.com/go/docs/reference/cloud.google.com/go/billing/latest/apiv1/billingpb#cloud_google_com_go_billing_apiv1_billingpb_PricingExpression_TierRate), which has a nano and a unit field, according to the type definition here: https://pkg.go.dev/google.golang.org/genproto/googleapis/type/money#Money.
From the docs:
> 	// For example $-1.75 is represented as `units`=-1 and `nanos`=-750,000,000.

pkg/google/client/sqladmin.go
leonorfmartins: no particular reason, I just found more examples for that version. According to the `important` banner https://cloud.google.com/sql/docs/mysql/admin-api/, v1 is being made to be compatible with v1 and there isn't any need to upgrade it to v1 just now.

pkg/google/cloudsql/cloudsql.go
leonorfmartins: good point, we should still refresh it every day in case there are new DBs!

pkg/google/cloudsql/cloudsql.go
leonorfmartins: locally for `grafanalabs-global` is takes ~2s, which I think it's acceptable? What do you think changing this part if we start hitting timeouts?
This reminds me... we should revisit our SLO on this app and see if it takes into consideration these cases, it totally should IMO

pkg/google/cloudsql/cloudsql.go
leonorfmartins: this is to prevent listing repeated instances. The fastest way I found to check if it exists was to use a key:value pair instead of having to go through the all array to check. I'm not sure if this was what you're suggesting?
I don't think `connectionName` will ever be empty and the same goes for region, but I guess better safe than sorry! Would you advise to remove it still?

pkg/google/cloudsql/cloudsql_test.go
leonorfmartins: 🤦🏻‍♀️

PR #690, opened 2025-11-04
ci: multi-arch docker builds

ricky-undeadcoders:
This pull request significantly refactors `.github/workflows/build-and-deploy.yml` to introduce multi-arch docker builds, as well as a few other changes. 

The changes include:
- the adoption of a multi-architecture build matrix to build `arm` and `amd` images on the GitHub hosted `ubuntu-24.04` and `ubuntu-24.04-arm` runners
- a switch from `build-push-to-dockerhub` to `docker-build-push-image` (a requirement for multi-arch builds)
- an updated trigger so the workflow runs for all push events, but only pushes Docker images on `main` and for new versions. This helps to more quickly identify breaking changes to the build process.
- switched from `ubuntu-latest` to `ubuntu-24.04` runners to match the arm runners (`ubuntu-latest-arm` isn't a thing unfortunately).

Here is an [example of the new workflow.](https://github.com/grafana/cloudcost-exporter/actions/runs/19184386450)

To verify full functionality I setup a [test pipeline](https://github.com/grafana/cloudcost-exporter/actions/runs/19182656686) that also [pushed an image to DockerHub](https://hub.docker.com/repository/docker/grafana/cloudcost-exporter/tags/multi-arch-test/sha256:ebd799fc0fd529334496bcf02e319db1e510d19d949d631e09de787291b829fe). I verified that that image could be pulled and that its labels matched `grafana/cloudcost-exporter:latest`.

Related to https://github.com/grafana/shared-workflows/issues/45

### DIFF
--- a/.github/workflows/build-and-deploy.yml
+++ b/.github/workflows/build-and-deploy.yml
@@ -30,6 +30,11 @@ jobs:
 
     runs-on: ${{ matrix.runnerType }}
     steps:
+      - name: Checkout
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        with:
+          persist-credentials: false
+
       - name: Should we push?
         id: setup
         run: |
@@ -38,10 +43,6 @@ jobs:
           else
             echo "should-push=false" | tee -a "$GITHUB_OUTPUT"
           fi
-
-      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
-        with:
-          persist-credentials: false
 
       - name: Build and Push Image
         id: build


.github/workflows/build-and-deploy.yml
```diff
@@ -1,50 +1,85 @@
...
 name: Build, Push, and Deploy Image
 
-on:
-  push:
-    branches:
-      - main
-    tags:
-      - v*
-
-permissions:
```
suntala, line 11: Should we keep the workflow-level restrictions so that we don't risk future jobs using default permissions?

ricky-undeadcoders: I would argue "no" for 2 reasons.
1. This workflow is scoped to building and deploying docker images, so it's unlikely that anything will be added.
2. For our current workflow it's completely unnecessary since we override it at the job level, so the only reason for its inclusion would be for future proofing.

It's small so I can absolutely add it back if you would like... but IMO it's cleaner from a code and a security standpoint if we leave it out.

Issue #209, opened 2024-06-24
Setup global context with timeout


Pokom:
To safe guard `cloudcost-exporter` from initiating too many external requests that do not get cancelled, there should be a uniform way of passing a Context that has a [CancelFunc](https://pkg.go.dev/context#WithTimeout) defined with a sane default. This should start at `exporter.go` as a parameter with a `1m` default set and then passed to each provider that gets created. The provider can then generate a context with a timeout defined on each `Collect` call and pass this on to the collectors `Collect` methods. 

To achieve this, the collector interface needs to be refactored to have the `Collect` method accept a context as well.

Pokom:
@Pokom and @logyball will follow up with a design.

PR #699, opened 2025-11-17
Add uniform context handling with timeout for all collectors

rooneyshuman:
### Summary
Adds uniform context handling across all collectors to prevent long-running external requests. A context with a 1-minute default timeout is created at the exporter level and passed through providers to collectors.

### Changes in this PR

- Collector interface: `Collect()` now accepts `context.Context` as the first parameter
- Exporter: Creates context with 1-minute default timeout and passes it to providers
- Providers: Create timeout contexts on each Collect() call and pass them to collectors
- All 11 collectors: Updated to accept and use context parameter
- Standardized pattern to ensure the context cancellation hierarchy works as expected:
  - Collectors with background `goroutines` receive `ctx` in `New()` (VPC, ForwardingRule, GKE, AKS, NATGateway)
  - Collectors without background ops don't receive `ctx` in `New()` (EC2, S3, RDS, ELB, GCS)

### Benefits
- Prevents runaway external requests via timeout enforcement
- Consistent cancellation hierarchy: exporter → provider → collector
- All external API calls respect the timeout context
- Background operations can be cancelled from provider level

### Testing
- All existing tests updated and passing
```
❯ go test ./...
?       github.com/grafana/cloudcost-exporter   [no test files]
?       github.com/grafana/cloudcost-exporter/cloudcost-exporter-dashboards/convertor   [no test files]
?       github.com/grafana/cloudcost-exporter/cloudcost-exporter-dashboards/dashboards  [no test files]
ok      github.com/grafana/cloudcost-exporter/cmd/dashboards    (cached)
?       github.com/grafana/cloudcost-exporter/cmd/exporter      [no test files]
ok      github.com/grafana/cloudcost-exporter/cmd/exporter/config       (cached)
ok      github.com/grafana/cloudcost-exporter/cmd/exporter/web  (cached)
?       github.com/grafana/cloudcost-exporter/mocks/pkg/aws/services/elbv2      [no test files]
ok      github.com/grafana/cloudcost-exporter/pkg/aws   (cached)
ok      github.com/grafana/cloudcost-exporter/pkg/aws/client    (cached)
?       github.com/grafana/cloudcost-exporter/pkg/aws/client/mocks      [no test files]
ok      github.com/grafana/cloudcost-exporter/pkg/aws/ec2       (cached)
ok      github.com/grafana/cloudcost-exporter/pkg/aws/elb       (cached)
ok      github.com/grafana/cloudcost-exporter/pkg/aws/natgateway        (cached)
ok      github.com/grafana/cloudcost-exporter/pkg/aws/pricingstore      (cached)
ok      github.com/grafana/cloudcost-exporter/pkg/aws/rds       (cached)
ok      github.com/grafana/cloudcost-exporter/pkg/aws/s3        (cached)
?       github.com/grafana/cloudcost-exporter/pkg/aws/services  [no test files]
?       github.com/grafana/cloudcost-exporter/pkg/aws/services/costexplorer     [no test files]
?       github.com/grafana/cloudcost-exporter/pkg/aws/services/ec2      [no test files]
?       github.com/grafana/cloudcost-exporter/pkg/aws/services/elbv2    [no test files]
?       github.com/grafana/cloudcost-exporter/pkg/aws/services/mocks    [no test files]
?       github.com/grafana/cloudcost-exporter/pkg/aws/services/pricing  [no test files]
?       github.com/grafana/cloudcost-exporter/pkg/aws/services/vpc      [no test files]
ok      github.com/grafana/cloudcost-exporter/pkg/aws/vpc       (cached)
ok      github.com/grafana/cloudcost-exporter/pkg/azure (cached)
ok      github.com/grafana/cloudcost-exporter/pkg/azure/aks     (cached)
?       github.com/grafana/cloudcost-exporter/pkg/azure/client  [no test files]
?       github.com/grafana/cloudcost-exporter/pkg/azure/client/mocks    [no test files]
ok      github.com/grafana/cloudcost-exporter/pkg/google        (cached)
ok      github.com/grafana/cloudcost-exporter/pkg/google/client (cached)
ok      github.com/grafana/cloudcost-exporter/pkg/google/client/cache   (cached)
?       github.com/grafana/cloudcost-exporter/pkg/google/client/mocks   [no test files]
ok      github.com/grafana/cloudcost-exporter/pkg/google/gcs    (cached)
ok      github.com/grafana/cloudcost-exporter/pkg/google/gke    0.725s
?       github.com/grafana/cloudcost-exporter/pkg/google/metrics        [no test files]
ok      github.com/grafana/cloudcost-exporter/pkg/google/networking     (cached)
ok      github.com/grafana/cloudcost-exporter/pkg/google/vpc    (cached)
?       github.com/grafana/cloudcost-exporter/pkg/logger        [no test files]
?       github.com/grafana/cloudcost-exporter/pkg/provider      [no test files]
?       github.com/grafana/cloudcost-exporter/pkg/provider/mocks        [no test files]
ok      github.com/grafana/cloudcost-exporter/pkg/utils (cached)
?       github.com/grafana/cloudcost-exporter/scripts/aws-spot-pricing  [no test files]
?       github.com/grafana/cloudcost-exporter/scripts/gcp-fetch-skus    [no test files]
```
- Ran the exporter with a short timeout setting and observed long-running operations get cancelled as expected:
image

### Issue Ref
- https://github.com/grafana/cloudcost-exporter/issues/209

### DIFF
--- a/pkg/aws/ec2/ec2_test.go
+++ b/pkg/aws/ec2/ec2_test.go
@@ -45,7 +45,7 @@ func TestCollector_Collect(t *testing.T) {
 		})
 		ch := make(chan prometheus.Metric)
 		go func() {
-			err := collector.Collect(context.Background(), ch)
+			err := collector.Collect(t.Context(), ch)
 			close(ch)
 			assert.NoError(t, err)
 		}()
@@ -72,7 +72,7 @@ func TestCollector_Collect(t *testing.T) {
 			},
 		})
 		ch := make(chan prometheus.Metric)
-		err := collector.Collect(context.Background(), ch)
+		err := collector.Collect(t.Context(), ch)
 		close(ch)
 		assert.Error(t, err)
 	})
@@ -83,7 +83,7 @@ func TestCollector_Collect(t *testing.T) {
 			RegionMap: map[string]client.Client{},
 		})
 		ch := make(chan prometheus.Metric)
-		err := collector.Collect(context.Background(), ch)
+		err := collector.Collect(t.Context(), ch)
 		close(ch)
 		assert.ErrorIs(t, err, ErrClientNotFound)
 	})
@@ -102,7 +102,7 @@ func TestCollector_Collect(t *testing.T) {
 			},
 		})
 		ch := make(chan prometheus.Metric)
-		err := collector.Collect(context.Background(), ch)
+		err := collector.Collect(t.Context(), ch)
 		close(ch)
 		assert.ErrorIs(t, err, ErrListSpotPrices)
 	})
@@ -137,7 +137,7 @@ func TestCollector_Collect(t *testing.T) {
 		})
 		ch := make(chan prometheus.Metric)
 		defer close(ch)
-		assert.ErrorIs(t, collector.Collect(context.Background(), ch), ErrGeneratePricingMap)
+		assert.ErrorIs(t, collector.Collect(t.Context(), ch), ErrGeneratePricingMap)
 	})
 	t.Run("Test cpu, memory and total cost metrics emitted for each valid instance", func(t *testing.T) {
 		c := mock_client.NewMockClient(ctrl)
@@ -236,7 +236,7 @@ func TestCollector_Collect(t *testing.T) {
 
 		ch := make(chan prometheus.Metric)
 		go func() {
-			if err := collector.Collect(context.Background(), ch); err != nil {
+			if err := collector.Collect(t.Context(), ch); err != nil {
 				assert.NoError(t, err)
 			}
 			close(ch)
@@ -262,7 +262,7 @@ func Test_PopulateStoragePricingMap(t *testing.T) {
 		expected            map[string]*StoragePricing
 	}{
 		"can populate storage pricing map": {
-			ctx: context.Background(),
+			ctx: t.Context(),
 			regions: []ec2Types.Region{
 				{
 					RegionName: aws.String("af-south-1"),
@@ -283,7 +283,7 @@ func Test_PopulateStoragePricingMap(t *testing.T) {
 			},
 		},
 		"errors listing storage prices propagate": {
-			ctx: context.Background(),
+			ctx: t.Context(),
 			regions: []ec2Types.Region{{
 				RegionName: aws.String("af-south-1"),
 			}},
@@ -295,7 +295,7 @@ func Test_PopulateStoragePricingMap(t *testing.T) {
 			expected:      map[string]*StoragePricing{},
 		},
 		"errors generating the map from listed prices propagate too": {
-			ctx: context.Background(),
+			ctx: t.Context(),
 			regions: []ec2Types.Region{
 				{
 					RegionName: aws.String("af-south-1"),
@@ -370,7 +370,7 @@ func Test_FetchVolumesData(t *testing.T) {
 		wg := sync.WaitGroup{}
 		wg.Add(len(collector.Regions))
 		ch := make(chan []ec2Types.Volume)
-		go collector.fetchVolumesData(context.Background(), c, regionName, ch)
+		go collector.fetchVolumesData(t.Context(), c, regionName, ch)
 		go func() {
 			wg.Wait()
 			close(ch)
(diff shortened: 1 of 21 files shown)


pkg/aws/ec2/ec2_test.go
```diff
@@ -45,7 +45,7 @@ func TestCollector_Collect(t *testing.T) {
 		})
 		ch := make(chan prometheus.Metric)
 		go func() {
-			err := collector.Collect(ch)
+			err := collector.Collect(context.Background(), ch)
```
inkel: The testing interface has a method that returns a context more suitable to use in tests:
```suggestion
			err := collector.Collect(t.Context(), ch)
```
Same for the below usages of `context.Background`.

pkg/aws/s3/s3.go
```diff
@@ -76,7 +76,7 @@ func (c *Collector) Describe(ch chan<- *prometheus.Desc) error {
 }
 
 // Collect is the function that will be called by the Prometheus client anytime a scrape is performed.
-func (c *Collector) Collect(ch chan<- prometheus.Metric) error {
+func (c *Collector) Collect(ctx context.Context, ch chan<- prometheus.Metric) error {
 	up := c.CollectMetrics(ch)
```
inkel, line 80: Note that `*Collector.CollectMetrics`, defined in line 114, uses a context, so we need to pass `ctx` down this one and use it instead of `context.Background`.

pkg/google/gcs/gcs.go
```diff
@@ -146,33 +142,43 @@ func (c *Collector) Register(registry provider.Registry) error {
 }
 
 // CollectMetrics is by `c.Collect` and can likely be refactored directly into `c.Collect`
-func (c *Collector) CollectMetrics(_ chan<- prometheus.Metric) float64 {
+// Deprecated: CollectMetrics is deprecated and will be removed in a future release.
+func (c *Collector) CollectMetrics(ch chan<- prometheus.Metric) float64 {
```
inkel, line 146: Why not passing a context to this one?

nikimanoledaki (commented):
Collectors implement the Prometheus Collector interface. We should verify whether this interface accepts a context as the first parameter 🤔 

Found this open issue: https://github.com/prometheus/client_golang/issues/1538

Upstream Go library mainly show examples for `func (c *Collector) Collect(ch chan<- prometheus.Metric)`: https://github.com/search?q=repo%3Aprometheus%2Fclient_golang%20Collect(ch%20chan%3C-%20prometheus.Metric)&type=code

I think it might not be possible, but the real test would be to deploy it to a live (dev) environment. In this case we'd need to check whether it'll be scraped by Alloy. The steps to take would be to create a dev image, push it to the dev container registry, and roll it to a dev cluster - happy to share more info on how to do that if needed.

I think the safer way to propagate context that doesnt change the interface would be via a `context.Background()` call at the start of each `Collect()` func rather than passed directly to it.

pkg/aws/ec2/ec2_test.go
rooneyshuman: updating all tests to use `t.Context()`

pkg/aws/s3/s3.go
rooneyshuman: good catch! I'm not sure it's worth refactoring `CollectMetrics()` across the codebase to take a context. 

This is a function that's part of the `Collector` interface and as such defined across all collectors that implement the interface so it wouldn't be a small lift. It seems that the intent is for this function to be entirely deprecated based on [this comment](https://github.com/grafana/cloudcost-exporter/blob/main/pkg/aws/ec2/ec2.go#L98-L99) and its empty implementation in newer collectors.

Happy to tackle that deprecation in a future PR. Let me know what you think!

pkg/google/gcs/gcs.go
rooneyshuman: as mentioned in my previous comment, this function is deprecated. You'll notice the `Collect()` method actually now calls `collectMetrics()` directly, passing the calling context. 

The `CollectMetrics()` (capital C) method which used to act as a wrapper for `collectMetrics()`, is now bypassed and only included to fulfill the interface and not called anywhere.

pkg/aws/s3/s3.go
inkel: 🤔 good point. If it's deprecated then yeah, it isn't probably worth to make the change in this PR.

inkel (approved):
🔥 :shipit:

PR #705, opened 2025-11-21
chore: Update ELB mock destination

nikimanoledaki:
* Update the ELB mock destination to get rid of outlier top-level dir: https://github.com/grafana/cloudcost-exporter/tree/main/mocks/pkg/aws/services/elbv2
* Ran `make test`, which updated the RDS mock as well.


pkg/aws/services/elbv2/elbv2.go
```diff
@@ -1,4 +1,4 @@
-//go:generate mockgen -source=elbv2.go -destination=../../../../mocks/pkg/aws/services/elbv2/ELBv2.go
+//go:generate mockgen -source=elbv2.go -destination=mocks/client.go
```
nikimanoledaki, line 1: This is the only manual change - everything else was generated by running `make test`'s mockgen

PR #711, opened 2025-11-25
Add histogram using gatherer interface to detect errors when collecting prometheus metrics

leonorfmartins:
### What this does

[Gatherers interface](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#Gatherers) is from the prometheus package and it allows us to inspect metrics being collected and check for errors.
We scan each collector and push a native histogram that looks like this:
```
cloudcost_exporter_collector_duration_seconds{
  collector=<my_collector>,
  duration=<scrape_dur>
}
```

This histogram can be used to then plot charts for each collector's health (e.g. error rate and duration). Since it's an histogram, it will be easier to make SLOs out of it as well.

Besides it, there are 2 other counter metrics:
- `cloudcost_exporter_collector_total`
- `cloudcost_exporter_collector_error_total`

which will increment on each scrape and on errors, respectively.

We can also leverage the gatherer to more precisely assert on metrics expectations.

### Test
- GCP:
```
# TYPE cloudcost_exporter_collector_duration_seconds histogram
cloudcost_exporter_collector_duration_seconds_bucket{collector="GCS",le="+Inf"} 1
cloudcost_exporter_collector_duration_seconds_sum{collector="GCS"} 0.000228667
cloudcost_exporter_collector_duration_seconds_count{collector="GCS"} 1
cloudcost_exporter_collector_duration_seconds_bucket{collector="cloudsql",le="+Inf"} 1
cloudcost_exporter_collector_duration_seconds_sum{collector="cloudsql"} 5.3625e-05
cloudcost_exporter_collector_duration_seconds_count{collector="cloudsql"} 1
cloudcost_exporter_collector_duration_seconds_bucket{collector="gcp_gke",le="+Inf"} 1
cloudcost_exporter_collector_duration_seconds_sum{collector="gcp_gke"} 2.125e-06
cloudcost_exporter_collector_duration_seconds_count{collector="gcp_gke"} 1
# TYPE cloudcost_exporter_collector_total counter
cloudcost_exporter_collector_total{collector="GCS"} 1
cloudcost_exporter_collector_total{collector="cloudsql"} 1
cloudcost_exporter_collector_total{collector="gcp_gke"} 1
```

- AWS:
```
# TYPE cloudcost_exporter_collector_duration_seconds histogram
cloudcost_exporter_collector_duration_seconds_bucket{collector="aws_elb",le="+Inf"} 1
cloudcost_exporter_collector_duration_seconds_sum{collector="aws_elb"} 1.4375e-05
cloudcost_exporter_collector_duration_seconds_count{collector="aws_elb"} 1
cloudcost_exporter_collector_duration_seconds_bucket{collector="aws_rds",le="+Inf"} 1
cloudcost_exporter_collector_duration_seconds_sum{collector="aws_rds"} 1.75e-06
cloudcost_exporter_collector_duration_seconds_count{collector="aws_rds"} 1
# TYPE cloudcost_exporter_collector_total counter
cloudcost_exporter_collector_total{collector="aws_elb"} 3
cloudcost_exporter_collector_total{collector="aws_rds"} 3
```

- Azure:
```
# HELP cloudcost_exporter_collector_duration_seconds Duration of a collector scrape in seconds
# TYPE cloudcost_exporter_collector_duration_seconds histogram
cloudcost_exporter_collector_duration_seconds_bucket{collector="azure_aks",le="+Inf"} 1
cloudcost_exporter_collector_duration_seconds_sum{collector="azure_aks"} 0.000239417
cloudcost_exporter_collector_duration_seconds_count{collector="azure_aks"} 1
# HELP cloudcost_exporter_collector_last_scrape_duration_seconds Duration of the last scrape in seconds.
# TYPE cloudcost_exporter_collector_last_scrape_duration_seconds gauge
cloudcost_exporter_collector_last_scrape_duration_seconds{collector="azure_aks",provider="azure"} 0.000239417
# TYPE cloudcost_exporter_collector_total counter
cloudcost_exporter_collector_total{collector="azure_aks"} 1
```

Relates to https://github.com/grafana/deployment_tools/issues/413121

### DIFF
--- a/pkg/google/gcp.go
+++ b/pkg/google/gcp.go
@@ -175,6 +175,13 @@ func (g *GCP) Describe(ch chan<- *prometheus.Desc) {
 	ch <- collectorLastScrapeErrorDesc
 	ch <- collectorDurationDesc
 	ch <- collectorLastScrapeTime
+	for _, c := range g.collectors {
+		if err := c.Describe(ch); err != nil {
+			g.logger.LogAttrs(context.Background(), slog.LevelError, "Error calling describe",
+				slog.String("message", err.Error()),
+			)
+		}
+	}
 }
 
 // Collect implements the prometheus.Collector interface and will iterate over all the collectors instantiated during New and collect their metrics.


leonorfmartins:
thanks for the pointer Niki! It got me wondering if we should also return the total number of metrics, [like ksm does](https://github.com/kubernetes/kube-state-metrics/blob/ef2af5bf468b75336b96850cb6d249368215ebd0/pkg/util/utils.go#L157)... on the other hand, since we are emitting now one metric per collector, I guess we could just count the number of metrics 🤔

pkg/aws/aws.go
```diff
@@ -247,18 +248,17 @@ func (a *AWS) Collect(ch chan<- prometheus.Metric) {
 	wg.Add(len(a.collectors))
 	for _, c := range a.collectors {
 		go func(c provider.Collector) {
-			now := time.Now()
 			defer wg.Done()
+
+			duration, hasError := gatherer.CollectWithGatherer(collectCtx, c, ch, a.logger)
```
nikimanoledaki, line 253: Are we collecting metrics twice here via `CollectWithGatherer`? 🤔

leonorfmartins:
yes, I can definitely add documentation for these metrics 😃  I'll just give some context here, for the sake of quickly answering your questions:
>Are we collecting metrics twice here?

Yes, temporarily, specially all the metrics you are mentioning below such as `cloudcost_exporter_collector_success` or `cloudcost_collector_last_scrape_error` and `cloudcost_collector_last_scrape_time`. They are still being used by some of our dashboards, but we have seen how they are not very reliable. Yet, until we have an improved dashboard to monitor all our collectors, I wouldn't want to get rid of them, as having some collectors being monitored is better than having none. As you already pointed out, existing metrics don't even cover all provider collectors.
I can communicate this better with a comment on the code highlighting this temporary situation as well 👍 

So, the plan is in the future to have only the metrics we are adding in this PR. Metrics such as
- `cloudcost_exporter_collector_success`
- `cloudcost_collector_last_scrape_error`
- `cloudcost_collector_last_scrape_time`
- `cloudcost_exporter_<csp>`

are to be deprecated as they don't accurately provide a way for us to monitor our collectors.

pkg/gatherer/gatherer.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+// CollectWithGatherer collects metrics from a collector and uses the Gatherer interface to detect errors.
+func CollectWithGatherer(ctx context.Context, c provider.Collector, ch chan<- prometheus.Metric, logger *slog.Logger) (float64, bool) {
+	start := time.Now()
+	var hasError bool
+	var duration float64
+
+	tempRegistry := prometheus.NewRegistry()
+	// also register errors if the remporary registry to detect errors via Gatherer interface fails
```
nikimanoledaki:
```suggestion
	// also register errors if the temporary registry to detect errors via Gatherer interface fails
```

pkg/gatherer/gatherer.go
```diff
@@ -0,0 +1,112 @@
...
+	if err := c.Register(tempRegistry); err != nil {
+		hasError = true
+		logger.LogAttrs(ctx, slog.LevelError, "could not register collector with gatherer",
+			slog.String("collector", c.Name()),
+			slog.String("message", err.Error()),
+		)
+	}
+
+	duration = time.Since(start).Seconds()
+	collectErr := c.Collect(ctx, ch)
```
nikimanoledaki: We need to calculate the duration once the `Collect` method has finished:

```suggestion
	collectErr := c.Collect(ctx, ch)
	duration = time.Since(start).Seconds()
```

pkg/gatherer/gatherer.go
```diff
@@ -0,0 +1,112 @@
...
+			slog.String("message", collectErr.Error()),
+		)
+	}
+
+	if _, err := tempRegistry.Gather(); err != nil {
+		hasError = true
+		logger.LogAttrs(ctx, slog.LevelError, "did not detect gatherer",
+			slog.String("collector", c.Name()),
+			slog.String("message", err.Error()),
+		)
```
nikimanoledaki, line 94: This call to `tempRegistry.Gather()` won't work because we are calling `c.Collect()` directly earlier, which means that the metrics are sent to a `ch` channel. 👀 Essentially it repeats the metric collection, which is either expensive or yields no metrics.

It can't be used as a backup for `c.Collect`, since calling `tempRegistry.Gather()` would call `c.Collect()` again and fail with the same reason.

We can remove this safely.

```suggestion

```

pkg/google/gcp.go
```diff
@@ -174,13 +175,6 @@ func (g *GCP) Describe(ch chan<- *prometheus.Desc) {
 	ch <- collectorLastScrapeErrorDesc
 	ch <- collectorDurationDesc
 	ch <- collectorLastScrapeTime
-	for _, c := range g.collectors {
-		if err := c.Describe(ch); err != nil {
-			g.logger.LogAttrs(context.Background(), slog.LevelError, "Error calling describe",
-				slog.String("message", err.Error()),
-			)
-		}
-	}
```
nikimanoledaki, line 183: We `c.Describe` upfront for AWS & Azure as well, but it's been removed for GCP. Was this intentional cleanup? Should we revert this change?

pkg/gatherer/gatherer.go
leonorfmartins: Ok, this took me a while to understand as well, but let me try to explain why removing `gather` really defeats the purpose of this PR and why it doesn't collect metrics twice:


>This call to tempRegistry.Gather() won't work because we are calling c.Collect() directly earlier, which means that the metrics are sent to a ch channel.

`gather` doesn't collect metrics, it just reads the state of the metric registered in the temporary registry.

We need to call `c.Collect` because that call is the only one actually collecting metrics, e.g. doing the API call. `Gather` is called on a tempRegistry which is the registry Gather will use to check the metrics status and see if there are any errors. The purpose of the gather is not to be used as a backup but to "ensure that the
	// returned slice is valid and self-consistent so that it can be used
	// for valid exposition. As an exception to the strict consistency
	// requirements described for metric.Desc, Gather will tolerate
	// different sets of label names for metrics of the same metric family." ([source](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#Gatherer))

Basically, it doesn't collect again, it just checks the temp registry where metrics are and checks if they are ok or not.

How?

I'm not a prometheus expert so I'm not sure I can answer all the questions but from what I can understand, looking at the gather implementation fn, here's the part when the metric is actually checked: https://github.com/prometheus/client_golang/blob/2cd067eb23c940d3a1335ebc75eaf94e3037d8a9/prometheus/registry.go#L456

It calls collector.Collect. Collector is a [prometheus type](https://github.com/prometheus/client_golang/blob/main/prometheus/collector.go) which implements gather and describe. For gather, all it does is [read itself here](https://github.com/prometheus/client_golang/blob/2cd067eb23c940d3a1335ebc75eaf94e3037d8a9/prometheus/collector.go#L118-L120). So, it doesn't actually collect anything.

To sum up, why do I think using gather is interesting:
it's something which will just inspect how metrics are being collected, without actually collecting those. It seems suitable to what we want as we need to monitor if there were any errors.

I'm not sure if I was able to answer your questions, let me know!

pkg/aws/aws.go
leonorfmartins: temporarily yes, until we deprecate the old metrics, see the comment right below: https://github.com/grafana/cloudcost-exporter/pull/711/changes#diff-557e359d6147fd859c120871f3c4b242f8781a32fb05b6a343978ff33c3b8e2fR255

PR #715, opened 2025-11-25
perf(ec2): reduce scrape latency with concurrent worker pool

simaarfania:
Part of https://github.com/grafana/deployment_tools/issues/394392. 

In the cloudcost-exporter, EC2 collector scrape latency reaches 10+ seconds in production, causing PromScrapeFlapping alerts. This PR adds a concurrent worker pool pattern with 5 workers for parallel processing of instances and volumes. Testing showed that higher worker counts (20) actually decreased performance due to lock contention and network issues, so keeping at 5 workers. While this speeds up the processing phase, testing showed that AWS API pagination remains the primary bottleneck - large regions (eu-central-1 and us-east-2) take 9-12 seconds just to list instances due to AWS API response times.

### DIFF
--- a/pkg/aws/ec2/pricing_map.go
+++ b/pkg/aws/ec2/pricing_map.go
@@ -234,7 +234,7 @@ func (cpm *ComputePricingMap) processOnDemandBatch(batch []string) error {
 	for _, product := range batch {
 		var productInfo computeProduct
 		if err := json.Unmarshal([]byte(product), &productInfo); err != nil {
-			return err
+			return fmt.Errorf("%w: %w", ErrGeneratePricingMap, err)
 		}
 		if productInfo.Product.Attributes.InstanceType == "" {
 			// If there are no instance types, let's just continue on. This is the most important key


pkg/aws/ec2/ec2.go
```diff
@@ -223,87 +224,155 @@ func (c *Collector) fetchVolumesData(ctx context.Context, regionClient client.Cl
...
+		pricetier,
+		string(instance.Architecture),
+	}
+
+	// Use mutex to safely send to channel from multiple goroutines
+	mu.Lock()
+	ch <- prometheus.MustNewConstMetric(InstanceCPUHourlyCostDesc, prometheus.GaugeValue, price.Cpu, labelValues...)
+	ch <- prometheus.MustNewConstMetric(InstanceMemoryHourlyCostDesc, prometheus.GaugeValue, price.Ram, labelValues...)
+	ch <- prometheus.MustNewConstMetric(InstanceTotalHourlyCostDesc, prometheus.GaugeValue, price.Total, labelValues...)
+	mu.Unlock()
```
nikimanoledaki: We shouldn't need to use locks around channels since channels are concurrent goroutines 🤔

simaarfania: You're right! Fixed this :)

pkg/aws/ec2/ec2.go
```diff
@@ -223,87 +224,147 @@ func (c *Collector) fetchVolumesData(ctx context.Context, regionClient client.Cl
...
+	if err != nil {
+		c.logger.Error(fmt.Sprintf("error getting price for instance type %s: %s", instance.InstanceType, err))
+		return
 	}
+
+	labelValues := []string{
+		*instance.PrivateDnsName,
+		*instance.InstanceId,
+		region,
+		c.computePricingMap.InstanceDetails[string(instance.InstanceType)].InstanceFamily,
```
rooneyshuman: we need an RLock if we're going to read from the computePricingMap. We can create a getter function similar to to [GetPriceForInstanceType()](https://github.com/grafana/cloudcost-exporter/blob/44aa6032e477d744387d52f545bcda277c2688e3/pkg/aws/ec2/pricing_map.go#L483) to get the instance family safely

pkg/aws/ec2/pricing_map.go
```diff
@@ -121,10 +162,79 @@ func (cpm *ComputePricingMap) GenerateComputePricingMap(ctx context.Context) err
...
+	return nil
+}
+
+// processOnDemandBatch processes a batch of ondemand prices
+func (cpm *ComputePricingMap) processOnDemandBatch(batch []string) error {
+	for _, product := range batch {
 		var productInfo computeProduct
 		if err := json.Unmarshal([]byte(product), &productInfo); err != nil {
-			return fmt.Errorf("%w: %w", ErrGeneratePricingMap, err)
+			return err
```
rooneyshuman: we wrap other `json.Unmarshal` errors in this module [(example)](https://github.com/grafana/cloudcost-exporter/blob/44aa6032e477d744387d52f545bcda277c2688e3/pkg/aws/ec2/pricing_map.go#L401), let's do the same here

rooneyshuman (approved):

PR #718, opened 2025-11-28
Log less often when RDS fails to list instances

leonorfmartins:
It's enough to log if any errors occurred at the end of the loop when instances are being fetched, instead of on each entry.

### DIFF
--- a/pkg/aws/rds/rds.go
+++ b/pkg/aws/rds/rds.go
@@ -45,6 +45,12 @@ type Config struct {
 	ScrapeInterval time.Duration
 }
 
+type listError struct {
+	region string
+	error  error
+	reason string
+}
+
 const (
 	serviceName = "rds"
 )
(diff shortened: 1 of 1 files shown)


pkg/aws/rds/rds.go
```diff
@@ -70,23 +76,38 @@ func (c *Collector) CollectMetrics(_ chan<- prometheus.Metric) float64 {
...
 	var instances = []rdsTypes.DBInstance{}
+	var regionErrors []listError
 	for _, region := range c.regions {
 		regionName := *region.RegionName
 		regionClient, ok := c.regionMap[regionName]
 		if !ok {
-			logger.Error("no client found for region", "region", regionName)
+			regionErrors = append(regionErrors, listError{
+				region: regionName,
+				error:  fmt.Errorf("no client found for region"),
```
nikimanoledaki, line 86: I'm thinking through what is really means if no client was found for a region. It means that no client was created for that region [here](https://github.com/grafana/cloudcost-exporter/blob/acc0cb8d58d4a06cf2d5072157a7fbf16a8304a2/pkg/aws/aws.go#L113). 

I'm wondering where the region comes from if there was no client created for it. 

Or, is this really an error? Should it just be a `warning` or `info`-level log? 🤔 Not entirely sure, wanted to get your take on this!

leonorfmartins: IMO it should be an error: we get the regions from this [fn](https://github.com/grafana/cloudcost-exporter/blob/acc0cb8d58d4a06cf2d5072157a7fbf16a8304a2/pkg/aws/aws.go#L107). The `false` means that the aws client should only fetch regions for which we have an account enabled.
We can make the argument of "right, but we are already sending an error when we [fail to fetch the regions](https://github.com/grafana/cloudcost-exporter/blob/acc0cb8d58d4a06cf2d5072157a7fbf16a8304a2/pkg/aws/aws.go#L107), do we need to send an error again when down the line, it fails to fetch RDS instances because it failed to fetch regions and therefore, to have a region client?"
I think we could still keep an error since it would give us visibility on the consequence of failing to fetch a region... but I'm happy to discuss other options :)

PR #779, opened 2026-02-10
Update grafana-foundation-sdk to v0.0.6 for dashboard builds

nikimanoledaki:
## Problem

The CI check "Check for Dashboards Drift" is failing because the Makefile's `build-dashboards` target uses `@latest` which now resolves to v0.0.6, but go.mod still has the older commit hash from Feb 6.

## Solution

Update go.mod and go.sum to v0.0.6 to match what `make build-dashboards` generates.


nikimanoledaki:
Duplicate of: https://github.com/grafana/cloudcost-exporter/pull/780

Issue #743, opened 2026-01-07
Migrate Helm chart artifacts to `grafana/cloudcost-exporter` for CD ownership & consistency


nikimanoledaki:
## What?

Migrate the Helm chart artefacts into the `grafana/cloudcost-exporter` repo by creating a workflow in this repo.

### 1. Ownership

When there is a new release of the cloudcost-exporter application, we need 2 reviews for PRs to the [`grafana/helm-charts` repository](https://github.com/grafana/helm-charts/tree/main/charts/cloudcost-exporter), which slows down our release process. We have fallen behind on the application version used in the Helm chart. We could fully automate the release and deployment process if the Helm chart artefacts were closer to the application code.

### 2. Consistency

Having the Helm chart artefacts close to the source code would maintain consistency with the OSS strategy of other `grafana/<projects>` repositories:
  * [Alloy](https://github.com/grafana/alloy/tree/main/operations/helm/charts/alloy)  
  * [Alloy Operator](https://github.com/grafana/alloy-operator/tree/main/charts/alloy-operator) \- [Release workflow](https://github.com/grafana/alloy-operator/blob/main/.github/workflows/release-alloy-operator.yaml) (I made this most recently, so it utilizes more of our shared workflows)  
  * [Loki](https://github.com/grafana/loki/tree/main/production/helm/loki)

## How?

The workflow should abstract away as many steps as possible by using shared workflows. For example, @petewall shared the [Alloy Operator's Helm chart](https://github.com/grafana/alloy-operator/tree/main/charts/alloy-operator) uses this [release workflow](https://github.com/grafana/alloy-operator/blob/main/.github/workflows/release-alloy-operator.yaml) that references many Grafana Labs shared workflows.

PR #797, opened 2026-02-19
Upgrade cloudcost-exporter-gh-runners helm chart to the latest version

leonorfmartins:
Part of https://github.com/grafana/cloudcost-exporter/issues/743

### DIFF
--- a/charts/cloudcost-exporter/Chart.yaml
+++ b/charts/cloudcost-exporter/Chart.yaml
@@ -5,10 +5,10 @@ type: application
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 1.0.9
+version: 1.0.10
 
 # This is the version of cloudcost-exporter to be deployed, which should be incremented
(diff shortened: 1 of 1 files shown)


charts/cloudcost-exporter/Chart.yaml
```diff
@@ -5,10 +5,10 @@ type: application
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 1.0.9
+version: 1.0.10
 
 # This is the version of cloudcost-exporter to be deployed, which should be incremented
 # with each release.
-appVersion: "0.16.0"
+appVersion: "0.22.0"
```
suntala, line 12: The comment above mentions incrementing with each release. I assume it's fine to jump several minor versions.

leonorfmartins: I was referring to the helm version which is being incremented. We are jumping over several versions from the cloudcost-exporter so it finally matches to the latest release :)

PR #802, opened 2026-02-20
fix: stop pricing refresh tickers on goroutine exit

stephan-rayner:
See: #803 

## What

  Pricing refresh goroutines in five collectors were creating `time.Ticker` instances outside the
  goroutine body, with no call to `ticker.Stop()`. This leaks the ticker's internal goroutine and
  its associated resources for the lifetime of the process, even after context cancellation.

  Affected collectors:
  - `pkg/aws/ec2` — `computeTicker`, `storageTicker`
  - `pkg/aws/natgateway` — `priceTicker`
  - `pkg/azure/aks` — `priceTicker`, `machineTicker`, `diskTicker`
  - `pkg/google/gke` — `priceTicker`
  - `pkg/google/networking` — `priceTicker`

## Why

**Issue:** Goroutines started without proper cleanup mechanisms.

**Pricing Map Refresh Goroutines:**
All rely solely on context cancellation for cleanup:

- `pkg/aws/ec2/ec2.go:99-122` - Two goroutines started
- `pkg/google/gke/gke.go:245-257` - Background refresh
- `pkg/azure/aks/aks.go:127-157` - Three goroutines

**The Problem:** If context is not properly cancelled, goroutines leak indefinitely.

## How

Moved each `time.NewTicker(...)` call to the first line inside its goroutine and added an
immediate `defer ticker.Stop()`. This guarantees the ticker is stopped when the goroutine exits
(via context cancellation or otherwise), releasing its internal resources.

`pkg/google/vpc` and `pkg/aws/vpc` already followed this pattern correctly and were not changed.


nikimanoledaki (approved):
LGTM, thanks @stephan-rayner! 🎉

Issue #818, opened 2026-02-23
fix: replace context.Background() with passed context throughout codebase


stephan-rayner:
## Summary

Inappropriate use of `context.Background()` in multiple places where a real context is already available and should be passed down. This breaks the context cancellation chain and makes graceful shutdown unreliable.

## Problem

When a function receives a `ctx` parameter but internally calls `context.Background()`, it severs the cancellation chain. Any timeout, deadline, or cancellation signal from the caller is silently ignored.

### `context.Background()` violations (context is available but ignored)

| File | Issue |
|------|-------|
| `pkg/aws/s3/s3.go:123` | Uses `context.Background()` in API call instead of passing `ctx` |
| `pkg/aws/aws.go:216, 233, 255` | Logging with `context.Background()` instead of passed context |
| `pkg/google/gcs/gcs.go:146` | `c.collectMetrics(context.Background(), ch)` |
| `pkg/google/gke/gke.go:69` | `c.Collect(context.Background(), ch)` |
| `pkg/google/vpc/vpc.go:148` | `c.collectMetrics(context.Background(), ch)` |

### `context.TODO()` usage (should be replaced with a properly derived context)

| File | Location |
|------|----------|
| `scripts/aws-spot-pricing/main.go` | Lines 19, 38 |
| `pkg/azure/client/azure_client.go` | Line 35 |

## Impact

- **Breaks context cancellation chains** — a cancelled context from the caller has no effect
- **Prevents timeout propagation** — deadlines set by callers are ignored
- **Makes graceful shutdown unreliable** — components may not stop when the root context is cancelled

## Proposed Fix

Replace all `context.Background()` calls with the `ctx` parameter that is already in scope:

```go
// Before
c.collectMetrics(context.Background(), ch)

// After
c.collectMetrics(ctx, ch)
```

For `context.TODO()`, derive the context appropriately from the caller's context or the application root context.

## Acceptance Criteria

- [ ] All `context.Background()` calls inside non-`main` functions replaced with the passed `ctx`
- [ ] All `context.TODO()` calls replaced with properly derived contexts
- [ ] Consider adding `contextcheck` linter rule to prevent regressions

PR #820, opened 2026-02-23
  fix: remove deprecated CollectMetrics from Collector interface

stephan-rayner:
Removes CollectMetrics(chan<- prometheus.Metric) float64 from the
  Collector interface. Because it lacked a context parameter, every
  implementation was forced to call context.Background() when delegating
  to real work.

  - pkg/provider: remove method from interface, regenerate mock
  - pkg/aws/s3: inline CollectMetrics body into Collect, replacing context.Background() with the propagated ctx
  - All other collectors: delete the no-op / delegating stub (ec2, rds, vpc, natgateway, elb, gcs, gke, google/vpc, cloudsql, forwarding_rule, azure/aks)
  - Tests: update s3, gcs, and gke tests to call Collect(ctx, ch) and assert on error instead of float64
  
This is part of a broader effort defined in #818

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -22,7 +22,7 @@ require (
 	github.com/aws/aws-sdk-go-v2/service/sts v1.41.6
 	github.com/google/go-cmp v0.7.0
 	github.com/googleapis/gax-go/v2 v2.17.0
-	github.com/grafana/grafana-foundation-sdk/go v0.0.9
+	github.com/grafana/grafana-foundation-sdk/go v0.0.10
 	github.com/prometheus/client_golang v1.23.2
 	github.com/prometheus/client_model v0.6.2
 	github.com/prometheus/common v0.67.5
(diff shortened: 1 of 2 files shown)


pkg/aws/s3/s3.go
```diff
@@ -77,10 +76,22 @@ func (c *Collector) Describe(ch chan<- *prometheus.Desc) error {
 
 // Collect is the function that will be called by the Prometheus client anytime a scrape is performed.
 func (c *Collector) Collect(ctx context.Context, ch chan<- prometheus.Metric) error {
```
rooneyshuman:
```suggestion
func (c *Collector) Collect(ctx context.Context, _ chan<- prometheus.Metric) error {
```
while most collectors write metrics to the channel passed in, there are a few exceptions that instead register gauges and update values when `Gather()` runs. For these exceptions, let's use the Go convention of the blank identifier `_` to mark the parameter as intentionally unused

pkg/google/gcs/gcs.go
```diff
@@ -140,15 +140,6 @@ func (c *Collector) Register(registry provider.Registry) error {
...
-// Deprecated: CollectMetrics is deprecated and will be removed in a future release.
-func (c *Collector) CollectMetrics(ch chan<- prometheus.Metric) float64 {
-	if err := c.collectMetrics(context.Background(), ch); err != nil {
-		return 0
-	}
-	return 1
-}
-
 // collectMetrics performs the actual collection work
 func (c *Collector) collectMetrics(ctx context.Context, ch chan<- prometheus.Metric) error {
```
rooneyshuman:
```suggestion
func (c *Collector) collectMetrics(ctx context.Context) error {
```
the channel isn't used in this function

pkg/google/gcs/gcs.go
rooneyshuman:
```suggestion
func (c *Collector) Collect(ctx context.Context, _ chan<- prometheus.Metric) error {
return c.collectMetrics(ctx)
```
along with the channel changes, do we need to have `Collect()` be a wrapper for `collectMetrics()` at all here? I think we can just insert that logic into the outer function

Issue #742, opened 2026-01-07
Remove GCS discounts-specific logic


nikimanoledaki:
Cloudcost-exporter currently factors in discounts for GCS in its pricing logic:
https://github.com/grafana/cloudcost-exporter/blob/093b2b87dee37dfb794e360e5dd1ca920914aee6/pkg/google/gcs/gcs.go#L24

These discount ratios need to be phased out. Discount ratios should be injected outside of the pricing logic.

stephan-rayner:
These metrics may still be used so removing them may require additional steps first. I am opening this as a discussion to the team.

jjo:
Next (safe) two steps:

### 1) deployment_tools

**Completely remove CCE GCS provided discount metrics**, can (and should) be done in a **single PR per wave** (ie `dev`, then `ops` follow-up) by below **two simultaneous changes** comprising **four edits**:

a) **Remove below three stanzas that use CCE's `cloudcost_gcp_gcs_storage_discount_by_location_usd_per_gibyte_hour`**, as these are joins to get multiplied-by `1 - discount`, just removing them will let the main value thru (1st dev, then ops):
  * [dev, ops] https://github.com/grafana/deployment_tools/blob/6697c3c4077b9816f4ae6dc7cc36f20910236df9/ksonnet/lib/cost/rules/no_prepare.libsonnet#L50-L55
  * [dev, ops] https://github.com/grafana/deployment_tools/blob/6697c3c4077b9816f4ae6dc7cc36f20910236df9/ksonnet/lib/cost/rules/no_prepare.libsonnet#L180-L184
  *  [dev, ops] https://github.com/grafana/deployment_tools/blob/6697c3c4077b9816f4ae6dc7cc36f20910236df9/ksonnet/lib/cost/rules/no_prepare.libsonnet#L215-L219

 The key ~challenge there is make it single-wave aware (progressive dev -> ops).

b) **Remove GCS exception** from `discount_data_dev.yaml` (then `discount_data.yaml` for ops):
* [dev] https://github.com/grafana/deployment_tools/blob/6697c3c4077b9816f4ae6dc7cc36f20910236df9/ksonnet/lib/cost-querier/discounts_data_dev.yaml#L40-L44

* [ops] https://github.com/grafana/deployment_tools/blob/6697c3c4077b9816f4ae6dc7cc36f20910236df9/ksonnet/lib/cost-querier/discounts_data.yaml#L40-L44

[...]

PR #823, opened 2026-02-24
fix: removing GCS discounts-specific logic

stephan-rayner:
Cloudcost-exporter currently factors in discounts for GCS in its pricing logic.

These discount ratios need to be phased out. Discount ratios should be injected outside of the pricing logic.

See: https://github.com/grafana/cloudcost-exporter/issues/742 for details


pkg/google/gcs/gcs_test.go
```diff
@@ -280,9 +269,7 @@ func TestCollector_Collect(t *testing.T) {
 
 	metricNames := []string{
 		"cloudcost_gcp_gcs_storage_by_location_usd_per_gibyte_hour",
-		"cloudcost_gcp_gcs_storage_discount_by_location_usd_per_gibyte_hour",
 		"cloudcost_gcp_gcs_operation_by_location_usd_per_krequest",
-		"cloudcost_gcp_gcs_operation_discount_by_location_usd_per_krequest",
```
nikimanoledaki, line 285: Check whether these metrics are used 👀 
- `cloudcost_gcp_gcs_storage_discount_by_location_usd_per_gibyte_hour`
- `cloudcost_gcp_gcs_operation_discount_by_location_usd_per_krequest`

stephan-rayner:
This PR has gotten so far out of date I don't think it is worth saving

PR #827, opened 2026-02-25
perf(ec2): filter terminated instances at API level to reduce scrape latency

simaarfania:
`cloudcost-exporter` is hitting the 10s Prometheus scrape timeout for the EC2 collector in `prod-us-east-2`, causing alerts in the warnings channel. This seems to be because `DescribeInstances` is fetching all instances (including terminated), and only silently discarding terminated ones inside `processInstance` via the empty `PrivateDnsName` check.

This PR adds an `instance-state-name` filter  to exclude terminated/shutting-down instances at the API level, as they are non-billable. This moves the exclusion earlier so we never fetch those instances in the first place. This results in no data loss as they were already excluded from all metrics by the check in [processInstance](https://github.com/grafana/cloudcost-exporter/blob/main/pkg/aws/ec2/ec2.go#L258-L260).

Only `running` and `stopping` instances incur compute costs per the [EC2 instance billing docs](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-lifecycle.html#instance-billing-by-state). `stopping` is included specifically to cover hibernated instances. All other states (pending, stopped, shutting-down, terminated) incur no instance-hour charges.


cmd/test-filter/main.go
rooneyshuman, line 1: the `cmd/` dir doesn't seem like the right place for these tests?

PR #833, opened 2026-02-27
Use slog across the whole project. Fix error handling on string output

leonorfmartins:
What this PR does:
- standardize logging with slog — log and slog packages were being used depending on the package. We should use the same standardised library for the whole project.
- fix error handling on string output — we should validate that strings aren't empty before using them.

Fixes https://github.com/grafana/deployment_tools/issues/497062

### DIFF
--- a/pkg/google/client/billing.go
+++ b/pkg/google/client/billing.go
@@ -74,7 +74,7 @@ func (b *Billing) exportBilling(ctx context.Context, serviceName string, m *metr
 		if sku.Description == "Bucket Tagging Storage" {
 			continue
 		}
-		if strings.HasSuffix(sku.Category.ResourceGroup, "Storage") {
+		if sku.Category != nil && strings.HasSuffix(sku.Category.ResourceGroup, "Storage") {
 			if strings.Contains(sku.Description, "Early Delete") {
 				continue // to skip "Unknown sku"
 			}


pkg/google/client/billing.go
nikimanoledaki:
```suggestion
		if sku.Category != nil && strings.HasSuffix(sku.Category.ResourceGroup, "Storage") {
```

pkg/google/client/billing.go
```diff
@@ -79,17 +79,17 @@ func (b *Billing) exportBilling(ctx context.Context, serviceName string, m *metr
 				continue // to skip "Unknown sku"
 			}
 			if err := parseStorageSku(sku, m); err != nil {
-				log.Printf("error parsing storage sku: %v", err)
+				slog.Error("error parsing storage sku", "error", err)
 			}
 			continue
 		}
 		if strings.HasSuffix(sku.Category.ResourceGroup, "Ops") {
```
nikimanoledaki, line 86:
```suggestion
		if sku.Category != nil && strings.HasSuffix(sku.Category.ResourceGroup, "Ops") {
```

leonorfmartins: applied [here](https://github.com/grafana/cloudcost-exporter/pull/833/commits/e949bdccc2056a7646d348ea2c35f44d1453a3dd)

pkg/google/client/billing.go
```diff
@@ -74,22 +74,22 @@ func (b *Billing) exportBilling(ctx context.Context, serviceName string, m *metr
...
 			if err := parseOpSku(sku, m); err != nil {
-				log.Printf("error parsing op sku: %v", err)
+				slog.Error("error parsing op sku", "error", err)
 			}
 			continue
 		}
-		log.Printf("Unknown sku: %s\n", sku.Description)
+		slog.Warn("Unknown sku", "description", sku.Description)
```
nikimanoledaki, line 92:
```suggestion
		slog.Debug("Unknown sku", "description", sku.Description)
```

PR #835, opened 2026-03-02
add region label to operational metrics

leonorfmartins:
Adding the region label to:
- cloudcost_exporter_collector_duration_seconds histogram
- cloudcost_exporter_collector_total
- cloudcost_exporter_collector_errors

## How
For Azure and AWS:
- regions are gotten via a Regions() interface implementation. Regions values either come from the regions client (AWS) or from types.Region when listing the resource
For GCP:
-  used by collectors that receive regions at construction time and store them directly as strings.

## Test
**GCP**
```
cloudcost_exporter_collector_duration_seconds_bucket{collector="ForwardingRule",region="africa-south1",le="+Inf"} 1
cloudcost_exporter_collector_duration_seconds_sum{collector="ForwardingRule",region="africa-south1"} 9.329538958
cloudcost_exporter_collector_duration_seconds_count{collector="ForwardingRule",region="africa-south1"} 1
```

**AWS**
```
cloudcost_exporter_collector_duration_seconds_bucket{collector="NATGATEWAY",region="af-south-1",le="+Inf"} 1
cloudcost_exporter_collector_duration_seconds_sum{collector="NATGATEWAY",region="af-south-1"} 0.000425583
cloudcost_exporter_collector_duration_seconds_count{collector="NATGATEWAY",region="af-south-1"} 1
```

**Azure**
```
cloudcost_exporter_collector_duration_seconds_bucket{collector="azure_aks",region="westeurope",le="+Inf"} 1
cloudcost_exporter_collector_duration_seconds_sum{collector="azure_aks",region="westeurope"} 0.001469708
```

Part of https://github.com/grafana/deployment_tools/issues/413121


cmd/exporter/exporter.go
```diff
@@ -76,6 +76,7 @@ func providerFlags(fs *flag.FlagSet, cfg *config.Config) {
 	fs.Var(&cfg.Providers.Azure.Services, "azure.services", "Azure service(s).")
 	fs.Var(&cfg.Providers.GCP.Services, "gcp.services", "GCP service(s).")
 	flag.StringVar(&cfg.Providers.AWS.Region, "aws.region", "", "AWS region")
+	flag.StringVar(&cfg.Providers.Azure.Region, "azure.region", "", "Azure region")
```
nikimanoledaki: I'd remove this Azure region flag altogether because the `<csp>.region` flag is the region that the CSP uses to authenticate. This is something we use for AWS only. This could definitely be better documented here (in a follow-up PR) to avoid confusion 👍 

From: https://github.com/grafana/cloudcost-exporter/tree/main/docs/deploying/aws#containerargs-required

> Set the region that the client should authenticate with. Check which region is used by the [AWS IAM endpoint](https://docs.aws.amazon.com/general/latest/gr/iam-service.html). This is often a global endpoint so only needs to be set to one region.
>
>  - "--aws.region=us-east-1"

pkg/aws/aws.go
```diff
@@ -250,7 +250,7 @@ func (a *AWS) Collect(ch chan<- prometheus.Metric) {
 		go func(c provider.Collector) {
 			defer wg.Done()
 
-			duration, hasError := gatherer.CollectWithGatherer(collectCtx, c, ch, a.logger)
+			duration, hasError := gatherer.CollectWithGatherer(collectCtx, c, ch, a.logger, a.Config.Region)
```
nikimanoledaki: This is using the config region, which is the region CCE authenticated with / is running on - see `createAWSConfig` func. It's not the same as the region the clients will try collecting data from. 

The region to collect data from would come from the list of `regions` returned by the client [here](https://github.com/grafana/cloudcost-exporter/blob/d2c2d076de069420ead289d1db52400deb7f8d49/pkg/aws/aws.go#L108) and used to create the region map in the `newRegionClientMap` func [here](https://github.com/grafana/cloudcost-exporter/blob/d2c2d076de069420ead289d1db52400deb7f8d49/pkg/aws/aws.go#L268). 

We could potentially do a liveness check for each region when we create a the new region client map if that's possible, and/or only bubble up the region to the operational metric. WDYT?

leonorfmartins: I'm going more for the approach (AWS case obly, GCP and Azure need to be done differently) of implementing a regions interface which will make sure to collect the region as we list each resource. I think this would be a better approach than creating a new region client: since regions are only ever used for operational metrics, it would add a bit more latency by creating this new client.

pkg/gatherer/gatherer.go
```diff
@@ -72,6 +72,13 @@ func CollectWithGatherer(ctx context.Context, c provider.Collector, ch chan<- pr
 		)
 	}
 
+	regions := []string{"unknown"}
```
nikimanoledaki: WDYT about surfacing `unknown` as a const in a shared library? That way we can make it the default fallback region across providers 👍 It's used primarily here and in AWS's billing service so far 👀

pkg/aws/ec2/ec2.go
```diff
@@ -374,6 +376,10 @@ func (c *Collector) Name() string {
 	return subsystem
 }
 
+func (c *Collector) Regions() []string {
+	return slices.Collect(maps.Keys(c.awsRegionClientMap))
+}
+
```
nikimanoledaki, line 380: This is repeated in other services. Can it be refactored as a helper func (might just need to take the region map as an arg)?

pkg/aws/vpc/vpc.go
```diff
@@ -116,6 +116,14 @@ type Config struct {
 
 func (c *Collector) Name() string { return strings.ToUpper(serviceName) }
 
+func (c *Collector) Regions() []string {
+	regions := make([]string, 0, len(c.regions))
+	for _, r := range c.regions {
+		regions = append(regions, *r.RegionName)
+	}
+	return regions
+}
```
nikimanoledaki, line 121: This code is also repeated and could be extracted into a helper func 👀

pkg/aws/ec2/ec2.go
leonorfmartins: addressed in https://github.com/grafana/cloudcost-exporter/pull/835/commits/1bc94fd0948a71d7ade594bbf156f80ccc53096c

pkg/gatherer/gatherer.go
leonorfmartins: addressed in https://github.com/grafana/cloudcost-exporter/pull/835/commits/1bc94fd0948a71d7ade594bbf156f80ccc53096c

nikimanoledaki (approved):
LGTM, thank you for taking the time to implement my refactoring suggestions. It's so clean! *chef's kiss* 💯

PR #840, opened 2026-03-06
refactor: improve testability of provider selection and add unit tests in `cmd/exporter`

stephan-rayner:
This PR improves the testability of the exporter's provider selection logic and adds unit tests for previously uncovered functions in \`cmd/exporter\`.

**Key changes:**

- Refactored \`selectProvider\` in \`cmd/exporter/exporter.go\` to use a new \`selectProviderWith\` function. This introduces a dependency injection pattern by allowing provider constructors (AWS, Azure, GCP) to be passed as arguments, enabling them to be stubbed during unit testing.
- Added \`Test_selectProvider\` in \`cmd/exporter/exporter_test.go\` using a table-driven approach to verify:
    - Correct selection of AWS, Azure, and GCP providers based on configuration.
    - Error propagation when any provider constructor (AWS, Azure, or GCP) fails.
    - Error handling for unknown provider types.
    - The zero-value \`collectorTimeout\` default: when \`cfg.Collector.Timeout\` is unset, a 1-minute default is forwarded to the constructor.
    - Explicit timeout values are passed through to the constructor unchanged.
- Added \`Test_createPromRegistryHandler\` to verify the Prometheus registry setup, ensuring that collectors are registered correctly and that the resulting HTTP handler functions as expected.
- Integrated \`gomock\` to mock the \`Provider\` interface, allowing for robust verification of interactions during the exporter's initialization phase.

### DIFF
--- a/cmd/exporter/exporter_test.go
+++ b/cmd/exporter/exporter_test.go
@@ -6,6 +6,7 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"testing"
+	"time"
 
 	"github.com/grafana/cloudcost-exporter/cmd/exporter/config"
 	"github.com/grafana/cloudcost-exporter/pkg/aws"
@@ -18,12 +19,14 @@ import (
 
 func Test_selectProvider(t *testing.T) {
 	tests := map[string]struct {
-		providerName   string
-		awsCalled      bool
-		azureCalled    bool
-		gcpCalled      bool
-		constructorErr error
-		wantErr        bool
+		providerName         string
+		collectorTimeout     time.Duration
+		awsCalled            bool
+		azureCalled          bool
+		gcpCalled            bool
+		constructorErr       error
+		wantErr              bool
+		wantCollectorTimeout time.Duration // if non-zero, asserts the timeout forwarded to the constructor
 	}{
 		"aws provider": {
 			providerName: "aws",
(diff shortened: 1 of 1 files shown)


cmd/exporter/exporter.go
```diff
@@ -208,6 +208,20 @@ func createPromRegistryHandler(csp provider.Provider, region string) (http.Handl
...
+	)
+}
+
+func selectProviderWith(
+	ctx context.Context,
+	cfg *config.Config,
+	newAWS func(context.Context, *aws.Config) (provider.Provider, error),
+	newAzure func(context.Context, *azure.Config) (provider.Provider, error),
+	newGCP func(context.Context, *google.Config) (provider.Provider, error),
+) (provider.Provider, error) {
```
inkel, line 226: (Doing the best I can with GitHub UI, so please test this locally) If I'm not mistaken, you can make this a bit more readable (IMNSHO) with generics:
```suggestion
	return selectProviderWith(ctx, cfg, aws.New, azure.New, google.New)
}

type newProviderFunc[T any, P provider.Provider] func(context.Context, T) (P, error)

func selectProviderWith(
	ctx context.Context,
	cfg *config.Config,
	newAWS newProviderFunc[*aws.Config, *aws.AWS],
	newAzure newProviderFunc[*azure.Config, *azure.Azure],
	newGCP newProviderFunc[*google.Config, *google.GCP],
) (provider.Provider, error) {
```

It's still a bit more verbose than I'd like to, but seems a bit cleaner. I'd probably move the new type definition before the definition of `selectProvider`.

inkel (commented):
Just my 2 cents 😸

inkel (approved):

PR #845, opened 2026-03-13
Listen to error on Collect and emit error count

leonorfmartins:
## Problem
`cloudcost_exporter_collector_error_total` was never [incremented](https://ops.grafana-ops.net/goto/effvx5c0mt79cf?orgId=stacks-27821), even when there were errors. This is because the prometheus.Gatherers approach would only work if collectors implemented prometheus.Gatherer (returning ([]*dto.MetricFamily, error)). The current provider.Collector.Collect has a different signature — it pushes to a channel and returns an error directly.

## Proposal
- Ditch the gatherer strategy and emit error metrics directly on `Collect`
- Rename files to match current strategy
- Rename `cloudcost_exporter_collector_error_total` to `cloudcost_exporter_collector_error` to make metric name more explicit.

## Test
- When triggering a collector error on purpose, e.g, not logged in against the provider's API:
```
cloudcost_exporter_collector_error{collector="VPC"} 4
cloudcost_exporter_collector_error{collector="cloudsql"} 4
cloudcost_exporter_collector_total{collector="VPC"} 4
cloudcost_exporter_collector_total{collector="cloudsql"} 4
```
- when there is no collector error:
```
cloudcost_exporter_collector_total{collector="NATGATEWAY"} 3
cloudcost_exporter_collector_total{collector="S3"} 3
cloudcost_exporter_collector_total{collector="VPC"} 3
cloudcost_exporter_collector_total{collector="aws_elb"} 3
cloudcost_exporter_collector_total{collector="aws_rds"} 1
```

### DIFF
--- a/pkg/collectormetrics/collectormetrics_test.go
+++ b/pkg/collectormetrics/collectormetrics_test.go
@@ -124,15 +124,6 @@ func TestCollect_MultipleCollectors(t *testing.T) {
 			},
 			expectedErrors: []bool{true, true, true},
 		},
-		"multiple collectors with mixed failures": {
-			collectors: []collectorConfig{
-				{name: "collector_1", collect: func(context.Context, chan<- prometheus.Metric) error { return assert.AnError }},
-				{name: "collector_2", collect: func(context.Context, chan<- prometheus.Metric) error { return nil }},
-				{name: "collector_3", collect: func(context.Context, chan<- prometheus.Metric) error { return nil }},
-				{name: "collector_4", collect: func(context.Context, chan<- prometheus.Metric) error { return assert.AnError }},
-			},
-			expectedErrors: []bool{true, false, false, true},
-		},
 	}
 	for name, tt := range tests {
 		t.Run(name, func(t *testing.T) {


suntala (commented):
> prometheus.Gatherers approach would only work if collectors implemented prometheus.Gatherer (returning ([]*dto.MetricFamily, error))

And why is it that we aren't pursuing the Gatherers approach? Is implementing prometheus.Gatherer not feasible?

pkg/collectormetrics/collectormetrics_test.go
```diff
@@ -0,0 +1,152 @@
...
+		},
+		"multiple collectors all fail": {
+			collectors: []collectorConfig{
+				{name: "collector_1", collect: func(context.Context, chan<- prometheus.Metric) error { return assert.AnError }},
+				{name: "collector_2", collect: func(context.Context, chan<- prometheus.Metric) error { return assert.AnError }},
+				{name: "collector_3", collect: func(context.Context, chan<- prometheus.Metric) error { return assert.AnError }},
+			},
+			expectedErrors: []bool{true, true, true},
+		},
+		"multiple collectors with mixed failures": {
```
suntala: Would it be possible to update the description here to clarify the difference between this one and `"multiple collectors some fail"`?

pkg/collectormetrics/collectormetrics_test.go
```diff
@@ -0,0 +1,152 @@
...
+	c.EXPECT().Collect(gomock.Any(), ch).Return(assert.AnError)
+
+	_, hasError := Collect(context.Background(), c, ch, slog.Default())
+	close(ch)
+
+	assert.True(t, hasError, "hasError should be true when Collect fails")
+
+	// Verify that the error counter metric was emitted to the channel.
+	// All metrics in errorCounterVec share the same Desc, so matching
+	// on Desc() is sufficient to identify the error counter family.
```
suntala: I'm wondering if this verification method might be brittle. I could imagine Desc() changing for one or more metrics eventually.

leonorfmartins: addressed in https://github.com/grafana/cloudcost-exporter/pull/845/commits/72711de39a8c1f60baedfc4dc624c7c41c3f4339 and in https://github.com/grafana/cloudcost-exporter/pull/845/commits/389bcde47378008f923994ca7daa39192d0db48b

pkg/collectormetrics/collectormetrics_test.go
leonorfmartins: since now we dont distinguish between error types on the collector, it was safe to remove this test case! Removed in https://github.com/grafana/cloudcost-exporter/pull/845/commits/0e216ec645ad5d52ac5eb9687c47ba6192718301

PR #846, opened 2026-03-13
Add MSK cluster cost metrics

Blinkuu:
## Summary

- Add a new MSK collector that exports list-price hourly cost metrics (`compute` and `storage`) for provisioned Amazon MSK clusters across all configured regions
- Refactor `PricingStore` to require explicit fetcher functions, replacing the implicit `Refresh` pattern so each collector controls its own pricing resolution
- Add documentation for MSK metrics, IAM permissions, and configuration

## Details

**New collector: `pkg/aws/msk/`**

Emits two gauge metrics per cluster:
- `cloudcost_aws_msk_compute_hourly_rate_usd_per_hour` — broker-hour list price × broker count
- `cloudcost_aws_msk_storage_hourly_rate_usd_per_hour` — provisioned EBS storage cost (broker count × volume size)

Clusters requiring unsupported pricing dimensions (serverless, Express, tiered storage) are skipped with a warning.

**Pricing store refactor**

`PricingStore.Refresh()` no longer implicitly knows how to fetch prices. Instead, callers pass fetcher functions via `WithFetchers(...)`, making the store reusable across collectors (EC2, RDS, NAT Gateway, and now MSK) without modifying its internals.

**Scope:** V1 covers provisioned clusters with local EBS storage only. Serverless, Express, tiered storage, provisioned throughput, replicator, and transfer pricing are out of scope.

### DIFF
--- a/pkg/aws/aws.go
+++ b/pkg/aws/aws.go
@@ -216,13 +216,12 @@ func newWithDependencies(ctx context.Context, config *Config, awsClient client.C
 			})
 			collectors = append(collectors, collector)
 		case serviceMSK:
-			pricingConfig := awsConfig
-			if pricingConfig.Region != "us-east-1" || config.Profile != "" || config.RoleARN != "" {
-				var err error
-				pricingConfig, err = createAWSConfig(ctx, "us-east-1", config.Profile, config.RoleARN)
-				if err != nil {
-					return nil, err
-				}
+			// The AWS Price List API is served from a small set of endpoint regions.
+			// We standardize on us-east-1 for pricing lookups; the actual priced
+			// region is selected by the GetProducts filters.
+			pricingConfig, err := createAWSConfig(ctx, "us-east-1", config.Profile, config.RoleARN)
+			if err != nil {
+				return nil, err
 			}
 			awsMSKClient := client.NewAWSClient(client.Config{
 				PricingService: awsPricing.NewFromConfig(pricingConfig),


pkg/aws/msk/msk.go
```diff
@@ -0,0 +1,342 @@
...
+	mskStorageUsageType      = "Kafka.Storage.GP2"
+)
+
+var (
+	ComputeHourlyGaugeDesc = utils.GenerateDesc(
+		cloudcostexporter.MetricPrefix,
+		subsystem,
+		"compute_hourly_rate_usd_per_hour",
+		"Hourly cost of AWS MSK broker compute by cluster. Cost represented in USD/hour",
+		[]string{"region", "cluster_name", "cluster_arn", "instance_type"},
```
Pokom, line 36: Hmm, I _think_ it's fine, but we should double that that this doesn't interfere with `cluster`, which refers to the k8s cluster.

pkg/aws/msk/msk.go
```diff
@@ -0,0 +1,342 @@
...
+			c.logger.Error("error listing MSK clusters", "region", regionName, "error", err)
+			continue
+		}
+
+		for _, cluster := range clusters {
+			c.collectCluster(ch, regionName, cluster)
+		}
+	}
+
+	return nil
```
Pokom, line 174: I'd need to refresh my memory, but it might be worthwhile to run regions in parallel. I believe we had to do that for S3/ec2, but it's been awhile since I've looked at the code.

Blinkuu: This is already parallelized in the `pricing_store.go`, specifically in the `fetchAllPrices()` method.

pkg/aws/msk/msk.go
Blinkuu: I see some other metrics with this label, so I think we're good.

pkg/aws/msk/msk.go
Pokom: The pricing is parallelized, but you could also do the same for regions. I don't think it will make much of a difference for msk clusters as we're not in every region, but we had to do this for s3/ec2 as we had ~20+ regions with 100's of clusters/buckets to discover. 

Don't over index on it, there should be metrics/dashboards to measure if the scrape gets close to 60 seconds.

PR #853, opened 2026-03-24
Build new operational dashboard with newly introduced metrics

leonorfmartins:
Build new operational dashboard for cloudcost-exporter, withe the following structure:
### Overview row

- Active Collectors — count of distinct collectors currently reporting metrics (active vs total)
- P95 Collection Duration — 95th percentile of collector run time across all clusters/collectors
- Success Rate — overall collection success rate (%)
- Error Rate by Collector (%) — time series, per-collector error rate over time
- Success Rate by Collector (%) — time series, per-collector success rate over time
- Collection Rate vs Error Rate — overlay of collection throughput vs error rate
- Version History per Environment — tracks exporter version changes per environment

### By Provider row

- P95 Duration — one stat panel each for AWS, GCP, Azure
- Success Rate — one stat panel each for AWS, GCP, Azure

### Logs row

- Error Log Rate — time series of error log volume
- Error Logs — Loki log panel filtered to errors
- All Logs (cloudcost-exporter namespace) — Loki log panel for all exporter logs

### By Collector row

- P99 / P95 / P50 Duration by Collector — three time series panels at different percentiles
- P95 Duration by Collector (current, ranked) — bar chart ranking collectors by current P95 latency
- By Region row
- P95 Top 10 Slowest Regions — time series of the slowest regions overall
- P95 Duration — GCP Regions — time series per GCP region
- P95 Duration — AWS Regions — time series per AWS region
- P95 Duration — Azure Regions — time series per Azure region

The dashboard has 4 template variables — cluster, collector, prometheus datasource and logs datasource — allowing filtering of all panels by environment and specific collector. All duration metrics come from cloudcost_exporter_collector_duration_seconds and error/success counts from cloudcost_exporter_collector_total.

Fixes https://github.com/grafana/deployment_tools/issues/413121

### DIFF
--- a/cloudcost-exporter-dashboards/dashboards/operations_dashboard.go
+++ b/cloudcost-exporter-dashboards/dashboards/operations_dashboard.go
@@ -116,8 +116,8 @@ func OperationsDashboard() *dashboard.DashboardBuilder {
 			Unit("short").
 			Thresholds(dashboard.NewThresholdsConfigBuilder().
 				Mode("absolute").
-				Steps([]dashboard.Threshold{dashboard.Threshold{Value: cog.ToPtr[float64](0), Color: "green"},
-					dashboard.Threshold{Value: cog.ToPtr[float64](80), Color: "red"}})).
+				Steps([]dashboard.Threshold{{Value: cog.ToPtr[float64](0), Color: "green"},
+					{Value: cog.ToPtr[float64](80), Color: "red"}})).
 			ColorScheme(dashboard.NewFieldColorBuilder().
 				Mode("thresholds").
 				FixedColor("#ad46ff")).
@@ -147,9 +147,9 @@ func OperationsDashboard() *dashboard.DashboardBuilder {
 			Unit("s").
 			Thresholds(dashboard.NewThresholdsConfigBuilder().
 				Mode("absolute").
-				Steps([]dashboard.Threshold{dashboard.Threshold{Value: cog.ToPtr[float64](0), Color: "green"},
-					dashboard.Threshold{Value: cog.ToPtr[float64](30), Color: "yellow"},
-					dashboard.Threshold{Value: cog.ToPtr[float64](120), Color: "red"}})).
+				Steps([]dashboard.Threshold{{Value: cog.ToPtr[float64](0), Color: "green"},
+					{Value: cog.ToPtr[float64](30), Color: "yellow"},
+					{Value: cog.ToPtr[float64](120), Color: "red"}})).
 			ColorScheme(dashboard.NewFieldColorBuilder().
 				Mode("thresholds")).
 			GraphMode("none").
@@ -181,9 +181,9 @@ func OperationsDashboard() *dashboard.DashboardBuilder {
 			Max(100).
 			Thresholds(dashboard.NewThresholdsConfigBuilder().
 				Mode("absolute").
-				Steps([]dashboard.Threshold{dashboard.Threshold{Value: cog.ToPtr[float64](0), Color: "red"},
-					dashboard.Threshold{Value: cog.ToPtr[float64](90), Color: "yellow"},
-					dashboard.Threshold{Value: cog.ToPtr[float64](95), Color: "green"}})).
+				Steps([]dashboard.Threshold{{Value: cog.ToPtr[float64](0), Color: "red"},
+					{Value: cog.ToPtr[float64](90), Color: "yellow"},
+					{Value: cog.ToPtr[float64](95), Color: "green"}})).
 			ColorScheme(dashboard.NewFieldColorBuilder().
 				Mode("thresholds")).
 			GraphMode("area").
@@ -213,7 +213,7 @@ func OperationsDashboard() *dashboard.DashboardBuilder {
 			Min(0).
 			Thresholds(dashboard.NewThresholdsConfigBuilder().
 				Mode("absolute").
-				Steps([]dashboard.Threshold{dashboard.Threshold{Value: cog.ToPtr[float64](0), Color: "green"}})).
+				Steps([]dashboard.Threshold{{Value: cog.ToPtr[float64](0), Color: "green"}})).
 			ColorScheme(dashboard.NewFieldColorBuilder().
 				Mode("palette-classic")).
 			Legend(common.NewVizLegendOptionsBuilder().
@@ -272,7 +272,7 @@ func OperationsDashboard() *dashboard.DashboardBuilder {
 			Max(100).
 			Thresholds(dashboard.NewThresholdsConfigBuilder().
 				Mode("absolute").
-				Steps([]dashboard.Threshold{dashboard.Threshold{Value: cog.ToPtr[float64](0), Color: "green"}})).
+				Steps([]dashboard.Threshold{{Value: cog.ToPtr[float64](0), Color: "green"}})).
 			ColorScheme(dashboard.NewFieldColorBuilder().
 				Mode("palette-classic")).
 			Legend(common.NewVizLegendOptionsBuilder().
@@ -336,13 +336,13 @@ func OperationsDashboard() *dashboard.DashboardBuilder {
 			Unit("reqps").
 			Thresholds(dashboard.NewThresholdsConfigBuilder().
 				Mode("absolute").
-				Steps([]dashboard.Threshold{dashboard.Threshold{Value: cog.ToPtr[float64](0), Color: "green"}})).
+				Steps([]dashboard.Threshold{{Value: cog.ToPtr[float64](0), Color: "green"}})).
 			ColorScheme(dashboard.NewFieldColorBuilder().
 				Mode("palette-classic")).
 			Overrides([]cog.Builder[dashboard.DashboardFieldConfigSourceOverrides]{dashboard.NewDashboardFieldConfigSourceOverridesBuilder().
 				Matcher(dashboard.MatcherConfig{Id: "byName", Options: "Errors/s"}).
-				Properties([]dashboard.DynamicConfigValue{dashboard.DynamicConfigValue{Id: "color", Value: map[string]interface{}{"fixedColor": "red", "mode": "fixed"}}})}).
-			WithOverride(dashboard.MatcherConfig{Id: "byName", Options: "Errors/s"}, []dashboard.DynamicConfigValue{dashboard.DynamicConfigValue{Id: "color", Value: map[string]interface{}{"fixedColor": "red", "mode": "fixed"}}}).
+				Properties([]dashboard.DynamicConfigValue{{Id: "color", Value: map[string]interface{}{"fixedColor": "red", "mode": "fixed"}}})}).
+			WithOverride(dashboard.MatcherConfig{Id: "byName", Options: "Errors/s"}, []dashboard.DynamicConfigValue{{Id: "color", Value: map[string]interface{}{"fixedColor": "red", "mode": "fixed"}}}).
 			Legend(common.NewVizLegendOptionsBuilder().
 				DisplayMode("table").
 				Placement("right").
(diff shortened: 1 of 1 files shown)


cloudcost-exporter-dashboards/dashboards/operations_dashboard.go
```diff
@@ -1,287 +1,1185 @@
...
 	"github.com/grafana/grafana-foundation-sdk/go/timeseries"
 )
 
+// Main entry point to generate the dashboard JSON, by calling builder.Build()
 func OperationsDashboard() *dashboard.DashboardBuilder {
-	builder := dashboard.NewDashboardBuilder("CloudCost Exporter Operations Dashboard").
-		// leaving this for BC reasons, but a proper human-readable UID would be better.
-		Uid("1a9c0de366458599246184cf0ae8b468").
+	builder := dashboard.NewDashboardBuilder("CloudCost Exporter operational dashboard").
+		Id(3826820004253696).
```
simaarfania, line 21: wondering if we shouldn't be hardcoding this dashboard ID?

cloudcost-exporter-dashboards/dashboards/operations_dashboard.go
```diff
@@ -1,287 +1,1185 @@
...
+				"30s",
+				"1m",
+				"5m",
+				"15m",
+				"30m",
+				"1h",
+				"2h",
+				"1d"})).
+		LiveNow(false).
+		Version(0x9).
```
simaarfania: Dashboard version is managed internally by Grafana (I believe) and will diverge the moment anyone saves the dashboard in the UI. Should this be removed?

cloudcost-exporter-dashboards/dashboards/operations_dashboard.go
```diff
@@ -1,287 +1,1185 @@
...
+			TextMode("auto").
+			ReduceOptions(common.NewReduceDataOptionsBuilder().
+				Values(false).
+				Calcs([]string{"lastNotNull"})).
+			PercentChangeColorMode("standard").
+			Orientation("auto")).
+		WithPanel(stat.NewPanelBuilder().
+			Id(0x2d).
+			Targets([]cog.Builder[variants.Dataquery]{prometheus.NewDataqueryBuilder().
+				Expr("0 * sum(rate(cloudcost_exporter_collector_total{cluster=~\"$cluster\", collector=~\"$collector\"}[$__rate_interval])) + 100").
```
simaarfania: `0 * sum(...) + 100` always evaluates to 100% regardless of what's happening. is this a placeholder?

cloudcost-exporter-dashboards/dashboards/operations_dashboard.go
```diff
@@ -1,287 +1,1185 @@
...
+		WithPanel(timeseries.NewPanelBuilder().
+			Id(0x19).
+			Targets([]cog.Builder[variants.Dataquery]{prometheus.NewDataqueryBuilder().
+				Expr("histogram_quantile(0.99, sum by(collector) (rate(cloudcost_exporter_collector_duration_seconds{cluster=~\"$cluster\", collector=~\"$collector\"}[$__rate_interval])))").
+				Range().
+				LegendFormat("{{collector}}").
+				RefId("A").
+				QueryType("range").
+				Datasource(common.DataSourceRef{Type: cog.ToPtr[string]("prometheus"), Uid: cog.ToPtr[string]("${datasource}")})}).
+			Title("P99 Duration by Collector").
```
simaarfania, line 767: Description says "P50, P95, and P99 latency percentiles" but this panel only queries P99 I believe!

cloudcost-exporter-dashboards/dashboards/operations_dashboard.go
```diff
@@ -1,287 +1,1185 @@
...
+			Id(0x11).
+			Targets([]cog.Builder[variants.Dataquery]{prometheus.NewDataqueryBuilder().
+				Expr("count(count by(collector) (rate(cloudcost_exporter_collector_duration_seconds{cluster=~\"$cluster\", collector=~\"$collector\"}[$__rate_interval])))").
+				Instant().
+				LegendFormat("Active").
+				RefId("A").
+				QueryType("instant").
+				Datasource(common.DataSourceRef{Type: cog.ToPtr[string]("prometheus"), Uid: cog.ToPtr[string]("${datasource}")}),
+				prometheus.NewDataqueryBuilder().
+					Expr("count(count by(collector) (cloudcost_exporter_collector_duration_seconds{cluster=~\"$cluster\", collector=~\"$collector\"}))").
```
simaarfania: This queries the raw metric without rate, so "Total" will include any collector series still in TSDB retention (depending on how long this is, it could mean that we are including ones that haven't been recently active). Should this use `rate(...[$__rate_interval])` to match the "Active" query above?

cloudcost-exporter-dashboards/dashboards/operations_dashboard.go
leonorfmartins: I think it should be ok even if we treat this dashboard as an entirely different one, maybe all we need is to update the runbooks? I don't want to be changing UIDs too much with the fear of creating even more inconsistencies in the database as I don't know in how many more places this UID is checked. What if we change it here but we'd also need to change it somewhere else as well?

cloudcost-exporter-dashboards/dashboards/operations_dashboard.go
leonorfmartins: we can but the goal is to make changes to this dashboard only as code right?

cloudcost-exporter-dashboards/dashboards/operations_dashboard.go
leonorfmartins: addressed in https://github.com/grafana/cloudcost-exporter/pull/853/commits/437d41e38722e0e83fb2dc1bbf1648098fe3b9d7 :)

cloudcost-exporter-dashboards/dashboards/operations_dashboard.go
leonorfmartins: that was actually my intention!... so, we sadly don't have a metric telling us which collectors should be active (total ones). We only know which ones are present today, but we don't know all of them which were ever present!
Having said that, I realised only later that the total collector metric was targeting collectors only present today which doesn't help... so I changed it to count_over_time so it counts all collectors which were ever present today or in the past for the dashboard's time range. Addressed in https://github.com/grafana/cloudcost-exporter/pull/853/commits/2633dea9bd2368027a99194d4fffd583a746e755

cloudcost-exporter-dashboards/dashboards/operations_dashboard.go
leonorfmartins: fixed, thank you!

cloudcost-exporter-dashboards/dashboards/operations_dashboard.go
```diff
@@ -1,287 +1,1180 @@
...
+			Unit("reqps").
+			Thresholds(dashboard.NewThresholdsConfigBuilder().
+				Mode("absolute").
+				Steps([]dashboard.Threshold{{Value: cog.ToPtr[float64](0), Color: "green"}})).
+			ColorScheme(dashboard.NewFieldColorBuilder().
+				Mode("palette-classic")).
+			Overrides([]cog.Builder[dashboard.DashboardFieldConfigSourceOverrides]{dashboard.NewDashboardFieldConfigSourceOverridesBuilder().
+				Matcher(dashboard.MatcherConfig{Id: "byName", Options: "Errors/s"}).
+				Properties([]dashboard.DynamicConfigValue{{Id: "color", Value: map[string]interface{}{"fixedColor": "red", "mode": "fixed"}}})}).
+			WithOverride(dashboard.MatcherConfig{Id: "byName", Options: "Errors/s"}, []dashboard.DynamicConfigValue{{Id: "color", Value: map[string]interface{}{"fixedColor": "red", "mode": "fixed"}}}).
```
simaarfania: Looks like this might be a duplicate! .`Overrides() `on line 336 already sets the same red colour for Errors/s. Think one of these can be removed :slightly_smiling_face:

cloudcost-exporter-dashboards/dashboards/operations_dashboard.go
```diff
@@ -1,287 +1,1180 @@
...
+				Tooltip(false).
+				Legend(false).
+				Viz(false)).
+			InsertNulls(common.BoolOrFloat64{Bool: cog.ToPtr[bool](false)}).
+			SpanNulls(common.BoolOrFloat64{Bool: cog.ToPtr[bool](false)}).
+			AxisBorderShow(false)).
+		WithPanel(statetimeline.NewPanelBuilder().
+			Id(0x46).
+			Targets([]cog.Builder[variants.Dataquery]{prometheus.NewDataqueryBuilder().
+				Expr("topk without(version) (1, max by(cluster, version) (cloudcost_exporter_build_info)) and on(version) topk(1, count by(version) (max by(cluster, version) (cloudcost_exporter_build_info)))").
```
simaarfania: would `and on(version) topk(1, ...)` filter to only the single most-deployed version across all clusters? If one cluster is ahead on a newer version while the rest are on the previous one, that cluster would be silently dropped from the panel. it seems like we'd be missing data during upgrades which is probably when we'd want to be using the dashboard the most?

maybe we could do `max by(cluster, version) (cloudcost_exporter_build_info)` instead?

leonorfmartins: that's correct, my intent was to have only one cluster per bar or row, so the panel would not get too big during rollouts... do you think it would help to keep both? Something like
```
prod-1 - v.2.0                       |-----------
prod-1 - v.1.0 |-----------|
```
?

I can change it to be that way if it would help!

cloudcost-exporter-dashboards/dashboards/operations_dashboard.go
leonorfmartins: well spotted! Removed the extra one in https://github.com/grafana/cloudcost-exporter/pull/853/commits/c61012793d2f3b9ff06118eb2a1abbe9c9c764b5

cloudcost-exporter-dashboards/dashboards/operations_dashboard.go
simaarfania: I think showing both rows per cluster during a rollout  would be awesome!!  it seems the simplest way to get that would be dropping the `topk` filtering entirely and just using `max by(cluster, version) (cloudcost_exporter_build_info)`!

cloudcost-exporter-dashboards/dashboards/operations_dashboard.go
```diff
@@ -1,287 +1,1177 @@
...
+			dashboard.NewDatasourceVariableBuilder("loki_datasource").
+				Name("loki_datasource").
+				Type("loki").
+				Label("Loki Datasource").
+				Hide(0),
+			dashboard.NewQueryVariableBuilder("cluster").
+				Name("cluster").
+				Label("Cluster").
+				Hide(0).
+				Query(dashboard.StringOrMap{Map: map[string]interface{}{"label": "cluster", "metric": "cloudcost_exporter_collector_duration_seconds", "qryType": 1, "query": "label_values(cloudcost_exporter_collector_duration_seconds,cluster)", "refId": "PrometheusVariableQueryEditor-VariableQuery"}}).
```
simaarfania: I think the `cluster` label isn't emitted by the exporter itself, all operational metrics only have `collector` and `region` labels. This variable works in our setup because `cluster` is added as an external label by Prometheus, but an OSS user running a standalone instance won't have it, meaning this dropdown will be empty and all dashboard queries will return no data. 

Maybe we should write somewhere somewhere (e.g. in docs/metrics/collectors.md) that the `cluster` label is expected to come from the Prometheus `external_labels` config? Otherwise I'm worried the dashboard would appear broken for some users

PR #863, opened 2026-03-27
fix(aws): skip failing RDS and VPC collectors instead of aborting provider

### LINKED ISSUE #716: AWS: Handle collector errors gracefully
Errors during initialisation of AWS collectors should be logged not returned. See
See https://github.com/grafana/cloudcost-exporter/pull/664#discussion_r2549625724

stephan-rayner:
## Summary

- RDS and VPC collector init failures previously returned `nil, err`, killing the entire AWS provider and dropping all metrics from every other service
- Aligns RDS and VPC error handling with S3 and EC2: log the error at ERROR level and `continue` so remaining collectors stay intact

Fixes #716

## Test plan

- [x] `make test` passes
- [ ] Verify that a pricing config error for RDS/VPC no longer prevents S3/EC2 metrics from being exported


stephan-rayner:
This effort spawned the discussion in #869 and later the work captured under #870

This is no longer worth reviewing as it will be handed under a broader effort

Issue #864, opened 2026-03-30
fix(azure/aks): price store fails to populate — API query times out after 15 minutes


Duologic:
## Summary

The AKS VM price store never populates successfully. The Azure Retail Prices API query times out after the 15-minute context deadline, leaving the price store empty. All VM pricing metrics are missing.

## Symptoms

```
time=2026-03-30T00:05:13.773Z level=INFO  msg="populating price store"
time=2026-03-30T00:20:13.773Z level=ERROR msg="error populating prices" err="max retries reached: failed to advance page: context deadline exceeded"
time=2026-03-30T00:20:36.493Z level=ERROR msg="region not found in price map" region=westeurope
```

- `PopulatePriceStore` starts and runs for exactly 15 minutes before hitting the context deadline
- The retry mechanism fails to recover (see below)
- The price store remains empty, causing `"region not found in price map"` errors on every scrape
- Other stores (disk, machine) populate without issue

## Root cause

The API filter in `pkg/azure/aks/price_store.go` is too broad:

```
serviceName eq 'Virtual Machines' and priceType eq 'Consumption'
```

This returns every VM-related SKU across all regions (including Cloud Services, Dedicated Hosts, Reserved instances, etc.), producing a result set too large to paginate within 15 minutes. The code then filters most of these out client-side in `validateMachinePriceIsRelevantFromSku`.

For comparison, the disk pricing store uses a more specific filter and fetches 9,010 prices in ~2 seconds.

A stricter server-side filter exists as a comment in the code (`price_store.go:159-164`):

```
serviceName eq 'Virtual Machines' and priceType eq 'Consumption' and contains(productName, 'Virtual Machines') and contains(skuName, 'Low Priority') ne true
```

This filter matches the client-side logic exactly, so it should return the same data. However, the comment (introduced by @logyball in #221 as part of the original implementation) claims this server-side filter is slower in practice. There is no evidence in the git history that this was tested — the comment shipped with the initial code. This claim needs to be validated, as the current broad filter cannot finish paginating in 15 minutes.

## Additional issues

These are not the root cause but would reduce impact if the API call fails for other reasons:

### Retries share a single expired context

All 5 `try.Do` retry attempts reuse the same `pricingCtx` created at the top of `PopulatePriceStore`. Once the 15-minute deadline expires, all retries fail instantly (~36ms total for all 5 attempts).

```go
pricingCtx, cancel := context.WithTimeout(ctx, 15*time.Minute) // created once
// ...
[...]

PR #865, opened 2026-03-31
[cloudcost-exporter] Fix: Fetching Azure pricing data times out

rooneyshuman:
### Summary
  Fixes the AKS VM price store failing to populate due to Azure Retail Prices API timeouts. The root cause was an overly broad API filter that returned every VM-related SKU across all Azure regions (too large to paginate within the timeout) leaving the price store empty and causing `region not found in price map` errors on every scrape.

### Changes in this PR

  - `pkg/azure/aks/price_store.go`
    - Replaced the global API filter with a per-region `armRegionName` equality filter built from the regions where AKS clusters exist, eliminating the timeout
    - `PopulateVMPriceStore` now accepts `regions []string` and returns `bool` to signal success or failure to the caller
    - Each retry attempt creates its own `context.WithTimeout` so retries are not instantly cancelled when a previous attempt times out
    - Added `priceRefreshRetryInterval` (30 minutes) so a failed population retries sooner than the standard 24-hour interval

  - `pkg/azure/aks/machine_store.go`
    - Added `ready` channel (closed via `sync.Once` after first population) and `Ready() <-chan struct{}` method so callers can wait for the machine store before depending on its region data
       - `ready` implies the store is usable, but the channel closes after the first population attempt regardless of outcome

  - `pkg/azure/aks/aks.go`
    - Price goroutine now waits on `machineStore.Ready()` before making its first API call, ensuring regions are available
    - Resets the refresh ticker to `priceRefreshRetryInterval` on failure, `priceRefreshInterval` on success

### Testing
Price store populates in ~15 seconds
Screenshot 2026-03-31 at 12 28 22 PM

### Issue Ref
#864

### DIFF
--- a/pkg/azure/aks/machine_store.go
+++ b/pkg/azure/aks/machine_store.go
@@ -73,8 +73,8 @@ type MachineStore struct {
 	MachineMap     map[string]*VirtualMachineInfo
 	machineMapLock *sync.RWMutex
 
-	readyOnce sync.Once
-	ready     chan struct{}
+	initialPopulationOnce sync.Once
+	initialPopulation     chan struct{}
 }
 
 func NewMachineStore(parentCtx context.Context, parentLogger *slog.Logger, azClientWrapper client.AzureClient) (*MachineStore, error) {
@@ -91,7 +91,7 @@ func NewMachineStore(parentCtx context.Context, parentLogger *slog.Logger, azCli
 		MachineMap:     make(map[string]*VirtualMachineInfo),
 		machineMapLock: &sync.RWMutex{},
 
-		ready: make(chan struct{}),
+		initialPopulation: make(chan struct{}),
 	}
 
 	// Populate before using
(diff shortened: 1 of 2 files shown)


pkg/azure/aks/machine_store.go
```diff
@@ -72,6 +72,9 @@ type MachineStore struct {
 
 	MachineMap     map[string]*VirtualMachineInfo
 	machineMapLock *sync.RWMutex
+
+	readyOnce sync.Once
+	ready     chan struct{}
```
jjo: small naming nit: iiuc this will be `ready` for "single use" (even with a possible failure, leading to 30min retry)?
maybe `initialized` or `initialPopulationDone` ?

rooneyshuman: I see what you're saying - maybe `initialized` is still too similar to something like `fully ready`. I can update it to `initialPopulationDone`, though that's pretty wordy. We could do `initialPopulation` and with it's method `Done()`? still doesn't necessarily imply success that way and is already idiomatically Go (eg, `ctx.Done()`)

jjo: hm well, tbh `ready` is fine also -- maybe just adding a comment there regarding its "single-shot" nature?

PR #883, opened 2026-04-07
Add Azure Event Hubs cost collector

Blinkuu:
## Summary
- add an Azure Event Hubs collector that exports hourly compute and storage cost metrics for Standard Kafka-compatible namespaces
- extend the Azure client with Event Hubs namespace listing and Azure Monitor metrics queries for namespace usage
- wire the collector into the Azure provider under `EVENTHUBS` and `EVENTHUB`, and document the new metrics in `docs/metrics/azure/eventhubs.md`

## New metrics

| Metric | Description |
|--------|-------------|
| `cloudcost_azure_eventhubs_compute_hourly_rate_usd_per_hour` | Hourly compute cost per namespace, including throughput units, the Standard Kafka endpoint charge, and estimated ingress |
| `cloudcost_azure_eventhubs_storage_hourly_rate_usd_per_hour` | Hourly storage overage cost per namespace above the included `84 GB` per throughput unit allowance |

Labels: `region`, `namespace_name`, `namespace`, `sku`

## Scope

V1 covers Standard namespaces with fixed throughput capacity only. It skips Premium, Dedicated, and auto-inflate namespaces.

Compute folds estimated ingress into the compute metric. The estimate is derived from Azure Monitor `IncomingMessages` and `IncomingBytes` using `max(messages, ceil(bytes / 64 KB))` per minute. Storage uses the `Size` metric and the regional Blob Storage Hot LRS list price to estimate retention overage above the included allowance.

A follow-up TODO is left in code to split ingress into its own metric later without changing current dashboards.

## Test plan
- [x] `go test ./pkg/azure/eventhubs ./pkg/azure/client ./pkg/azure -count=1`
- [x] `tmp_home=$(mktemp -d) && HOME="$tmp_home" AWS_CONFIG_FILE="$tmp_home/config" AWS_SHARED_CREDENTIALS_FILE="$tmp_home/credentials" go test ./... -count=1`


pkg/azure/eventhubs/eventhubs.go
```diff
@@ -0,0 +1,558 @@
...
+	billableIngressEventBytes          = 64 * 1000
+	includedStorageGBPerThroughputUnit = 84
+	bytesPerGB                         = 1000 * 1000 * 1000
+)
+
+var (
+	ComputeHourlyGaugeDesc = utils.GenerateDesc(
+		cloudcostexporter.MetricPrefix,
+		subsystem,
+		"compute_hourly_rate_usd_per_hour",
```
simaarfania, line 60: Wondering if we should get rid of the `_hourly` suffix as it can be redunant! The project pattern is typically `instance_total_usd_per_hour,` so `compute_usd_per_hour `and `storage_usd_per_hour` might fit better 👍

pkg/azure/eventhubs/eventhubs.go
```diff
@@ -0,0 +1,558 @@
...
+	eventHubsMetricNamespace = "Microsoft.EventHub/namespaces"
+	incomingBytesMetricName  = "IncomingBytes"
+	incomingMessagesMetric   = "IncomingMessages"
+	sizeMetricName           = "Size"
+
+	metricsLookback = time.Hour
+	metricsInterval = "PT1M"
+	maxMetricsBatch = 50
+
+	billableIngressEventBytes          = 64 * 1000
```
simaarfania, line 51: I believe existing Azure collectors use binary units (e.g. `MiBsToGiB` = 1024). If that convention applies here, the billing boundary should be 64 * 1024 (65,536 bytes) rather than 64 * 1000 I think!

pkg/azure/eventhubs/eventhubs.go
Blinkuu: I think the original reason I named it this way was because I based the AWS MSK collector implementation on the AWS RDS collector and it uses an "hourly" prefix. From what I see there are plenty of other metrics that use the `hourly` prefix/infix. I'd rather not change this to keep it consistent, but would be happy to see it updated across all metrics on a new major release (breaking change).

pkg/azure/eventhubs/eventhubs.go
Blinkuu: The documentation specifically talks about kilobytes (KB) not kibibytes (KiB) here: 

https://learn.microsoft.com/en-us/azure/event-hubs/event-hubs-faq#how-are-ingress-events-calculated-

I'm not sure how we can validate it other than running this new collector and eventually comparing if our cost estimations differ a lot from an actual cloud bill.

simaarfania (approved):
Looks good to me besides the merge conflicts! Feel free to resolve those and release!

Issue #870, opened 2026-04-02
Fix: Standardise Collector Error Propagation (Option B)


stephan-rayner:
## Background

Several collectors silently swallow errors during scrapes: they log a warning, skip the affected resource, and return `nil` from `Collect()`. From the exporter's perspective the scrape succeeded, so `cloudcost_exporter_collector_error` never increments and SLO dashboards show the collector as healthy. Hundreds of VMs or volumes can go unpriced with no alertable signal.

The `collectormetrics` wrapper in `pkg/gatherer` already increments the error counter whenever `Collect()` returns a non-nil error - the infrastructure for observable failures exists. The gap is that collectors need to surface errors rather than absorb them.

> **Note:** This issue is about operational metrics - metrics that describe the health of the exporter itself (e.g. `cloudcost_exporter_collector_scrape_errors_total`) - not the cost rate metrics that collectors emit.

---

## Option Definitions (from #869)

**Option A - Return Error, Provider Logs and Skips**
At init, `New()` returns an error and the provider logs it and skips the collector. At scrape, `Collect()` returns an error and the provider logs it. Other collectors are unaffected at both phases. Failures are only observable via logs.

**Option B - Return Error, Log, Skip, and Increment a Metric** <- *this issue*
Same as A, but the provider also increments an error counter labelled by collector name. If the EC2 collector fails to initialize, the AWS provider logs the error, skips EC2, and increments `cloudcost_exporter_collector_init_errors_total{collector="ec2"}`. An alert can fire on that counter without any log monitoring.

**Option C - Never Fail, Defer All Errors**
`New()` always succeeds. `Collect()` re-emits stale cached values or serves background-refreshed data rather than returning an error. A broken collector silently appears healthy.

**Option D - Fail Fast, Fail the Provider or Scrape**
Any collector failure fails the entire provider at init, or fails the entire scrape at scrape time.

---

## Problem

Several collectors currently implement Option C: they swallow errors in `Collect()`, return `nil`, and the `cloudcost_exporter_collector_error` counter never increments. SLO dashboards show these collectors as healthy even when they silently fail to price resources.

The `collectormetrics.Collect()` wrapper in `pkg/gatherer` already increments the error counter when `Collect()` returns a non-nil error. The gap is entirely on the collector side: collectors need to return errors rather than swallowing them.

---

### Scrape Phase - Option C to B

These collectors currently swallow per-item errors and always return `nil` from `Collect()`:

[...]

PR #884, opened 2026-04-07
fix(natgateway): Surface Pricing Store Errors at Init and Scrape

stephan-rayner:
## Summary

Advances #870 (partial — NAT Gateway and pricing store only).

Two gaps in NAT Gateway error visibility:

- `New()` previously swallowed a pricing store init failure and returned a
  collector with an empty map and no error. It now returns an error, which
  `aws.go` logs and uses to skip the collector.
- Background pricing refresh failures were silent. `Collect()` now emits a
  `cloudcost_aws_natgateway_pricing_last_refresh_error` gauge (1 on failure,
  0 on success) so the last refresh state is queryable without relying on logs.

`pricingstore.NewPricingStore` now returns `(*PricingStore, error)`. On init
failure it returns the empty store and the error; callers decide whether to
proceed. MSK is updated to match the new signature and logs the error.

## Changes

- `pkg/aws/pricingstore/pricing_store.go` — `NewPricingStore` returns
  `(*PricingStore, error)` instead of `*PricingStore`
- `pkg/aws/natgateway/natgateway.go` — `New` returns `(*Collector, error)`;
  adds `lastRefreshErr atomic.Bool` and `PricingRefreshErrorDesc` gauge
- `pkg/aws/aws.go` — handles the new `natgateway.New` error return; logs and
  skips the collector on failure
- `pkg/aws/msk/msk.go` — updated for new `NewPricingStore` signature; logs
  error on init failure

## Out of Scope

Azure AKS scrape-phase fix and the first-`Collect()` error for background-init
collectors are tracked in #870 and will follow in separate PRs.

## Test Plan

- [x] `make test` passes
- [x] `TestNew_ReturnsErrorWhenPricingAPIUnavailable` covers the `New()` error path
- [x] `TestCollector_Collect` asserts `PricingRefreshErrorDesc` is emitted with
  value `0.0` on a clean scrape
- [x] `TestCollector_Describe` asserts three descriptors including
  `PricingRefreshErrorDesc`
- [ ] `TestCollect_EmitsRefreshErrorGaugeAsOneAfterBackgroundFailure` asserts
  `PricingRefreshErrorDesc` is emitted with value `1.0` after a background
  refresh failure
- [ ] `TestNewPricingStore_ReturnsStoreAndErrorWhenFetchFails` asserts
  `NewPricingStore` returns a non-nil empty store alongside the error


pkg/aws/natgateway/natgateway.go
```diff
@@ -93,14 +111,24 @@ func (c *Collector) Regions() []string { return c.regions }
...
 func (c *Collector) Collect(ctx context.Context, ch chan<- prometheus.Metric) error {
 	c.logger.LogAttrs(ctx, slog.LevelInfo, "calling collect")
 
+	lastRefreshErr := c.lastRefreshErr.Load()
+
+	refreshErrVal := 0.0
+	if lastRefreshErr {
+		refreshErrVal = 1.0
+	}
+	ch <- prometheus.MustNewConstMetric(PricingRefreshErrorDesc, prometheus.GaugeValue, refreshErrVal)
```
leonorfmartins: I don't think we need to emit any particular metric here. All collector metrics are being captured in https://github.com/grafana/cloudcost-exporter/blob/main/pkg/collectormetrics/collectormetrics.go and automatically used in the dashboard.
Or do you think there should be a new metric specific for refresh errors?

pkg/aws/aws.go
```diff
@@ -276,6 +295,10 @@ func (a *AWS) Describe(ch chan<- *prometheus.Desc) {
 }
 
 func (a *AWS) Collect(ch chan<- prometheus.Metric) {
+	for _, name := range a.failedCollectors {
+		ch <- prometheus.MustNewConstMetric(collectorInitErrorDesc, prometheus.GaugeValue, 1.0, subsystem, name)
```
leonorfmartins, line 299: any error emitted by collectors will be captured in https://github.com/grafana/cloudcost-exporter/blob/main/pkg/collectormetrics/collectormetrics.go as well. Do we need a new metric for initialisation?

pkg/aws/natgateway/natgateway.go
stephan-rayner: You are right about this 100% this can come out and it would simply the PR

stephan-rayner:
This PR was principally made for demo and discussion purposes. Having served its purpose I am going to close it now.

PR #885, opened 2026-04-08
fix(azure/aks): Tighten VM Price Store API Filter to Prevent Timeout

stephan-rayner:
Addresses: #864

The Azure Retail Prices API filter was very broad. It returned Cloud Services, Dedicated Hosts, Reserved instances, and Low Priority SKUs alongside standard consumption SKUs. Paginating the result set took longer than the 15-minute context deadline, leaving the VM price store empty and causing "region not found in price map" errors on every scrape.

This moves the two checks already enforced client-side by `validateMachinePriceIsRelevantFromSku` into the OData query itself:
- contains(productName, 'Virtual Machines') drops non-VM product SKUs
- contains(skuName, 'Low Priority') ne true drops spot SKUs

The API now returns only what the client-side filter would have kept, hopefully this will reduce the pagination when compared with the original result set. `validateMachinePriceIsRelevantFromSku` is kept as a defensive check.

## Azure Retail Prices API Documentation

  - [Azure Retail Prices API](https://learn.microsoft.com/en-us/rest/api/cost-management/retail-prices/azure-retail-prices)

  This page covers the filter syntax, supported OData operators (`eq`, `ne`, `contains()`, `and`/`or`), and pagination behaviour, all directly
  relevant to the changes in #865 and #885.

  The `NOT` operator is not supported by this API. This is why the Low Priority exclusion uses `contains(skuName, 'Low Priority') ne true` instead of
  `not contains(skuName, 'Low Priority')`.

### DIFF
--- a/pkg/azure/aks/vm_price_store_test.go
+++ b/pkg/azure/aks/vm_price_store_test.go
@@ -74,7 +74,6 @@ func TestPopulateVMPriceStore(t *testing.T) {
 				{ArmSkuName: "Standard_D4s_v3", SkuName: "D4s v3", ArmRegionName: "westus", ProductName: "Virtual Machines D Series", RetailPrice: 0.1},
 				{ArmSkuName: "Standard_D8s_v3", SkuName: "D8s v3", ArmRegionName: "centraleurope", ProductName: "Virtual Machines D Series", RetailPrice: 0.1},
 				{ArmSkuName: "Standard_D16s_v3", SkuName: "D16s v3 Spot", ArmRegionName: "centraleurope", ProductName: "Virtual Machines D Series", RetailPrice: 0.01},
-				{ArmSkuName: "Standard_D4s_v3", SkuName: "D4s v3 Low Priority", ArmRegionName: "centraleurope", ProductName: "Virtual Machines D Series", RetailPrice: 0.01}, // low priority machines are disregarded
 			},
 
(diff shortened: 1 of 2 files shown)


rooneyshuman (changes requested):
If we're filtering for this when we make the API call, we don't need [this](https://github.com/grafana/cloudcost-exporter/blob/main/pkg/azure/aks/vm_price_store.go#L169-L183) check post-fetch anymore, since it would now be redundant

stephan-rayner:
> If we're filtering for this when we make the API call, we don't need [this](https://github.com/grafana/cloudcost-exporter/blob/main/pkg/azure/aks/vm_price_store.go#L169-L183) check post-fetch anymore, since it would now be redundant

That is a fair point, how about I I add the removal of that to this PR so we can see what it will look like 😄

pkg/azure/aks/vm_price_store_test.go
rooneyshuman:
```suggestion
				SkuName: "Standard_D16s_v5",
```

for completion, since `Low Priority` machines are now filtered at the API level, let's replace this skuName with a valid OnDemand machine type we would get returned

stephan-rayner: Love it!

rooneyshuman (approved):

Issue #869, opened 2026-04-01
RFC: Standardize Collector Error Handling


stephan-rayner:
## Background

Collector error handling is inconsistent across the codebase at both initialization and scrape time. The design intent is:

- A failing collector should not block other collectors from the same provider
- A provider can fail if none of its collectors initialize successfully
- Failures must be observable without relying solely on log monitoring

In practice, 13 collectors implement at least four distinct behaviours at each phase, and there is no agreed standard. This issue exists to reach consensus on what that standard should be across both phases.

> **Note:** This issue is about operational metrics - metrics that describe the health of the exporter itself (e.g. `cloudcost_exporter_collector_scrape_errors_total`) - not the cost rate metrics that collectors emit (e.g. `cloudcost_aws_ec2_instance_cpu_usd_per_core_hour`).

---

### Initialization

| Collector | `New()` returns error? | What happens on failure |
|---|---|---|
| AWS EC2 | YES | Returns error; provider logs & skips |
| AWS S3 | YES | Partial - bad regions logged as warning, no error returned |
| GCP GCS | YES | Returns error; provider logs & skips |
| GCP GKE | YES | Returns error; provider logs & skips |
| GCP CLB | YES | Returns error; provider logs & skips |
| GCP Cloud SQL | YES | Returns error; provider logs & skips |
| Azure AKS | YES | Returns error; **provider fails entirely** |
| AWS RDS | NO | Never fails on init - defers all errors to scrape time |
| AWS NAT Gateway | NO | Never fails on init - background refresh, errors logged |
| AWS ELB | NO | Never fails on init - defers all errors to scrape time |
| AWS MSK | NO | Never fails on init - background refresh, errors logged |
| AWS VPC | NO | Swallows init error, logs it, always returns a collector |
| GCP VPC | NO | Swallows init error, logs it, always returns a collector |

**Four Patterns in Practice:**
1. Returns error -> provider logs & skips the collector (6 collectors)
2. Returns error -> provider fails entirely (1 collector: Azure AKS)
3. Never fails on init -> defers to scrape or background refresh (4 collectors)
4. Swallows init error -> logs it but always returns a healthy-looking collector (2 collectors)

### Scrape Time

**Quick Clarification of "Partial metrics on failure" Column**:
[...]

PR #887, opened 2026-04-08
fix(aws/pricingstore): Propagate Init Error from `NewPricingStore` and `VPC` to Callers

stephan-rayner:
## What

`NewPricingStore` and the init of `VPC` previously swallowed initialization errors, logging them and returning a store that would never have data. This change makes the store initialization return an error so callers can decide how to handle it.

The goal here is to move toward a standardized way of handling pricing store errors ahead of refining this to align with the consense reach in #869. That refinement will be in future issues.

## Changes

- `NewPricingStore` returns `(*PricingStore, error)` instead of `*PricingStore`
- `natgateway.New`, `msk.New`, and `vpc.New` propagate the error to their callers
- `aws.New` logs collector init failures and skips the failed collector, consistent with the existing silent-failure design for collectors
- `natgateway` logs errors from periodic pricing refreshes instead of ignoring them
- MSK no longer creates a separate pricing client for `us-east-1`; it uses the injected `client.Client`
- Tests cover the error path in `NewPricingStore` and `natgateway.New`

## Why

Silent failures mean a collector runs but never emits metrics, which is harder to diagnose than a startup error. Surfacing the error at init time gives operators a clear signal in the logs.


pkg/aws/msk/msk.go
```diff
@@ -273,6 +276,10 @@ func buildClusterPricingData(cluster msktypes.Cluster) (clusterPricingData, erro
 	}, nil
 }
 
+// newPriceFetcher returns a PriceFetchFunc for MSK pricing lookups.
```
stephan-rayner, line 284: I felt this comment made a bit more sense here than in `aws.go` I am open to reconsidering this though

pkg/aws/msk/msk_test.go
```diff
@@ -229,6 +232,7 @@ func TestCollectorCollectContinuesWhenRegionListingFails(t *testing.T) {
 		Client: pricingClient,
 		Logger: testLogger(),
 	})
+	require.NoError(t, err)
```
leonorfmartins, line 281: now that `New` can fail, should we add a text case testing for the error scenario as well?

pkg/aws/msk/msk.go
```diff
@@ -105,13 +105,16 @@ type clusterPricingData struct {
...
 
-func New(ctx context.Context, config *Config) *Collector {
+func New(ctx context.Context, config *Config) (*Collector, error) {
 	logger := slog.Default()
 	if config.Logger != nil {
 		logger = config.Logger.With("logger", serviceName)
 	}
 
-	pricingStore := pricingstore.NewPricingStore(ctx, logger, config.Regions, newPriceFetcher(config.Client))
+	pricingStore, err := pricingstore.NewPricingStore(ctx, logger, config.Regions, newPriceFetcher(config.Client))
```
leonorfmartins, line 116: this is outside of this PR's context, but I'm puzzled if fetching this API for `msk` resources even work? For other resources we are using the fixed `us-wast-1` region (see RDS and VPC), but in here, we are taking the client's region.
According to AWS docs, the pricing API is only available in [these regions](https://docs.aws.amazon.com/general/latest/gr/billing.html), so we might need to use the fixed `us-east-1` regions instead of `config.Regions`

stephan-rayner: Oh that is a good call, I'll dig into that!. It is funny because it seems to be working presently, but we could have gotten lucky. Ya I will investigate this and if it is small I will fold it into this PR

pkg/aws/msk/msk_test.go
stephan-rayner: I don't know how I missed that. Yes, yes I will do that

pkg/aws/msk/msk.go
stephan-rayner: This is now addressed as of cfebafd079ee0ec2974d48617c0e497c3a42b8d6

pkg/aws/msk/msk_test.go
stephan-rayner: This is now addressed as of 566b5400947e96d74576368681ecdcab80dafe45

pkg/aws/msk/msk.go
leonorfmartins: ok, nevermind, I got confused! I think this is correct. The pricing API needs a hardcoded region, but what's in play here is fetching the prices for all the regions where we have resources, which needs the config region. The AI could explain it better for me:

The pricingStore at line 114 is correctly defined. Here's the full picture:

Two different "region" concepts are in play:

Pricing API endpoint (pricingConfig in aws.go:226) — hardcoded to us-east-1. This is correct: the AWS Price List API endpoint only exists in a few regions. This is the client stored in config.Client.

Regions as filter parameters — config.Regions (all enabled regions) passed to NewPricingStore. The pricing store iterates over these regions and calls fetchPrices(ctx, regionName) for each, which translates to ListMSKServicePrices → makeServiceInput → GetProductsInput with regionCode = regionName as a filter field ([pricing.go:129-140](vscode-webview://0859vfiumbin3md1chvpjc7tqgikb1b1og7pimffo1lvq978d1sf/pkg/aws/client/pricing.go#L129-L140)).

leonorfmartins (approved):

PR #891, opened 2026-04-09
feat(AWS): Add an `account_id` label to every AWS cost metric across all seven collectors, account_id is resolved once at provider startup via STS `GetCallerIdentity` and propagated to each collector.

### LINKED ISSUE #890: Add `account_id` to exported AWS metrics
### Cloud Provider

AWS

### Feature Request

There's an inconsistency regarding AWS (missing) `account_id` label when compared with GCP:

* GCP
  * querying `count(count by (__name__)({__name__=~"cloudcost_.+", project!=""}))` -> `13` (i.e. 13 metrics w/`project` label
* AWS: 
  * querying `count(count by (__name__)({__name__=~"cloudcost_.+", account_id!=""}))` -> _nil_

jjo:
## What

Add an `account_id` label to every AWS cost metric across all seven collectors (EC2, S3, RDS, NAT Gateway, ELB, VPC, MSK). The account ID is resolved once at provider startup via STS `GetCallerIdentity` and propagated to each collector.

Fixes: #890.

## Why

Multiple deployments serving different AWS accounts (e.g. via role assumption) produce indistinguishable time series. An `account_id` label enables aggregation and filtering by account, matching the pattern GCP already uses with its `project` label.

## How

- `pkg/aws/aws.go`: added `AccountID` field to `Config`, call `sts.GetCallerIdentity` in `New()` before creating collectors, pass `config.AccountID` to every collector constructor
- `pkg/aws/`: propagate `config.AccountID` downstream to every collector
- `pkg/aws/s3/s3_test.go`, `pkg/aws/natgateway/natgateway_test.go`: updated test assertions for new label
- `docs/metrics/aws/*.md`: documented `account_id` label in all 7 metric docs
- _unrelated_: `go fmt ./...` (touching other files)

## Result

Operational metrics (`cloudcost_exporter_*`) are unchanged — `account_id` only applies to cost metrics (`cloudcost_*`).

Every AWS cost metric now carries `account_id="<...>"`. No new API permissions required (`sts:GetCallerIdentity` is allowed by default for all IAM identities). No breaking changes to existing PromQL neither cardinality (the new label is additive).

--
Signed-off-by: JuanJo Ciarlante <jjo@grafana.com>


pkg/aws/aws.go
```diff
@@ -122,6 +123,14 @@ func New(ctx context.Context, config *Config) (*AWS, error) {
 
 	regions = filterExcludedRegions(regions, config.ExcludeRegions)
 
+	// Resolve the AWS account ID from the active credentials.
+	stsClient := sts.NewFromConfig(ac)
+	identity, err := stsClient.GetCallerIdentity(ctx, &sts.GetCallerIdentityInput{})
+	if err != nil {
+		return nil, fmt.Errorf("resolving AWS account ID: %w", err)
+	}
+	config.AccountID = *identity.Account
```
rooneyshuman:
```suggestion
	config.AccountID = aws.ToString(identity.Account)
```
potential nil deref here - we can use [aws.ToString()](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/aws#ToString) just in case

rooneyshuman (approved):

pkg/aws/aws.go
jjo: ah nice! _TIL_, committing

rooneyshuman (approved):
🚀

Issue #54, opened 2023-05-05
[Object Storage]: Collect and export actual cost data for Azure object storage


Pokom:
When we initially started `cloudcost-exporter`, our focus was on GCP and then AWS with Azure as a very far distant priority. Before we go live with `cloudcost-exporter` we _should_ implement exporting Azure Object Storage costs.

jjo:
Possibly happening in Q3 together with compute metrics effort on Azure, depending on #[140127](https://github.com/grafana/deployment_tools/issues/140127).

Duologic:
considering #140127 is done, unblock

Pokom:
This is a good first issue for someone that would want to learn more about how to build a collector from scratch that isn't time sensitive. I've added appropriate labels and will update the issue with appropriate documentation and requirements.

PR #893, opened 2026-04-10
Azure: Add blob collector stub

suntala:
For https://github.com/grafana/cloudcost-exporter/issues/54

Adds an Azure blob cost collector behind `-azure.services blob`, including metrics wiring that follows the same gatherer pattern as AKS (cost vec collected via Collect rather than separate registration). Azure startup now skips collectors that fail to construct—logging the error and continuing—matching AWS behavior so one bad service does not stop the whole provider.

How to check it locally:

First run the exporter:

```
go run ./cmd/exporter/exporter.go \
  -provider azure \
  -azure.subscription-id "$(az account show --query id -o tsv)" \
  -azure.services blob \
  -log.level=info \
  -log.type=text
```

Then scrape the metrics:
```
curl -s localhost:8080/metrics | grep -E 'azure_blob|collector='
```

### DIFF
--- a/cmd/exporter/exporter.go
+++ b/cmd/exporter/exporter.go
@@ -224,19 +224,6 @@ func selectProvider(ctx context.Context, cfg *config.Config) (provider.Provider,
 
 type newProviderFunc[T any] func(context.Context, T) (provider.Provider, error)
 
-// expandAzureServices flattens repeated -azure.services values and comma-separated tokens (parity with AWS/GCP flags).
-func expandAzureServices(flags config.StringSliceFlag) []string {
-	var out []string
-	for _, s := range flags {
-		for _, part := range strings.Split(s, ",") {
-			if t := strings.TrimSpace(part); t != "" {
-				out = append(out, t)
-			}
-		}
-	}
-	return out
-}
-
 func selectProviderWith(
 	ctx context.Context,
 	cfg *config.Config,
@@ -256,7 +243,7 @@ func selectProviderWith(
 			Logger:           cfg.Logger,
 			SubscriptionId:   cfg.Providers.Azure.SubscriptionId,
 			ScrapeInterval:   cfg.Collector.ScrapeInterval,
-			Services:         expandAzureServices(cfg.Providers.Azure.Services),
+			Services:         strings.Split(cfg.Providers.Azure.Services.String(), ","),
 			CollectorTimeout: collectorTimeout,
 		})
 	case "aws":
--- a/cmd/exporter/exporter.go
+++ b/cmd/exporter/exporter.go
@@ -75,7 +75,7 @@ func providerFlags(fs *flag.FlagSet, cfg *config.Config) {
 	fs.Var(config.NewDeprecatedStringSliceFlag(&cfg.Providers.GCP.Projects, &cfg.Providers.GCP.BucketProjectsDeprecated), "gcp.bucket-projects", "GCP project(s). (deprecated: use --gcp.projects instead)")
 	fs.Var(&cfg.Providers.AWS.Services, "aws.services", "AWS service(s).")
 	fs.Var(&cfg.Providers.AWS.ExcludeRegions, "aws.exclude-regions", "AWS region(s) to exclude from cost collection.")
-	fs.Var(&cfg.Providers.Azure.Services, "azure.services", "Azure service(s).")
+	fs.Var(&cfg.Providers.Azure.Services, "azure.services", "Azure service(s): AKS, blob (comma-separated and/or repeat flag; case-insensitive).")
 	fs.Var(&cfg.Providers.GCP.Services, "gcp.services", "GCP service(s).")
 	flag.StringVar(&cfg.Providers.AWS.Region, "aws.region", "", "AWS region")
 	flag.StringVar(&cfg.Providers.AWS.RoleARN, "aws.roleARN", "", "Optional AWS role ARN to assume for cross-account access.")
@@ -224,6 +224,19 @@ func selectProvider(ctx context.Context, cfg *config.Config) (provider.Provider,
 
 type newProviderFunc[T any] func(context.Context, T) (provider.Provider, error)
 
+// expandAzureServices flattens repeated -azure.services values and comma-separated tokens (parity with AWS/GCP flags).
+func expandAzureServices(flags config.StringSliceFlag) []string {
+	var out []string
+	for _, s := range flags {
+		for _, part := range strings.Split(s, ",") {
+			if t := strings.TrimSpace(part); t != "" {
+				out = append(out, t)
+			}
+		}
+	}
+	return out
+}
+
 func selectProviderWith(
 	ctx context.Context,
 	cfg *config.Config,
@@ -243,7 +256,7 @@ func selectProviderWith(
 			Logger:           cfg.Logger,
 			SubscriptionId:   cfg.Providers.Azure.SubscriptionId,
 			ScrapeInterval:   cfg.Collector.ScrapeInterval,
-			Services:         cfg.Providers.Azure.Services,
+			Services:         expandAzureServices(cfg.Providers.Azure.Services),
 			CollectorTimeout: collectorTimeout,
 		})
 	case "aws":
--- a/cmd/exporter/exporter.go
+++ b/cmd/exporter/exporter.go
@@ -242,6 +242,7 @@ func selectProviderWith(
 		return newAzure(ctx, &azure.Config{
 			Logger:           cfg.Logger,
 			SubscriptionId:   cfg.Providers.Azure.SubscriptionId,
+			ScrapeInterval:   cfg.Collector.ScrapeInterval,
 			Services:         cfg.Providers.Azure.Services,
 			CollectorTimeout: collectorTimeout,
 		})
--- a/docs/metrics/azure/blob.md
+++ b/docs/metrics/azure/blob.md
@@ -2,7 +2,7 @@
 
 ## Status
 
-The `BLOB` service is registered when you pass `BLOB` in `--azure.services`. The collector **does not query Azure Cost Management yet**. One storage cost `GaugeVec` is **registered** and appears in `Describe`, but **there are no samples** because **`Collect` does not call `Set`** until that integration exists.
+The `BLOB` service is registered when you pass `BLOB` in `--azure.services`. The collector **does not query Azure Cost Management yet**. The storage cost `GaugeVec` appears in `Describe` but is **not** `MustRegister`'d on the root registry (same pattern as `azure_aks`); **`Collect` forwards** `StorageGauge.Collect(ch)` to the parent Azure gatherer. **There are no samples** until **`Collect` calls `Set`** on label values (not implemented yet).
 
 ## Planned behavior (see issue #54):
 
@@ -20,6 +20,7 @@ The `BLOB` service is registered when you pass `BLOB` in `--azure.services`. The
 ### Provider operational metrics
 
 - **`cloudcost_exporter_collector_*` with `collector="azure_blob"`** — same mechanism as other Azure collectors (e.g. `azure_aks`) via `pkg/azure/azure.go` and the shared gatherer pattern.
+- **Registry:** Blob cost `GaugeVec`s are not registered separately on the Prometheus registry; `Describe` lists them for the parent Azure collector, and **`Collect` forwards** `StorageGauge.Collect(ch)` so metric descriptors are not duplicated.
 
 ## Cost metrics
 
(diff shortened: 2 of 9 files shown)


pkg/azure/blob/blob.go
```diff
@@ -0,0 +1,99 @@
...
+	cloudcost_exporter "github.com/grafana/cloudcost-exporter"
+)
+
+const subsystem = "azure_blob"
+
+// metrics holds Prometheus collectors for blob cost rates. Vectors are not registered on the root registry;
+// Azure's top-level Collector gathers them via Collect → GaugeVec.Collect (same pattern as pkg/azure/aks).
+type metrics struct {
+	StorageGauge *prometheus.GaugeVec
+	// Planned future work: operation request rate (parity with S3/GCS cloudcost_*_operation_by_location_usd_per_krequest).
```
leonorfmartins: we might already have that with existing metrics, see https://ops.grafana-ops.net/goto/ffio5ap805csgf?orgId=stacks-27821

but it depends what kind of operation rate we are talking about :)

pkg/azure/azure.go
```diff
@@ -90,17 +92,28 @@ func New(ctx context.Context, config *Config) (*Azure, error) {
...
 			}
 			collectors = append(collectors, collector)
+		case strings.EqualFold(svc, "blob"):
+			collector, err := blob.New(&blob.Config{
+				Logger:         logger,
+				SubscriptionId: config.SubscriptionId,
+				ScrapeInterval: config.ScrapeInterval,
+			})
+			if err != nil {
+				return nil, err
```
leonorfmartins: this is an interesting case and one we are actively discussing in https://github.com/grafana/cloudcost-exporter/issues/869: we had just one case, I think it would be fine to return an error, but because now we have 2 cases, if one of them errors out, the other collector will not get initialised as well.
Having said that, we should be able to throw an error there and collect in in metrics (not in this PR), but I dont think we should return the error as that will block the other collector

pkg/azure/blob/blob.go
suntala: Good to know. I'll keep that in mind when digging into the next metric. Thanks!

pkg/azure/azure.go
suntala: Good point. Made the change here: [1647cba](https://github.com/grafana/cloudcost-exporter/pull/893/commits/1647cba5a5ab04c6f472c546ef3b0a3131121498)

pkg/azure/azure.go
```diff
@@ -90,15 +92,32 @@ func New(ctx context.Context, config *Config) (*Azure, error) {
 		return nil, err
 	}
 
-	// Collector Registration
+	// Collector Registration (--azure.services matching is case-insensitive).
 	for _, svc := range config.Services {
-		switch strings.ToUpper(svc) {
-		case "AKS":
+		svc = strings.TrimSpace(svc)
```
rooneyshuman, line 97: do we need to `TrimSpace()` here again? I think the new `expandAzureServices()` already trims it before the services are passed to `azure.New()`

cmd/exporter/exporter.go
```diff
@@ -242,7 +255,8 @@ func selectProviderWith(
 		return newAzure(ctx, &azure.Config{
 			Logger:           cfg.Logger,
 			SubscriptionId:   cfg.Providers.Azure.SubscriptionId,
-			Services:         cfg.Providers.Azure.Services,
+			ScrapeInterval:   cfg.Collector.ScrapeInterval,
+			Services:         expandAzureServices(cfg.Providers.Azure.Services),
```
rooneyshuman: why not just use the inline pattern used by the other providers? eg, `strings.Split(cfg.Providers.Azure.Services.String(), ","),`

docs/metrics/azure/blob.md
```diff
@@ -0,0 +1,31 @@
...
+- **Labels:** S3 uses `region` and `class` for storage (and `tier` for operations); GCS uses `location` and `storage_class` (and operation class labels for ops). Azure label sets would follow whatever dimensions the Cost Management query returns; the above collectors show the cross-cloud “by location + storage class/tier” pattern only.
+
+### Azure Cost Management
+
+- **Query API:** Azure documentation mentions `POST https://management.azure.com/{scope}/providers/Microsoft.CostManagement/query` and states that `{scope}` includes **`/subscriptions/{subscriptionId}/`** for subscription scope. Sample responses in that snapshot include cost columns such as **`PreTaxCost`** alongside dimensions like **`ResourceLocation`** in example filters.
+- **Permissions:** Azure documentation states that you must have **Cost Management permissions at the appropriate scope** to use Cost Management APIs (the captured article links to Microsoft’s assign-access documentation for details). No specific Azure role name is asserted here.
```
rooneyshuman: I'd remove future work plans from this doc. The docs is meant to describe what functionality currently exists. We could add this info to the issue though!

docs/metrics/azure/blob.md
```diff
@@ -0,0 +1,31 @@
+# Azure Blob Storage metrics
+
+## Status
+
+The Blob Collector is registered when you pass `blob` in `--azure.services` (matching is case-insensitive). The collector **does not query Azure Cost Management yet**. The storage cost `GaugeVec` appears in `Describe` but is **not** `MustRegister`'d on the root registry (same pattern as `azure_aks`); **`Collect` forwards** `StorageGauge.Collect(ch)` to the parent Azure gatherer. **There are no samples** until **`Collect` calls `Set`** on label values (not implemented yet).
```
rooneyshuman:
```suggestion
Pass blob to -azure.services to enable this collector (matching is case-insensitive).
```
Let's use active voice here. We can iterate on this doc, let's not add what the collector _doesn't_ do

docs/metrics/azure/blob.md
```diff
@@ -0,0 +1,31 @@
...
+- **`cloudcost_exporter_collector_*` with `collector="azure_blob"`** — same mechanism as other Azure collectors (e.g. `azure_aks`) via `pkg/azure/azure.go` and the shared gatherer pattern.
+- **Registry:** Blob cost `GaugeVec`s are not registered separately on the Prometheus registry; `Describe` lists them for the parent Azure collector, and **`Collect` forwards** `StorageGauge.Collect(ch)` so metric descriptors are not duplicated.
+
+## Cost metrics
+
+| Metric name | Type | Labels | Samples |
+|-------------|------|--------|---------|
+| `cloudcost_azure_blob_storage_by_location_usd_per_gibyte_hour` | Gauge | `region`, `class` | None; exporter does not populate this metric yet (no billing API calls in `Collect`) |
+
+**Planned future work:** `cloudcost_azure_blob_operation_by_location_usd_per_krequest` (Gauge; `region`, `class`, `tier`) — parity with S3/GCS operation metrics; commented out in `pkg/azure/blob/blob.go` until operation pricing is implemented. How we group cost rows into labels is TBD when we add the API integration.
```
rooneyshuman: same comment here on including future work

cmd/exporter/exporter.go
stephan-rayner: **Context**:
There are two ways in which this could be argued to be better:

If we get whitespace in the list split would do this `"thing, blob" => "thing", " blob"` while this option would do `"thing, blob" => "thing", "blob"`. But since latter in this flow we trim white space this isn't a huge save.

If for some reason `cfg.Providers.Azure.Services == ""` then this will be mapped to [""] in the string split option and `nil` in this option. In the split option this will lead to `"unknown service"` being logged eventually... In the case of `nil` being returned it would be a no-op (could be made to send a metric or log or something more communicative)

**Suggestion**:
AWS and GCP already use the string split approach and we don't have problems there, so lets do that, it also means we can pull the test out. 

**Future**:
That said, I think this would be better served as a helper utility we would incorporate across the codebase to avoid the extra confusing log line (because `""` isn't an unknown service it is no service). I think the utility-ification of this could be done in a future PR.

stephan-rayner (changes requested):

pkg/azure/azure.go
suntala: True, we don't need it twice. I'm removing `expandAzureServices()` based on the following comment so I'll leave it as a check before the switch. Thanks for the review!

cmd/exporter/exporter.go
suntala: Thank you both for the reviews. 

Sounds good. Changed it here: [e5ed442](https://github.com/grafana/cloudcost-exporter/pull/893/commits/e5ed44294126dc9d2535c2026bc780c255b9af9a)[e5ed442](https://github.com/grafana/cloudcost-exporter/pull/893/commits/e5ed44294126dc9d2535c2026bc780c255b9af9a). Also added a check for `""` so that we don't enter the switch at all for this edge case (and therefore no "unkown service" logging).

docs/metrics/azure/blob.md
suntala: My intention was to outline my approach rather than introducing a future work plans section that would permanently live in the documentation. I wanted the intended approach to be reviewed in addition to the code. I see now that it should have gone in the PR description. Removed the section as you suggested.

pkg/azure/blob/blob.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	// Planned future work: register operation cost per 1k requests (labels region, class, tier) when Cost Management dimensions support it.
+	// m.OperationsGauge = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+	// 	Name: prometheus.BuildFQName(cloudcost_exporter.MetricPrefix, subsystem, "operation_by_location_usd_per_krequest"),
+	// 	Help: "Operation cost of blob objects by region, class, and tier. Cost represented in USD/(1k req). No samples until Cost Management is integrated.",
+	// },
+	// 	[]string{"region", "class", "tier"},
+	// )
+
```
rooneyshuman:
```suggestion
```
the [styleguide](https://github.com/grafana/cloudcost-exporter/blob/main/AGENTS.md#code-patterns) specifically prohibits commented-out code. Since this repo is open-source, let's keep only actual functionality committed

pkg/azure/blob/blob.go
rooneyshuman:
```suggestion
```
the [styleguide](https://github.com/grafana/cloudcost-exporter/blob/main/AGENTS.md#code-patterns) specifically prohibits commented-out code. Since this repo is open-source, let's keep only actual functionality committed

docs/metrics/azure/blob.md
```diff
@@ -0,0 +1,11 @@
...
+
+Pass `blob` in `--azure.services` to enable this collector. Matching is case-insensitive.
+
+The collector defines a storage cost `GaugeVec` that the Azure provider includes in its `Describe` and `Collect` fan-out (same gatherer pattern as `azure_aks`). The parent Azure collector forwards `StorageGauge.Collect(ch)` so blob cost metrics share one registration path with the rest of the Azure exporter. Scrape instrumentation publishes `cloudcost_exporter_collector_*` with label `collector="azure_blob"`.
+
+## Cost metrics
+
+| Metric name | Metric type | Description | Labels |
+|-------------|-------------|-------------|--------|
+| cloudcost_azure_blob_storage_by_location_usd_per_gibyte_hour | Gauge | Storage cost rate for Blob Storage by region and class. Cost represented in USD/(GiB*h) | |
```
rooneyshuman: should there be something in the `labels` column?

pkg/azure/blob/blob_test.go
```diff
@@ -0,0 +1,86 @@
...
+	for d := range ch {
+		got = append(got, d.String())
+	}
+	require.Len(t, got, 1)
+	joined := strings.Join(got, " ")
+	prefix := cloudcost_exporter.MetricPrefix + "_azure_blob_"
+	assert.Contains(t, joined, prefix+"storage_by_location_usd_per_gibyte_hour")
+}
+
+// #TODO: remove when we have more functionality in place
```
rooneyshuman: Tests should grow with the implementation, not be planned for removal. What's the value these `TODO: remove ...` tests provide now? Why remove them later?

pkg/azure/blob/blob.go
suntala: Thanks for the review! Good to know about the styleguide.

docs/metrics/azure/blob.md
suntala: It could go either way since the implementation isn't complete. I'll add them.

pkg/azure/blob/blob_test.go
suntala: The idea is that once more of the implementation is in place, more complex tests will check the functionality that these two small tests cover. They might be entirely new tests and I didn't want to forget to remove these smaller ones. I can make a note elsewhere though.

stephan-rayner (approved):

rooneyshuman (approved):
🚀

PR #900, opened 2026-04-13
fix: msk collector fanout by region

### LINKED ISSUE #895: AWS MSK collector: sequential region iteration with shared timeout causes intermittent scrape failures
See [Slack thread](https://raintank-corp.slack.com/archives/C07SUF6SESF/p1775833468192719) for initial context.

## Summary

The AWS MSK collector iterates through all regions **sequentially** under a shared collector timeout (default 1 minute). When any region's MSK endpoint is temporarily slow to respond, the sequential loop exceeds the timeout budget and the entire scrape fails with `context deadline exceeded`.

## Observed behaviour

Intermittent errors in the `aws_msk` collector:

```
level=ERROR msg="error listing MSK clusters" provider=aws logger=msk region=eu-north-1 error="operation error Kafka: ListClustersV2, https response error StatusCode: 0, RequestID: , canceled, context deadline exceeded"
level=ERROR msg="could not collect metrics" provider=aws collector=aws_msk message="context deadline exceeded"
```

The failing region rotates across different regions on different days (`eu-north-1`, `ap-northeast-3`, `ap-east-2`, `ca-west-1`, `ap-southeast-4`, etc.), confirming this is not a region-specific issue but a transient network slowness problem that can affect any region.

## Root cause

In [`pkg/aws/msk/msk.go`](https://github.com/grafana/cloudcost-exporter/blob/main/pkg/aws/msk/msk.go), `Collect()` iterates regions sequentially and passes the parent context directly to each `ListClustersV2` call:

```go
for _, region := range c.regions {
    select {
    case <-ctx.Done():
        return ctx.Err()  // fails entire collector if deadline exceeded
    default:
    }
    clusters, err := regionClient.ListMSKClusters(ctx)  // parent ctx passed directly
    ...
}
```

If one region's endpoint hangs (e.g. TCP connection that doesn't get a response), it consumes the entire shared timeout. The next iteration's `ctx.Done()` check then fires, returning an error for the whole collector.

By contrast, all other AWS collectors (`aws_ec2`, `aws_rds`, `NATGATEWAY`, etc.) track per-region and run in parallel, so a single slow region doesn't affect the others.

## Suggested fix

Wrap each `ListClustersV2` call in its own short sub-context (e.g. 10–15 seconds per region), independent of the outer collector timeout:

```go
regionCtx, cancel := context.WithTimeout(ctx, 15*time.Second)
clusters, err := regionClient.ListMSKClusters(regionCtx)
cancel()
if err != nil {
    c.logger.Error("error listing MSK clusters", "region", regionName, "error", err)
    continue  // slow/unavailable region no longer blocks the rest
}
```

This matches the intent of the existing `continue` on error — the goal is clearly to skip bad regions, but the shared context means a hanging region can still kill the whole scrape before the error is even returned.

## Impact

Intermittent gaps in MSK cost metrics. The collector works correctly most of the time (~14s average scrape across all regions) but fails periodically when any region is temporarily slow.

stephan-rayner:
Fixes: https://github.com/grafana/cloudcost-exporter/issues/895

## Problem

The MSK collector iterated regions sequentially, passing the shared parent context directly to each `ListMSKClusters` call. A slow or hanging region consumed the entire collector timeout budget. The `ctx.Done()` guard at the top of the next iteration then fired, returning an error for the whole collector, even though the problem was isolated to a single region.

## Solution

Fan out `ListMSKClusters` calls concurrently using `sync.WaitGroup`, one goroutine per region. A slow or erroring region no longer blocks the others. Context cancellation propagates directly through each `ListMSKClusters(ctx)` call, so the explicit `ctx.Done()` guard is no longer needed.

This matches the fan-out pattern used by the EC2 collector.

## Test Plan

- [x] All existing MSK tests pass
- [x] New test `TestCollectorCollectContinuesOnContextDeadlineExceeded` verifies that a `context.DeadlineExceeded` error from one region does not suppress metrics from another

### DIFF
--- a/pkg/aws/msk/msk.go
+++ b/pkg/aws/msk/msk.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"log/slog"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
@@ -147,13 +148,8 @@ func New(ctx context.Context, config *Config) (*Collector, error) {
 func (c *Collector) Collect(ctx context.Context, ch chan<- prometheus.Metric) error {
 	snapshot := c.pricingStore.Snapshot()
 
+	var wg sync.WaitGroup
 	for _, region := range c.regions {
-		select {
-		case <-ctx.Done():
-			return ctx.Err()
-		default:
-		}
-
 		if region.RegionName == nil || *region.RegionName == "" {
 			c.logger.Warn("skipping region with empty name")
 			continue
@@ -166,17 +162,20 @@ func (c *Collector) Collect(ctx context.Context, ch chan<- prometheus.Metric) er
 			continue
 		}
 
-		clusters, err := regionClient.ListMSKClusters(ctx)
-		if err != nil {
-			c.logger.Error("error listing MSK clusters", "region", regionName, "error", err)
-			continue
-		}
-
-		for _, cluster := range clusters {
-			c.collectCluster(ch, snapshot, regionName, cluster)
-		}
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			clusters, err := regionClient.ListMSKClusters(ctx)
+			if err != nil {
+				c.logger.Error("error listing MSK clusters", "region", regionName, "error", err)
+				return
+			}
+			for _, cluster := range clusters {
+				c.collectCluster(ch, snapshot, regionName, cluster)
+			}
+		}()
 	}
-
+	wg.Wait()
 	return nil
 }
 
--- a/pkg/aws/msk/msk.go
+++ b/pkg/aws/msk/msk.go
@@ -176,7 +176,7 @@ func (c *Collector) Collect(ctx context.Context, ch chan<- prometheus.Metric) er
 		}()
 	}
 	wg.Wait()
-	return nil
+	return ctx.Err()
 }
 
 func (c *Collector) collectCluster(ch chan<- prometheus.Metric, snapshot pricingstore.Snapshot, region string, cluster msktypes.Cluster) {
--- a/pkg/aws/client/mocks/client.go
+++ b/pkg/aws/client/mocks/client.go
@@ -16,9 +16,9 @@ import (
 
 	types "github.com/aws/aws-sdk-go-v2/service/ec2/types"
 	types0 "github.com/aws/aws-sdk-go-v2/service/elasticloadbalancingv2/types"
-	types3 "github.com/aws/aws-sdk-go-v2/service/kafka/types"
-	types1 "github.com/aws/aws-sdk-go-v2/service/pricing/types"
-	types2 "github.com/aws/aws-sdk-go-v2/service/rds/types"
+	types1 "github.com/aws/aws-sdk-go-v2/service/kafka/types"
+	types2 "github.com/aws/aws-sdk-go-v2/service/pricing/types"
+	types3 "github.com/aws/aws-sdk-go-v2/service/rds/types"
 	client "github.com/grafana/cloudcost-exporter/pkg/aws/client"
 	prometheus "github.com/prometheus/client_golang/prometheus"
 	gomock "go.uber.org/mock/gomock"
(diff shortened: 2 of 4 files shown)


pkg/aws/msk/msk.go
```diff
@@ -166,17 +162,20 @@ func (c *Collector) Collect(ctx context.Context, ch chan<- prometheus.Metric) er
...
-		for _, cluster := range clusters {
-			c.collectCluster(ch, snapshot, regionName, cluster)
-		}
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			clusters, err := regionClient.ListMSKClusters(ctx)
+			if err != nil {
+				c.logger.Error("error listing MSK clusters", "region", regionName, "error", err)
+				return
```
leonorfmartins, line 171: If we log an error, shouldn't we return one as well?

pkg/aws/msk/msk.go
```diff
@@ -166,17 +162,20 @@ func (c *Collector) Collect(ctx context.Context, ch chan<- prometheus.Metric) er
...
-			continue
-		}
-
-		for _, cluster := range clusters {
-			c.collectCluster(ch, snapshot, regionName, cluster)
-		}
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			clusters, err := regionClient.ListMSKClusters(ctx)
```
leonorfmartins, line 168: I wonder if it would be better to have a separate context per region instead of relying on the parent one. Something like
```
  regionCtx, cancel := context.WithTimeout(ctx, 15*time.Second)
  clusters, err := regionClient.ListMSKClusters(regionCtx)
  cancel()
```

that way each region would have its own context to fetch resources and we would make sure no context hangs if a TCP connection for some reason isn't terminated.

nikimanoledaki (commented):
Leaving some feedback for how we handle the collect call's errors to match the new standard 👍

pkg/aws/msk/msk.go
nikimanoledaki: +1 that we should update the approach to match Option B, as tracked here: https://github.com/grafana/cloudcost-exporter/issues/870

Since this PR already modifies this part of the code, @stephan-rayner we could already add this here. WDYT?

simaarfania (commented):
These generated file changes (Azure AKS region map, disk SKUs) appear unrelated to the MSK fix. Were these intentionally included?

stephan-rayner:
> These generated file changes (Azure AKS region map, disk SKUs) appear unrelated to the MSK fix. Were these intentionally included?

These were generated as a result of me running `make test` locally to try and debug CI. I assumed they must have been related. I'm going to roll them back. Thanks @simaarfania 

All the unrelated files are now removed!

pkg/aws/msk/msk.go
stephan-rayner: I think you are on to something here. This would be a novel approach when compared to the rest of the collectors. To that end I think the best spot is probably in client.Client so that it can be standardized across the collectors. I created https://github.com/grafana/cloudcost-exporter/issues/906 so we can explore how we would do this more broadly there

leonorfmartins: it's fine to fix this in another issue, so it's standardised in every collector 👍 thanks for creating the issue!

pkg/aws/msk/msk.go
stephan-rayner: In a sequential loop, logging and returning makes sense as the error is fatal to the whole operation. In a fan-out model, a single region failing shouldn't suppress metrics from the others./

Returning an error here would signal a collector level failure to the gatherer, which would treat all regions as failed rather than just the one. I think Logging and continuing is the right tradeoff for now. In theory operators see the error in logs (we are getting better at this), healthy regions still emit metrics. 

Longer term, I think the best solution would be bumping an error counter metric with a region tag would be a cleaner signal than either a log or a returned error. I created an [issue](https://github.com/grafana/cloudcost-exporter/issues/907) to track that as a follow-up

pkg/aws/msk/msk.go
```diff
@@ -166,17 +162,20 @@ func (c *Collector) Collect(ctx context.Context, ch chan<- prometheus.Metric) er
...
+				return
+			}
+			for _, cluster := range clusters {
+				c.collectCluster(ch, snapshot, regionName, cluster)
+			}
+		}()
 	}
-
+	wg.Wait()
 	return nil
```
simaarfania: previously, we returned `ctx.Err()` when the context was cancelled between regions. `collectormetrics.Collect` uses the error return to increment the error counter metric and log a scrape failure. I think with this change a cancelled scrape will now silently appear as a clean success, we should potentially change that!

stephan-rayner: Good catch this is now fixed!

simaarfania (approved):

PR #910, opened 2026-04-16
chore: bump Helm chart to appVersion 0.33.0 and trigger release workflow

simaarfania:
## Summary

- Bumps Helm chart `appVersion` from `0.25.0` to `0.33.0` and `version` from `1.1.2` to `1.1.3` to bring the gh-runners in line with the current app release.
- Moves the `Chart.yaml` bump logic from `update-helm-chart-on-release.yml` into `release-on-pr-merge.yml` as a new `update-helm-chart` job. The previous workflow was triggered by `release: published` which never fires when GoReleaser uses `GITHUB_TOKEN` — so it had zero runs since it was added.
- Deletes `update-helm-chart-on-release.yml` as it is now redundant.
- Adds a `trigger-deployment-tools-update` job to `release-helm-chart.yaml` that fires the `chartfile-version-update` Argo workflow in `platform-monitoring-cd` after the chart is published, so deployment_tools is updated automatically on every release.

## Why

The Helm chart has not been updated since `v0.25.0` due to broken automation. This PR closes the gap and wires up the full release chain so image and Helm chart rollouts stay at parity going forward.

## Breaking changes

None. The Helm chart templates and values are unchanged since `1.1.2`. Releases `v0.26.0` through `v0.33.0` consist of new collectors, bug fixes, and dependency updates. The only notable addition is an `account_id` label on AWS cost metrics introduced in `v0.33.0`, which is additive and does not affect existing dashboards or recording rules.


nikimanoledaki (approved):
thank you so much for debugging and fixing this! it's so much better as another job in the same workflow 👍

PR #914, opened 2026-04-17
maint: Standardize Collector Config Structs, Constructors, and Error Handling

stephan-rayner:
Standardizes collector constructors, `Config` structs, and error handling across AWS, GCP, and Azure collectors.

## Changes

**Constructor signatures**
- Standardize all collector constructors to `New(ctx context.Context, cfg *Config, ...) (*Collector, error)`.
- Applies to: `s3`, `rds`, `elb` (AWS), `gcs` (GCP), `blob` (Azure).

**Structured logging**
- Add `Logger *slog.Logger` to collector `Config` structs for `s3`, `rds`, `elb`, `gcs`, and `aks`.
- Replace package-level `slog` calls with instance loggers scoped with a `"collector"` attribute.

**Naming**
- Rename `SubscriptionId` to `SubscriptionID` per Go acronym conventions across `azure`, `blob`, and `aks`.
- Rename `elb.Config.RegionClients` to `RegionMap` for consistency with other collectors.
- Capitalize `rds.serviceName` from `"rds"` to `"RDS"`.

**Error handling**
- Change AWS provider to log and skip collectors that fail to initialize (`rds`, `elb`, `vpc` pricing config), rather than returning an error that kills the whole provider. This matches the existing pattern for other AWS collectors.

**Test fixes**
- Fix race condition in GCS tests by registering the gRPC service before calling `Serve`.

### DIFF
--- a/pkg/aws/ec2/ec2.go
+++ b/pkg/aws/ec2/ec2.go
@@ -76,14 +76,13 @@ type Collector struct {
 type Config struct {
 	ScrapeInterval time.Duration
 	Regions        []ec2Types.Region
-	Logger         *slog.Logger
 	RegionMap      map[string]client.Client
 	AccountID      string
 }
 
 // New creates an ec2 collector
-func New(ctx context.Context, config *Config) (*Collector, error) {
-	logger := config.Logger.With("collector", "ec2")
+func New(ctx context.Context, config *Config, logger *slog.Logger) (*Collector, error) {
+	logger = logger.With("collector", "ec2")
 	computeMap := NewComputePricingMap(logger, config)
 	storageMap := NewStoragePricingMap(logger, config)
 
--- a/pkg/aws/ec2/ec2.go
+++ b/pkg/aws/ec2/ec2.go
@@ -83,7 +83,7 @@ type Config struct {
 
 // New creates an ec2 collector
 func New(ctx context.Context, config *Config) (*Collector, error) {
-	logger := config.Logger.With("logger", "ec2")
+	logger := config.Logger.With("collector", "ec2")
 	computeMap := NewComputePricingMap(logger, config)
 	storageMap := NewStoragePricingMap(logger, config)
 
--- a/pkg/aws/elb/elb.go
+++ b/pkg/aws/elb/elb.go
@@ -57,7 +57,7 @@ type Config struct {
 	// PricingClient must be a client configured for us-east-1: the AWS Pricing API
 	// is only available in us-east-1 and ap-south-1.
 	PricingClient client.Client
-	RegionClients map[string]client.Client
+	RegionMap     map[string]client.Client
 	Logger        *slog.Logger
 	AccountID     string
 }
@@ -87,17 +87,17 @@ type elbProduct struct {
 	}
 }
 
-func New(config *Config) *Collector {
-	logger := config.Logger.With("logger", serviceName)
+func New(_ context.Context, config *Config) (*Collector, error) {
+	logger := config.Logger.With("collector", serviceName)
 	return &Collector{
 		regions:            config.Regions,
 		ScrapeInterval:     config.ScrapeInterval,
 		pricingClient:      config.PricingClient,
-		awsRegionClientMap: config.RegionClients,
+		awsRegionClientMap: config.RegionMap,
 		logger:             logger,
 		pricingMap:         NewELBPricingMap(logger),
 		accountID:          config.AccountID,
-	}
+	}, nil
 }
 
 func (c *Collector) Register(_ provider.Registry) error {
--- a/pkg/aws/elb/elb.go
+++ b/pkg/aws/elb/elb.go
@@ -58,7 +58,6 @@ type Config struct {
 	// is only available in us-east-1 and ap-south-1.
 	PricingClient client.Client
 	RegionMap     map[string]client.Client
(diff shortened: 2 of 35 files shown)


pkg/aws/elb/elb.go
```diff
@@ -52,7 +52,7 @@ type Collector struct {
 type Config struct {
 	ScrapeInterval time.Duration
 	Regions        []ec2Types.Region
-	RegionClients  map[string]client.Client
```
stephan-rayner: The rename was to bring it in line with all of the other collectors and what they call this field.

pkg/aws/ec2/ec2.go
```diff
@@ -83,7 +83,7 @@ type Config struct {
 
 // New creates an ec2 collector
 func New(ctx context.Context, config *Config) (*Collector, error) {
-	logger := config.Logger.With("logger", "ec2")
+	logger := config.Logger.With("collector", "ec2")
```
leonorfmartins: nothing against it, but curious why we are deciding to rename the logger label?

pkg/azure/blob/blob.go
```diff
@@ -44,20 +44,22 @@ type Collector struct {
...
 type Config struct {
 	Logger         *slog.Logger
-	SubscriptionId string
+	SubscriptionID string
 	ScrapeInterval time.Duration
 }
 
 // New builds a blob collector. It does not call Azure APIs yet; subscription and interval are stored for Cost Management integration.
-func New(cfg *Config) (*Collector, error) {
+// TODO: Add a provider client parameter (e.g. azClientWrapper) once Cost Management integration is implemented,
```
leonorfmartins: is this TODO here necessary? Maybe it's enough to create an issue for it?

pkg/google/gcp.go
```diff
@@ -90,10 +90,11 @@ func New(ctx context.Context, config *Config) (*GCP, error) {
 		var collector provider.Collector
 		switch strings.ToUpper(service) {
 		case "GCS":
-			collector, err = gcs.New(&gcs.Config{
+			collector, err = gcs.New(ctx, &gcs.Config{
 				ProjectId:      config.ProjectId,
 				Projects:       config.Projects,
 				ScrapeInterval: config.ScrapeInterval,
+				Logger:         config.Logger,
```
leonorfmartins: I'm aware some resource configs already have a logger, while others don't and that's why you are adding missing ones here. However I want to validate if we need the loggers at the config level already.

pkg/aws/rds/rds.go
```diff
@@ -54,24 +56,25 @@ type listError struct {
...
+func New(_ context.Context, config *Config) (*Collector, error) {
 	return &Collector{
 		pricingMap:     newPricingMap(),
 		regions:        config.Regions,
 		regionMap:      config.RegionMap,
 		scrapeInterval: config.ScrapeInterval,
 		Client:         config.Client,
 		accountID:      config.AccountID,
-	}
+		logger:         config.Logger.With("collector", serviceName),
```
leonorfmartins: we should guard this and check for error as well, if by any reason logger is nil, `New` will panic. Same thing for s3 and elb

pkg/aws/ec2/ec2.go
stephan-rayner: On main we are split roughly 50/50 between logger and collector, so I figured I should pick one. When I thought about from the perspective what it looks like in the log output `logger` didn't really make sense since this isn't an ec2 logger it is an ec2 collector. So when I choose to standardise I landed on `collector`

pkg/google/gcp.go
stephan-rayner: After reflecting on this I think you are right logger doesn't really make sense being in the config the signatrure should go from 

`New(ctx, cfg)` ===> `New(ctx, cfg, logger)`

pkg/aws/rds/rds.go
stephan-rayner: I think this will go away when I refactor for https://github.com/grafana/cloudcost-exporter/pull/914/changes#r3116207430

pkg/azure/blob/blob.go
stephan-rayner: I think that is a good call, I'll do that now!

pkg/azure/aks/aks.go
```diff
@@ -99,19 +99,20 @@ var (
 // Collector is a prometheus collector that collects cost metrics from AKS clusters.
 // Provides comprehensive cost tracking for both virtual machines and persistent volumes.
 type Collector struct {
-	logger *slog.Logger
+	logger         *slog.Logger
+	subscriptionID string
```
simaarfania, line 103: it seems that `subscriptionID` is stored but never read (c.subscriptionID has zero usages). maybe we can wire it into a metric label instead?

stephan-rayner: Adding `subscription_id` as a metric label is the intended goal, analogous to AWS's `account_id` and how we use it for that provider. My hope was to add this here and then wire it up in a separate PR. Specifically one that satisfies https://github.com/grafana/cloudcost-exporter/issues/922

I know it isn't in keeping with our desire (or AGENTS.md) for that matter but I think it is a sensible deviation from our standard to have a standardized standard.

How does that sit with you?

leonorfmartins (approved):

PR #915, opened 2026-04-17
[AWS ELB] Use correct region when fetching elb price locally

leonorfmartins:
Problem
When running the exporter locally for elb, we see these errors:
```
failed to get ELB pricing: operation error Pricing: GetProducts, https response error StatusCode: 0, RequestID: , request send failed, Post \"https://api.pricing.us-east-2.amazonaws.com/\": dial tcp: lookup api.pricing.us-east-2.amazonaws.com: no such host"
```

this is because the pricing API for AWS is only present in us-east-1.

- Add a dedicated us-east-1 pricing client for ELB, matching the pattern
  used by other collectors (e.g. RDS, VPC, and MSK). The AWS Pricing API is only
  available in us-east-1, so using per-region clients was incorrect.
- Scope the ELB logger with the service name (`logger.With("logger", "elb")`),
  matching the VPC collector pattern.
- Introduce `serviceName = "elb"` constant and derive `subsystem` from it.

### DIFF
--- a/pkg/aws/elb/pricing_map.go
+++ b/pkg/aws/elb/pricing_map.go
@@ -64,7 +64,7 @@ func (pm *ELBPricingMap) GetRegionPricing(region string) (*RegionPricing, error)
 	return pricing, nil
 }
 
-func (pm *ELBPricingMap) refresh(ctx context.Context, client map[string]client.Client, regions []ec2Types.Region) error {
+func (pm *ELBPricingMap) refresh(ctx context.Context, pricingClient client.Client, regions []ec2Types.Region) error {
 	pm.logger.Info("Refreshing ELB pricing data")
 
 	eg := errgroup.Group{}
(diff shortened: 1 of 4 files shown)


pkg/aws/elb/pricing_map.go
```diff
@@ -64,7 +64,7 @@ func (pm *ELBPricingMap) GetRegionPricing(region string) (*RegionPricing, error)
 	return pricing, nil
 }
 
-func (pm *ELBPricingMap) refresh(ctx context.Context, client map[string]client.Client, regions []ec2Types.Region) error {
+func (pm *ELBPricingMap) refresh(ctx context.Context, pricingClient client.Client, regions []ec2Types.Region) error {
```
simaarfania, line 67: should we add a test for this change in `refresh()`'s signature?  The existing tests only cover `NewELBPricingMap`, `SetRegionPricing`, and `GetRegionPricing`. The tests in `elb_test.go` confirm the pricing client is stored, but doesn't verify it's actually used during refresh. A test for `refresh()` or `FetchRegionPricing()` would help close this gap!

stephan-rayner: Hey @leonorfmartins, this overlaps with what I have in #914, both PRs rename `RegionClients -> RegionMap` and change the New signature in `elb.go`. Since this PR also adds the PricingClient us-east-1 fix (which #914 doesn't have), I'll rebase #914 on top of this once it merges. 🙂

If no one objects I can even add the test!

stephan-rayner: The tests are updated and still passing now 😄

stephan-rayner (approved):

PR #921, opened 2026-04-21
feat(aws): add Bedrock token pricing collector

simaarfania:
Part of https://github.com/grafana/cloudcost-exporter/issues/923. Adds a new AWS Bedrock collector that emits list-price token cost metrics.

- New `pkg/aws/bedrock` package: `Collector` implements the standard `provider.Collector` interface with `Collect`, `Describe`, `Name`, `Register`, and `Regions`
- Pricing data fetched from the AWS Pricing API (us-east-1 endpoint) and refreshed every 24 hours via the shared `pricingstore` machinery
- Three new metrics: input token cost, output token cost, and search unit cost
- Five `price_tier` values supported: `on_demand`, `on_demand_batch`, `on_demand_flex`, `on_demand_priority`, `cross_region`

## New metrics

```
cloudcost_aws_bedrock_token_input_usd_per_1k_tokens{account_id, region, model_id, family, price_tier}
cloudcost_aws_bedrock_token_output_usd_per_1k_tokens{account_id, region, model_id, family, price_tier}
cloudcost_aws_bedrock_search_unit_usd_per_1k_search_units{account_id, region, model_id, family, price_tier}
```

Cardinality: roughly 20 models x 6 regions x 5 tiers = ~600 series.

## How it works

The AWS Pricing API returns flat JSON with a `usagetype` field that encodes model and token type. The `inferenceType` attribute signals direction (input/output/search). Before the data enters the generic `pricingstore`, a preprocessing step rewrites the `usagetype` field into a composite key `family|direction|modelID|priceTier`. `Collect` splits this key back out to populate metric labels.

## IAM permissions required

```json
{
  "Effect": "Allow",
  "Action": ["pricing:GetProducts"],
  "Resource": "*"
}
```

### DIFF
--- a/pkg/aws/aws.go
+++ b/pkg/aws/aws.go
@@ -23,6 +23,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"golang.org/x/sync/errgroup"
 
+	"github.com/grafana/cloudcost-exporter/pkg/aws/bedrock"
 	"github.com/grafana/cloudcost-exporter/pkg/aws/client"
 	ec2Collector "github.com/grafana/cloudcost-exporter/pkg/aws/ec2"
 	"github.com/grafana/cloudcost-exporter/pkg/aws/elb"
@@ -90,10 +91,11 @@ const (
 	serviceS3    = "S3"
 	serviceEC2   = "EC2"
 	serviceRDS   = "RDS"
-	serviceNATGW = "NATGATEWAY"
-	serviceELB   = "ELB"
-	serviceVPC   = "VPC"
-	serviceMSK   = "MSK"
+	serviceNATGW   = "NATGATEWAY"
+	serviceELB     = "ELB"
+	serviceVPC     = "VPC"
+	serviceMSK     = "MSK"
+	serviceBedrock = "BEDROCK"
 )
 
 func New(ctx context.Context, config *Config) (*AWS, error) {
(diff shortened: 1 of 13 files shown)


pkg/aws/bedrock/bedrock.go
```diff
@@ -0,0 +1,301 @@
...
+	pricingStore pricingstore.PricingStoreRefresher
+	regions      []string
+	logger       *slog.Logger
+	accountID    string
+}
+
+func New(ctx context.Context, config *Config) (*Collector, error) {
+	logger := slog.Default()
+	if config.Logger != nil {
+		logger = config.Logger.With("logger", serviceName)
```
leonorfmartins: Can we standardise from upcoming changes in this PR and use `collector` as a label?

pkg/aws/aws.go
```diff
@@ -274,6 +276,28 @@ func newWithDependencies(ctx context.Context, config *Config, awsClient client.C
 				continue
 			}
 			collectors = append(collectors, collector)
+		case serviceBedrock:
+			// The AWS Pricing API is only available in us-east-1 and ap-south-1.
+			// Copy the already-loaded config and pin the region to us-east-1 so
+			// Bedrock pricing lookups succeed regardless of the collector's configured regions.
+			bedrockPricingConfig := awsConfig.Copy()
```
leonorfmartins, line 284: I don't think we need to have another awsConfig here, we can reuse the pricingClient being passed in this function's argument: https://github.com/grafana/cloudcost-exporter/blob/bdf763d509e631f9e20bf157f8c4a8e24b84f0c4/pkg/aws/aws.go#L140

pkg/aws/aws.go
```diff
@@ -274,6 +276,28 @@ func newWithDependencies(ctx context.Context, config *Config, awsClient client.C
 				continue
 			}
 			collectors = append(collectors, collector)
+		case serviceBedrock:
+			// The AWS Pricing API is only available in us-east-1 and ap-south-1.
+			// Copy the already-loaded config and pin the region to us-east-1 so
+			// Bedrock pricing lookups succeed regardless of the collector's configured regions.
```
jjo, line 280:
```suggestion
			// Bedrock pricing lookups succeed regardless of the collector's configured regions.
			// Note: this pins the *endpoint*, not the queried region — the collector still
			// fetches prices per configured region via a regionCode filter. See
			// pkg/aws/bedrock.go newPriceFetcher().
```

pkg/aws/bedrock/bedrock.go
```diff
@@ -0,0 +1,301 @@
...
+
+func (c *Collector) Regions() []string {
+	return c.regions
+}
+
+func (c *Collector) Register(_ provider.Registry) error {
+	return nil
+}
+
+// The AWS Pricing API is only available in us-east-1; callers must pin the client there.
```
jjo: Suggest expanding this doc as sth like:
```suggestion
// Endpoint vs. filter region. The AWS Pricing API is only served from us-east-1 and
// ap-south-1, so the *client* passed in must be pinned to one of those regions (see
// pkg/aws/aws.go where the Bedrock pricing client is created against us-east-1).
//
// The `region` argument passed into the returned PriceFetchFunc is a different thing:
// pkg/aws/client.listBedrockPrices -> listServicePrices applies it as a `regionCode`
// TermMatch filter on GetProducts, so each invocation returns only that region's SKUs.
// The pricingstore fans out one call per configured region; do NOT replace this with a
// single call, or the resulting snapshot will lose regional separation.
```

jjo (approved):
LGTM w/some non-blocking suggestions.

pkg/aws/aws.go
simaarfania: Thanks for the suggestion! The `awsClient` passed into `newWithDependencies` is constructed with `config.Region` (whatever the user configured, e.g. `eu-west-1`), so its `PricingService` isn't pinned to `us-east-1`. Using it directly would break pricing lookups in any deployment where the configured region isn't `us-east-1` sadly 😢This collector matches what MSK does as well!

Issue #923, opened 2026-04-22
feature request: add a collector for AWS Bedrock cost metrics


simaarfania:
### Cloud Provider

AWS

### Feature Request

This issue is a part of the work being discussed in Platform Monitoring squad's AI monitoring work.

## Problem

Bedrock spend is invisible below the invoice level. cloudcost-exporter covers EC2, RDS, and other compute costs but has no signal for LLM token consumption.

## Proposed solution

Add a `bedrock` collector to the AWS provider that emits list-price token cost metrics from the AWS Pricing API. Metrics cover input tokens, output tokens, and search units (Cohere Rerank) across all configured regions and pricing tiers.

**New metrics:**
- `cloudcost_aws_bedrock_token_input_usd_per_1k_tokens`
- `cloudcost_aws_bedrock_token_output_usd_per_1k_tokens`
- `cloudcost_aws_bedrock_search_unit_usd_per_1k_search_units`

## Out of scope

- Image, video, audio, cache, and guardrail SKUs
- CloudWatch usage metrics (measured spend)
- Fine-tuning, Agents, Knowledge Bases, and Guardrails pricing
- Committed use discounts or negotiated rates

PR #924, opened 2026-04-22
fix(bedrock): filter Bedrock pricing to Anthropic and Amazon families only

simaarfania:
Part of https://github.com/grafana/cloudcost-exporter/issues/923.

Adds a configurable family filter to the Bedrock pricing collector. By default it only emits metrics for `anthropic` (Claude) and `amazon` (Nova, Titan) models. The filter is controlled via a new flag:

```
--aws.bedrock.families
```

The value is a regex matched against the `family` label. The flag defaults to `.*` to emit all families.

Without filtering, the collector emits ~8,000 active series covering 113 models. The vast majority of those series represent third-party marketplace models we do not use. Filtering reduces active series by 90%.

## Changes

- `pkg/aws/bedrock/bedrock.go`: family filter compiled from config at startup and applied in `encodeBedrockPriceJSON`; defaults to `.*`
- `pkg/aws/aws.go`: `BedrockFamilyFilter` field added to `Config`, passed to the bedrock collector
- `cmd/exporter/config/config.go`: `BedrockFamilyFilter` field added to AWS provider config
- `cmd/exporter/exporter.go`: `--aws.bedrock.families` flag registered with default `.*`
- `pkg/aws/bedrock/bedrock_test.go`: updated tests to use Anthropic/Amazon providers
- `docs/metrics/aws/bedrock.md`: updated `family` description to reflect the default filter

### DIFF
--- a/pkg/aws/bedrock/bedrock.go
+++ b/pkg/aws/bedrock/bedrock.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"log/slog"
+	"regexp"
 	"strings"
 	"time"
 
@@ -71,11 +72,14 @@ type bedrockProductInfo struct {
 	} `json:"terms"`
 }
 
+const defaultFamilyFilter = "anthropic|amazon"
+
 type Config struct {
 	Regions       []ec2types.Region
 	PricingClient client.Client
 	Logger        *slog.Logger
 	AccountID     string
+	FamilyFilter  string // regex matched against the family label; defaults to "anthropic|amazon"
 }
 
 type Collector struct {
@@ -83,6 +87,7 @@ type Collector struct {
 	regions      []string
 	logger       *slog.Logger
 	accountID    string
+	familyFilter *regexp.Regexp
 }
 
 func New(ctx context.Context, config *Config) (*Collector, error) {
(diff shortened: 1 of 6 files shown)


pkg/aws/bedrock/bedrock.go
```diff
@@ -71,18 +72,22 @@ type bedrockProductInfo struct {
 	} `json:"terms"`
 }
 
+const defaultFamilyFilter = "anthropic|amazon"
```
jjo: I'd rather set this to ".*", following the principle of least-surprise (sp for non-grafanistas using CCE)

jjo (commented):
LGTM but I'd really like that generic default, and also adding some regex-ed case(s) to the tests.

pkg/aws/bedrock/bedrock.go
```diff
@@ -72,7 +72,7 @@ type bedrockProductInfo struct {
 	} `json:"terms"`
 }
 
-const defaultFamilyFilter = "anthropic|amazon"
+const defaultFamilyFilter = ".*"
```
jjo: hm given the fs.StringVar default, does it still make sense to "force" the `""` regex to fallback to this one?

jjo (approved):
LvGTM!, thanks for this :)

PR #926, opened 2026-04-23
feat(gcp/cloudsql): Retry Pricing Init on Transient API Failures

stephan-rayner:
Add exponential backoff retry to the Cloud SQL collector's \`New()\` using \`utils.Retry\`. Transient gRPC errors (\`Unavailable\`, \`DeadlineExceeded\`, \`Internal\`, \`ResourceExhausted\`) are retried up to 3 times; auth and config errors fail immediately.

Move the retryable error classifier to \`pkg/google/client\` so all GCP collectors can share it without duplication.

Fix a silent failure path where an empty SKU response from the billing API was treated as success, allowing the collector to register while producing no metrics. The empty SKU path now returns an exported \`ErrNoSKUsFound\` sentinel and is not retried. The API call succeeded but the data is absent, which retrying is unlikely to resolve.

## Tests

- \`TestNew_RetriesOnTransientError\`: verifies the collector succeeds after 2 transient failures followed by a successful call
- \`TestNew_DoesNotRetryOnAuthError\`: verifies \`PermissionDenied\` fails fast after a single attempt
- \`TestNew_DoesNotRetryOnEmptySKUResponse\`: verifies an empty SKU response fails fast with \`ErrNoSKUsFound\`

### DIFF
--- a/pkg/google/client/retry.go
+++ b/pkg/google/client/retry.go
@@ -1,0 +1,25 @@
+package client
+
+import (
+	"errors"
+
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+// IsRetryableError reports whether err represents a transient gRPC failure that
+// is worth retrying. Auth and configuration errors are not retryable.
+func IsRetryableError(err error) bool {
+	var grpcErr interface {
+		GRPCStatus() *status.Status
+	}
+	if !errors.As(err, &grpcErr) {
+		return false
+	}
+	switch grpcErr.GRPCStatus().Code() {
+	case codes.Unavailable, codes.DeadlineExceeded, codes.Internal, codes.ResourceExhausted:
+		return true
+	default:
+		return false
+	}
+}
--- a/pkg/google/cloudsql/cloudsql.go
+++ b/pkg/google/cloudsql/cloudsql.go
@@ -38,9 +38,7 @@ var (
 	initRetryAttempts     = 3
 	initRetryInitialDelay = 1 * time.Second
 	initRetryMaxDelay     = 30 * time.Second
-)
 
-var (
 	HourlyGaugeDesc = utils.GenerateDesc(
 		cloudcost_exporter.MetricPrefix,
 		subsystem,
--- a/pkg/google/cloudsql/cloudsql.go
+++ b/pkg/google/cloudsql/cloudsql.go
@@ -35,6 +35,12 @@ const (
 )
 
 var (
+	initRetryAttempts     = 3
+	initRetryInitialDelay = 1 * time.Second
+	initRetryMaxDelay     = 30 * time.Second
+)
+
+var (
 	HourlyGaugeDesc = utils.GenerateDesc(
 		cloudcost_exporter.MetricPrefix,
 		subsystem,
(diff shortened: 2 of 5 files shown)


pkg/google/client/retry.go
```diff
@@ -0,0 +1,25 @@
+package client
```
stephan-rayner, line 1: Could this be provider agnostic?

leonorfmartins: I think so, but we can do it in a later stage, if it helps :) I imagine `isRetryableError` could become an interface, implemented by each provider, for example.

pkg/google/client/retry.go
```diff
@@ -0,0 +1,25 @@
...
+import (
+	"errors"
+
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+// IsRetryableError reports whether err represents a transient gRPC failure that
+// is worth retrying. Auth and configuration errors are not retryable.
+func IsRetryableError(err error) bool {
```
leonorfmartins, line 12: We should have a unit test for this in any case

pkg/google/client/retry.go
stephan-rayner: Ya I think you may be onto something there

pkg/google/cloudsql/cloudsql.go
simaarfania: Init retries on transient errors but this path doesn't. A flaky API at refresh time drops the update silently, maybe we should consider wrapping this in the same `utils.Retry` call?

pkg/google/client/retry.go
simaarfania, line 1: should we add tests for this?

stephan-rayner: I forgot to push the latest changes. Good catch

pkg/google/cloudsql/cloudsql_test.go
```diff
@@ -132,6 +177,129 @@ func TestNew_FailsIfInitialSKUFetchFails(t *testing.T) {
...
+	require.NoError(t, err)
+
+	sqlAdminService, err := sqladmin.NewService(context.Background(), option.WithoutAuthentication(), option.WithEndpoint(sqlAdminSrv.URL))
+	require.NoError(t, err)
+
+	gcpClient := client.NewMock("test-project", 0, nil, nil, catalogClient, computeService, sqlAdminService, nil)
+	_, err = New(context.Background(), &Config{Projects: "test-project"}, slog.New(slog.NewTextHandler(os.Stdout, nil)), gcpClient)
+	require.NoError(t, err)
+	assert.Equal(t, 3, srv.callCount, "expected 3 ListServices calls (2 failures + 1 success)")
+}
```
rooneyshuman: looks like these setup steps are reused in all the new tests, can we extract this into a reusable helper?

pkg/google/cloudsql/pricingmap.go
```diff
@@ -95,8 +96,7 @@ func (pm *pricingMap) getSKus(ctx context.Context) error {
 
 	skus := pm.gcpClient.GetPricing(ctx, serviceFullName)
 	if len(skus) == 0 {
-		pm.logger.Warn("no SKUs found for Cloud SQL service", "serviceName", serviceFullName)
```
rooneyshuman, line 98: should we keep this log line?

pkg/google/cloudsql/cloudsql_test.go
stephan-rayner: I like that, I'm doing it now

pkg/google/cloudsql/pricingmap.go
stephan-rayner: If we return the error we will get a metric. But I think this is worth being clear about. Ya I've added it back!

rooneyshuman (approved):

PR #927, opened 2026-04-24
Escape line breaks in collector error logs

Blinkuu:
## Summary
- Escape collector error message line breaks before passing them to slog text output
- Update the `message` logfmt field to `error` to match the semantics
- Add regression coverage for newline and carriage-return escaping


rooneyshuman (approved):
lgtm!

PR #929, opened 2026-04-24
feat(gcp): add vertex collector

stephan-rayner:
Part of #930

## Summary

- Adds a `vertex` collector to the GCP provider
- Emits list-price token cost metrics from the GCP Billing API for Gemini and third-party Model Garden models, with a `price_tier` label covering on_demand, batch, thinking, cached, and other running modes
- Emits character cost metrics for character-billed models (e.g. translation)
- Emits custom training and prediction node compute pricing (`instance_total_usd_per_hour`) with `on_demand` and `spot` tiers
- Emits reranking cost metrics via the Cloud Discovery Engine billing service

## New Metrics

- `cloudcost_gcp_vertex_input_usd_per_1k_tokens`
- `cloudcost_gcp_vertex_output_usd_per_1k_tokens`
- `cloudcost_gcp_vertex_input_usd_per_1k_characters`
- `cloudcost_gcp_vertex_output_usd_per_1k_characters`
- `cloudcost_gcp_vertex_instance_total_usd_per_hour`
- `cloudcost_gcp_vertex_search_unit_usd_per_1k_search_units`

## Test Plan

- [x] Unit tests pass (`make test`)
- [x] Lint passes (`make lint`)
- [x] Metrics match documented names and labels in `docs/metrics/gcp/vertex.md`

### DIFF
--- a/docs/metrics/gcp/vertex.md
+++ b/docs/metrics/gcp/vertex.md
@@ -1,0 +1,110 @@
+# Vertex AI Metrics
+
+Metrics exported for the GCP Vertex AI service.
+
+## Token Pricing
+
+| Metric | Labels | Description |
+|--------|--------|-------------|
+| `cloudcost_gcp_vertex_input_usd_per_1k_tokens` | `model_id`, `family`, `region`, `price_tier` | Input cost in USD per 1k tokens, for models billed by token. Character-billed models use the character metric. |
+| `cloudcost_gcp_vertex_output_usd_per_1k_tokens` | `model_id`, `family`, `region`, `price_tier` | Output cost in USD per 1k tokens, for models billed by token. Character-billed models use the character metric. |
+
+### Labels
+
+| Label | Values | Description |
+|-------|--------|-------------|
+| `model_id` | e.g. `gemini-1.5-flash`, `gemma-4`, `llama-4-maverick` | Model name, normalised to lowercase with spaces replaced by hyphens |
+| `family` | `google`, `meta`, `alibaba`, `deepseek`, `minimax`, `moonshot`, `unknown` | Model provider family; `unknown` for unrecognised model prefixes |
+| `region` | e.g. `us-central1` | GCP region |
+| `price_tier` | see below | Running mode or pricing tier derived from the GCP SKU description |
+
+#### `price_tier` Values for Token Metrics
+
+Tiers are composed from up to three modifiers: a `thinking_` prefix, a `cached_` prefix, and a `_long_context` suffix. Not all combinations exist in GCP's SKU catalogue; only tiers with a matching SKU are emitted.
+
+**Simple tiers**
+
+| Value | Description |
+|-------|-------------|
+| `on_demand` | Standard real-time inference |
+| `batch` | Batch prediction |
+| `long_context` | Long-context window at standard priority |
+| `cached` | Context-cached input |
+| `cache_storage` | Context cache storage cost |
+| `thinking` | Extended thinking, standard priority |
+| `priority` | Priority tier |
+| `flex` | Flex tier |
+| `live` | Live (streaming) mode |
+
+**Compound tiers**
+
+| Value | Description |
+|-------|-------------|
+| `batch_long_context` | Batch + long context |
+| `priority_long_context` | Priority + long context |
+| `flex_long_context` | Flex + long context |
+| `cached_long_context` | Cached input + long context |
+| `cached_batch` | Cached input + batch |
+| `cached_flex` | Cached input + flex tier |
+| `cached_priority` | Cached input + priority tier |
+| `cached_batch_long_context` | Cached input + batch + long context |
+| `cached_flex_long_context` | Cached input + flex + long context |
+| `cached_priority_long_context` | Cached input + priority + long context |
+| `thinking_batch` | Thinking + batch |
+| `thinking_flex` | Thinking + flex tier |
+| `thinking_priority` | Thinking + priority tier |
+| `thinking_long_context` | Thinking + long context |
+| `thinking_batch_long_context` | Thinking + batch + long context |
+| `thinking_flex_long_context` | Thinking + flex + long context |
+| `thinking_priority_long_context` | Thinking + priority + long context |
+
+## Character Pricing
+
+| Metric | Labels | Description |
+|--------|--------|-------------|
+| `cloudcost_gcp_vertex_input_usd_per_1k_characters` | `model_id`, `family`, `region`, `price_tier` | Input cost in USD per 1k characters, for models billed by character (e.g. translation models). |
+| `cloudcost_gcp_vertex_output_usd_per_1k_characters` | `model_id`, `family`, `region`, `price_tier` | Output cost in USD per 1k characters, for models billed by character (e.g. translation models). |
+
+### Labels
+
+| Label | Values | Description |
+|-------|--------|-------------|
+| `model_id` | e.g. `translation-llm` | Model name, normalised to lowercase with spaces replaced by hyphens |
+| `family` | `google`, `unknown` | Model provider family; `unknown` for unrecognised model prefixes |
+| `region` | e.g. `global` | GCP region |
+| `price_tier` | `on_demand` | GCP Translation is flat-rate; no batch tier exists in the billing API |
+
+## Compute Pricing
+
+| Metric | Labels | Description |
+|--------|--------|-------------|
+| `cloudcost_gcp_vertex_instance_total_usd_per_hour` | `machine_type`, `use_case`, `region`, `price_tier` | Vertex AI custom training and prediction node cost in USD per hour |
+
+### Labels
+
+| Label | Values | Description |
+|-------|--------|-------------|
+| `machine_type` | e.g. `n1-standard-4` | Machine type used for the compute node |
+| `use_case` | `training`, `prediction` | Whether the node is used for custom training or online prediction |
+| `region` | e.g. `us-central1` | GCP region |
+| `price_tier` | `on_demand`, `spot` | Pricing tier; spot metrics are only emitted when a spot price exists |
+
+## Reranking
+
+| Metric | Labels | Description |
+|--------|--------|-------------|
+| `cloudcost_gcp_vertex_search_unit_usd_per_1k_search_units` | `model_id`, `family`, `region` | Vertex AI reranking cost in USD per 1k ranking requests |
+
+Reranking SKUs are fetched from the Cloud Discovery Engine billing service. If that service is unavailable at startup, reranking metrics are omitted and a warning is logged.
+
+### Labels
+
+| Label | Values | Description |
+|-------|--------|-------------|
+| `model_id` | e.g. `semantic-ranker-api` | Ranker model name, normalised to lowercase with spaces replaced by hyphens |
+| `family` | `google` | Model provider family; Discovery Engine reranking models are Google's |
+| `region` | e.g. `global` | GCP region |
+
+## Notes
+
+Pricing data is fetched from the GCP Billing API at startup and refreshed every 24 hours. SKU descriptions are matched using regular expressions; unknown SKUs are skipped. Verify SKU description patterns against the live Billing API when adding new models or machine types.
--- a/pkg/google/gcp.go
+++ b/pkg/google/gcp.go
@@ -18,6 +18,7 @@ import (
 	cloudcost_exporter "github.com/grafana/cloudcost-exporter"
 	"github.com/grafana/cloudcost-exporter/pkg/google/gcs"
 	"github.com/grafana/cloudcost-exporter/pkg/google/gke"
+	"github.com/grafana/cloudcost-exporter/pkg/google/vertex"
 	"github.com/grafana/cloudcost-exporter/pkg/google/vpc"
 	"github.com/grafana/cloudcost-exporter/pkg/provider"
 )
@@ -39,6 +40,7 @@ const (
 	serviceSQL          = "SQL"
 	serviceManagedKafka = "MANAGEDKAFKA"
 	serviceKafkaAlias   = "KAFKA"
+	serviceVertex       = "VERTEX"
 )
 
 // Services returns the collectors that can be enabled via -gcp.services.
@@ -52,6 +54,7 @@ func Services() []provider.ServiceInfo {
 		{Name: serviceManagedKafka, DisplayName: "Managed Kafka", Description: "Managed Service for Apache Kafka clusters", Aliases: []string{serviceKafkaAlias}},
 		{Name: serviceCLB, DisplayName: "CLB", Description: "Cloud Load Balancers via forwarding rules"},
 		{Name: serviceVPC, DisplayName: "VPC", Description: "Cloud NAT Gateway, VPN Gateway, Private Service Connect"},
+		{Name: serviceVertex, DisplayName: "Vertex AI", Description: "Vertex AI model token, character, compute, and reranking pricing"},
 	}
 }
 
@@ -203,6 +206,14 @@ func New(ctx context.Context, config *Config) (*GCP, error) {
 				Projects:       config.Projects,
 				ScrapeInterval: config.ScrapeInterval,
 			}, logger, gcpClient)
+			if err != nil {
+				logger.LogAttrs(ctx, slog.LevelError, "Error creating collector",
+					slog.String("service", service),
+					slog.String("message", err.Error()))
+				continue
+			}
+		case serviceVertex:
+			collector, err = vertex.New(ctx, logger, gcpClient)
 			if err != nil {
 				logger.LogAttrs(ctx, slog.LevelError, "Error creating collector",
 					slog.String("service", service),
--- a/pkg/google/vertex/pricing_map.go
+++ b/pkg/google/vertex/pricing_map.go
@@ -1,0 +1,385 @@
+package vertex
+
+import (
+	"context"
+	"fmt"
+	"log/slog"
+	"regexp"
+	"strings"
+	"sync/atomic"
+
+	"cloud.google.com/go/billing/apiv1/billingpb"
+	"github.com/grafana/cloudcost-exporter/pkg/google/client"
+)
+
+const (
+	vertexAIServiceName = "Vertex AI"
+	// discoveryEngineServiceName is the GCP Billing API service name for Vertex AI Search,
+	// which hosts the Ranking API used for reranking.
+	discoveryEngineServiceName = "Vertex AI Search"
+
+	// modalityGroup matches only the text modality in Vertex AI SKU descriptions. Audio,
+	// image, and video inputs are multimodal inputs priced separately from text tokens and
+	// are not represented by these metrics. Matching all modalities would cause each
+	// modality's price to overwrite the previous one, leaving the emitted value
+	// non-deterministic.
+	modalityGroup = `(?:text)`
+)
+
+var (
+	// computeRegex matches custom training/prediction compute SKU descriptions.
+	// Example: "Custom Training n1-standard-4 running in us-central1"
+	// Example: "Spot Custom Prediction n1-highmem-8 running in europe-west1"
+	// NOTE: Exact SKU description strings must be verified against the live GCP Billing API.
+	computeRegex = regexp.MustCompile(`(?i)^(Spot\s+)?Custom\s+(Training|Prediction)\s+(\S+)\s+running\s+in`)
+	// rerankRegex matches Discovery Engine Ranking API SKU descriptions.
+	// Example: "Semantic Ranker API Ranking Requests"
+	// NOTE: Exact SKU description strings must be verified against the live GCP Billing API.
+	rerankRegex = regexp.MustCompile(`(?i)^(.+?)\s+[Rr]anking\s+[Rr]equests?$`)
+)
+
+// skuPattern maps a compiled regex to the billing direction, billing type, and price tier.
+type skuPattern struct {
+	re          *regexp.Regexp
+	direction   string // "input" or "output"
+	billingType string // "token" or "char"
+	tier        string
+}
+
+func mustCompile(pattern string) *regexp.Regexp {
+	return regexp.MustCompile(strings.ReplaceAll(pattern, "{mod}", modalityGroup))
+}
+
+// skuPatterns is the ordered lookup table for Vertex AI token/character SKU descriptions.
+// More specific patterns must appear before generic ones to prevent the lazy (.+?) from
+// capturing too much.
+var skuPatterns = []skuPattern{
+	// Gemini output — "Thinking" prefix style (most specific first)
+	{mustCompile(`(?i)^(.+?)\s+Thinking\s+` + modalityGroup + `\s+Output\s+Priority\s+\(Long\)\s+-\s+Predictions$`), "output", "token", "thinking_priority_long_context"},
+	{mustCompile(`(?i)^(.+?)\s+Thinking\s+` + modalityGroup + `\s+Output\s+Priority\s+-\s+Predictions$`), "output", "token", "thinking_priority"},
+	{mustCompile(`(?i)^(.+?)\s+Thinking\s+` + modalityGroup + `\s+Output\s+Flex\s+\(Long\)\s+-\s+Predictions$`), "output", "token", "thinking_flex_long_context"},
+	{mustCompile(`(?i)^(.+?)\s+Thinking\s+` + modalityGroup + `\s+Output\s+Flex\s+-\s+Predictions$`), "output", "token", "thinking_flex"},
+	{mustCompile(`(?i)^(.+?)\s+Thinking\s+` + modalityGroup + `\s+Output\s+\(Long\)\s+-\s+Batch\s+Predictions$`), "output", "token", "thinking_batch_long_context"},
+	{mustCompile(`(?i)^(.+?)\s+Thinking\s+` + modalityGroup + `\s+Output\s+-\s+Batch\s+Predictions$`), "output", "token", "thinking_batch"},
+	{mustCompile(`(?i)^(.+?)\s+Thinking\s+` + modalityGroup + `\s+Output\s+\(Long\)\s+-\s+Predictions$`), "output", "token", "thinking_long_context"},
+	{mustCompile(`(?i)^(.+?)\s+Thinking\s+` + modalityGroup + `\s+Output\s+-\s+Predictions$`), "output", "token", "thinking"},
+
+	// Gemini output — "(Thinking On...)" parenthetical style
+	{mustCompile(`(?i)^(.+?)\s+` + modalityGroup + `\s+Output\s+Priority\s+\(Thinking\s+On\s+and\s+Long\)\s+-\s+Predictions$`), "output", "token", "thinking_priority_long_context"},
+	{mustCompile(`(?i)^(.+?)\s+` + modalityGroup + `\s+Output\s+Priority\s+\(Thinking\s+On\)\s+-\s+Predictions$`), "output", "token", "thinking_priority"},
+	{mustCompile(`(?i)^(.+?)\s+` + modalityGroup + `\s+Output\s+\(Thinking\s+On\s+and\s+Long\)\s+-\s+Batch\s+Predictions$`), "output", "token", "thinking_batch_long_context"},
+	{mustCompile(`(?i)^(.+?)\s+` + modalityGroup + `\s+Output\s+\(Thinking\s+On\s+and\s+Long\)\s+-\s+Predictions$`), "output", "token", "thinking_long_context"},
+	{mustCompile(`(?i)^(.+?)\s+` + modalityGroup + `\s+Output\s+\(Thinking\s+On\)\s+-\s+Batch\s+Predictions$`), "output", "token", "thinking_batch"},
+	{mustCompile(`(?i)^(.+?)\s+` + modalityGroup + `\s+Output\s+\(Thinking\s+On\)\s+-\s+Predictions$`), "output", "token", "thinking"},
+
+	// Gemini output — Live (before generic)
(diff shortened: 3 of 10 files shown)


pkg/google/vertex/vertex.go
```diff
@@ -0,0 +1,153 @@
...
+			case <-ticker.C:
+				if err := pm.Populate(ctx); err != nil {
+					logger.Error("failed to refresh vertex pricing", "error", err)
+				}
+			}
+		}
+	}()
+
+	projects := strings.Split(config.Projects, ",")
+	regions := client.RegionsForProjects(gcpClient, projects, logger)
```
leonorfmartins: I don't think regions is ever used, what was the initial intent of it?

pkg/google/vertex/pricing_map.go
```diff
@@ -0,0 +1,265 @@
...
+// priceFromSku extracts the unit price from a SKU's last tiered rate.
+func priceFromSku(sku *billingpb.Sku) float64 {
+	if sku == nil || len(sku.GetPricingInfo()) == 0 {
+		return 0
+	}
+	expression := sku.GetPricingInfo()[0].GetPricingExpression()
+	if expression == nil || len(expression.GetTieredRates()) == 0 {
+		return 0
+	}
+	rate := expression.GetTieredRates()[len(expression.GetTieredRates())-1].GetUnitPrice()
```
leonorfmartins, line 332: can we have a comment explaining the structure of tiered rates? It would help to visualise why we care about the last part for a tier :)

docs/metrics/gcp/vertex.md
```diff
@@ -0,0 +1,51 @@
+# Vertex AI Metrics
+
+Metrics exported for the GCP Vertex AI service.
+
+## Token Pricing
+
+| Metric | Labels | Description |
+|--------|--------|-------------|
+| `cloudcost_gcp_vertex_token_input_usd_per_1k_tokens` | `model`, `region` | Input token/character cost in USD per 1k tokens (covers Gemini, Claude, and embedding models) |
```
nikimanoledaki: One note here is that we should align on the labels.

For example, Bedrock collector has the following labels [here](https://github.com/grafana/cloudcost-exporter/blob/main/docs/metrics/aws/bedrock.md?plain=1#L3):
- account_id
- region
- model_id
- family
- price_tier

But also we need to align with Sigil [metrics](https://github.com/grafana/sigil/blob/main/docs/references/semantic-conventions.md#metrics) so that we can join them 👀 

From this [explore](https://ops.grafana-ops.net/explore?schemaVersion=1&panes=%7B%223ed%22:%7B%22datasource%22:%22edprtf91hz01se%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22sum%20by%20%28gen_ai_request_model%29%20%28gen_ai_client_token_usage_count%7B%7D%29%22,%22range%22:false,%22datasource%22:%7B%22type%22:%22prometheus%22,%22uid%22:%22edprtf91hz01se%22%7D,%22editorMode%22:%22code%22,%22legendFormat%22:%22__auto%22,%22instant%22:true,%22exemplar%22:false%7D%5D,%22range%22:%7B%22from%22:%22now-24h%22,%22to%22:%22now%22%7D,%22compact%22:false%7D%7D&orgId=1), it looks like the label for the model that Sigil uses is: `gen_ai_request_model`

We didn't have this metric dimension visibility last week since the Sigil SDK was integrated and deployed to prod only yesterday!

pkg/google/vertex/vertex.go
stephan-rayner: Good catch. `Collect()` doesn't use them directly since this is a pricing-only collector with no per-project fan-out at scrape time. I've removed the `Regions()` method since it isn't on the `provider.Collector` interface and has no callers.

I kept the `regions` and `projects` fields though. GKE, GCS, and CloudSQL all carry the same pattern, and I'd rather Vertex stay consistent with them than be the odd one out.

pkg/google/vertex/pricing_map.go
stephan-rayner: That is a good call! I've added a comment explaining the structure 😄 

```
// GCP tiered rates are ordered ascending by start usage amount. The first tier
// typically has a start amount of 0 and a unit price of $0 (free tier). Paid
// pricing begins at the next tier. Taking the last rate gives the steady-state
// paid price, which is what this exporter reports.
```

docs/metrics/gcp/vertex.md
stephan-rayner: Thinking about the difference between ours and theirs: 

| Dimension | Sigil | Ours | Joinable? |
|---|---|---|---|
| Model | `gen_ai_request_model` | `model_id` | No (different label names, tricks required) |
| Provider | `gen_ai_provider_name` | `family` | No (different label names, tricks required) |
| Token direction | `gen_ai_token_type=input/output` | separate metrics | Design Difference (Sigil uses one metric with a label to distinguish input and output and we use two separate metrics) |

stephan-rayner: `price_tier` is only on the compute metric since that's the only place it's meaningful. Compute nodes have genuine on-demand and spot tiers backed by separate SKUs in the billing API. Token pricing has one price per model per region with no tiers in the data, so adding a constant `price_tier="on_demand"` label might just be noise. What do you think?

Also updated `model` to `model_id` and added the `family` label to align with Bedrock.

nikimanoledaki: > Token pricing has one price per model per region with no tiers in the data, so adding a constant price_tier="on_demand" label might just be noise. What do you think?

Makes sense to skip `price_tier` for tokens 👍 

I don't think the `family` label returns data - seeing `unknown` in all metrics:
```
cloudcost_gcp_vertex_token_input_usd_per_1k_tokens{family="unknown",model_id="adaptive-machine-translation",region="global"} 0.025
```

Does the GCP API provide any other dimension to fetch the model's provider/family?

pkg/google/vertex/pricing_map.go
```diff
@@ -0,0 +1,269 @@
...
+	"cloud.google.com/go/billing/apiv1/billingpb"
+	"github.com/grafana/cloudcost-exporter/pkg/google/client"
+)
+
+const (
+	vertexAIServiceName = "Vertex AI"
+	// discoveryEngineServiceName is the GCP Billing API service name for Discovery Engine,
+	// which hosts the Ranking API used for reranking.
+	// NOTE: Must be verified against the live GCP Billing API.
+	discoveryEngineServiceName = "Cloud Discovery Engine"
```
nikimanoledaki: I see this log when I start the `vertex` service:
```
./cloudcost-exporter -gcp.services=vertex -gcp.projects=<foo> -provider=gcp
...
WARN failed to get Discovery Engine service name, reranking metrics will be unavailable error="service not found"
...
```

there might be some misconfiguration 👀

docs/metrics/gcp/vertex.md
stephan-rayner: Sadly Vertex does not offer any other mechanism to get family we are inferring it from them from the `model _id`

pkg/google/vertex/pricing_map.go
stephan-rayner: It was this was the wrong name. This is since fixed

docs/metrics/gcp/vertex.md
stephan-rayner: This is done here https://github.com/grafana/cloudcost-exporter/pull/929/changes#diff-07199c64804155df795a1a3f9a3d2620a672c182d00cff5f4d6157f9e34e4cccR158

pkg/google/vertex/vertex.go
rooneyshuman: I'm not sure the consistency argument holds here - in GKE/GCS/CloudSQL those fields are actually read (projects is iterated in `Collect()`, regions backs `Regions()`). Keeping them in `Vertex` means `RegionsForProjects()` makes unnecessary live GCP API calls at startup whose results are immediately discarded.

rooneyshuman (commented):
added a comment on the inclusion of `regions` and `projects` in the vertext collector

pkg/google/vertex/vertex.go
stephan-rayner: You're right! The consistency argument doesn't really hold when the fields do different work in different collectors. I'm going to remove both fields and the `RegionsForProjects` call.

pkg/google/gcp.go
```diff
@@ -160,6 +161,17 @@ func New(ctx context.Context, config *Config) (*GCP, error) {
 					slog.String("message", err.Error()))
 				continue
 			}
+		case "VERTEX":
+			collector, err = vertex.New(ctx, &vertex.Config{
+				Projects:       config.Projects,
+				ScrapeInterval: config.ScrapeInterval,
```
rooneyshuman: are we using these config fields anywhere?

stephan-rayner: I thought so let me check

stephan-rayner: These were being used but they are no longer, good catch. I'll pull them now!

docs/metrics/gcp/vertex.md
```diff
@@ -0,0 +1,53 @@
+# Vertex AI Metrics
+
+Metrics exported for the GCP Vertex AI service.
+
+## Token Pricing
+
+| Metric | Labels | Description |
+|--------|--------|-------------|
+| `cloudcost_gcp_vertex_token_input_usd_per_1k_tokens` | `model_id`, `family`, `region` | Input cost in USD per 1k tokens, for models billed by token. Character-billed models use the character metric. |
```
rooneyshuman: the actual metric name emitted is `cloudcost_gcp_vertex_input_usd_per_1k_tokens`. This is a different name

docs/metrics/gcp/vertex.md
```diff
@@ -0,0 +1,53 @@
...
+| `cloudcost_gcp_vertex_token_output_usd_per_1k_tokens` | `model_id`, `family`, `region` | Output cost in USD per 1k tokens, for models billed by token. Character-billed models use the character metric. |
```
rooneyshuman: same as above: `cloudcost_gcp_vertex_output_usd_per_1k_tokens` is the actual metric emitted

pkg/utils/consts.go
```diff
@@ -31,6 +31,16 @@ const (
 	TokenInputCostSuffix = "token_input_usd_per_1k_tokens"
 	// TokenOutputCostSuffix is the suffix for per-1k-output-token cost metrics.
 	TokenOutputCostSuffix = "token_output_usd_per_1k_tokens"
+	// InputTokenCostSuffix is the suffix for per-1k-input-token cost metrics without the redundant "token" prefix.
+	InputTokenCostSuffix = "input_usd_per_1k_tokens"
+	// OutputTokenCostSuffix is the suffix for per-1k-output-token cost metrics without the redundant "token" prefix.
+	OutputTokenCostSuffix = "output_usd_per_1k_tokens"
```
rooneyshuman: these constants are slightly different names for seemingly the same thing? What's the suffix we actually want to keep?

`TokenInputCostSuffix` vs `InputTokenCostSuffix`
`TokenOutputCostSuffix` vs `OutputTokenCostSuffix`

pkg/google/vertex/pricing_map.go
```diff
@@ -0,0 +1,307 @@
...
+					if snap.tokens[region] == nil {
+						snap.tokens[region] = make(map[string]*TokenPricing)
+					}
+					if snap.tokens[region][model] == nil {
+						snap.tokens[region][model] = &TokenPricing{}
+					}
+					snap.tokens[region][model].OutputPer1kTokens = price
+				}
+			}
+			continue
```
rooneyshuman, line 249: there's substantial duplication between this logic branch and the `tokenInputRegex` one. Can we extract that into a small helper?

pkg/google/vertex/pricing_map.go
```diff
@@ -0,0 +1,307 @@
...
+	vertexAIServiceName = "Vertex AI"
+	// discoveryEngineServiceName is the GCP Billing API service name for Vertex AI Search,
+	// which hosts the Ranking API used for reranking.
+	discoveryEngineServiceName = "Vertex AI Search"
+)
+
+var (
+	// tokenInputRegex matches Vertex AI input token/character SKU descriptions.
+	// Examples: "Gemini 1.5 Flash Input tokens", "Gemini Embedding 001 Input characters"
+	// NOTE: Exact SKU description strings must be verified against the live GCP Billing API.
```
rooneyshuman: can we add tests for this?

docs/metrics/gcp/vertex.md
```diff
@@ -0,0 +1,53 @@
+# Vertex AI Metrics
+
+Metrics exported for the GCP Vertex AI service.
+
+## Token Pricing
```
rooneyshuman, line 5: can we add another section to document the Character Pricing metrics?

rooneyshuman (dismissed):

docs/metrics/gcp/vertex.md
stephan-rayner: woops I missed one when I made this change thank you will fix that!

docs/metrics/gcp/vertex.md
stephan-rayner: Ya ok!

docs/metrics/gcp/vertex.md
stephan-rayner: same as above

pkg/google/vertex/pricing_map.go
stephan-rayner: Ya that is workable

pkg/utils/consts.go
stephan-rayner: The first one is used by Bedrock the second is used by Vertex

I thought the `token_output_usd_per_1k_tokens` had a redundant token in it so I went with `input_usd_per_1k_tokens`

I created this issue to track it on Bedrocks side
https://github.com/grafana/cloudcost-exporter/issues/953

As the metrics are being used in a dashboard I thought it would make sense to not change it here without speaking with @nikimanoledaki and @simaarfania about what should change in other deployment_tools and a couple dashboards. I could also follow the same convention and this would be cleaner but the name would still seem a little redundant.

How do you feel about it?

stephan-rayner: This has been addressed by https://github.com/grafana/cloudcost-exporter/pull/954

pkg/google/vertex/vertex.go
```diff
@@ -0,0 +1,188 @@
...
+	pricingMap *PricingMap
+	logger     *slog.Logger
+}
+
+// New creates and initialises a Vertex AI Collector.
+// Pricing is fetched at construction time; an error means the collector cannot be used.
+func New(ctx context.Context, logger *slog.Logger, gcpClient client.Client) (*Collector, error) {
+	logger = logger.With("collector", subsystem)
+
+	pm, err := NewPricingMap(ctx, gcpClient)
```
rooneyshuman: can we pass through the collector's logger when instantiating the pricing map? that way we can more easily scope the log lines

pkg/google/vertex/pricing_map.go
```diff
@@ -0,0 +1,280 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to get Vertex AI service name: %w", err)
+	}
+	skus := pm.gcpClient.GetPricing(ctx, serviceName)
+	if len(skus) == 0 {
+		return fmt.Errorf("no SKUs found for Vertex AI service")
+	}
+
+	if deSvcName, err := pm.gcpClient.GetServiceName(ctx, discoveryEngineServiceName); err != nil {
+		slog.Warn("failed to get Discovery Engine service name, reranking metrics will be unavailable", "error", err)
```
rooneyshuman: let's pass in the collector's logger and use it when creating log lines

pkg/google/vertex/vertex.go
```diff
@@ -0,0 +1,196 @@
...
+				ch <- prometheus.MustNewConstMetric(vertexCharacterOutputDesc, prometheus.GaugeValue,
+					price, model, familyFromModelID(model), region, tier)
+			}
+		}
+	}
+	for region, machines := range snapshot.compute {
+		for machineType, useCases := range machines {
+			for useCase, pricing := range useCases {
+				ch <- prometheus.MustNewConstMetric(vertexComputeCostDesc, prometheus.GaugeValue,
+					pricing.OnDemandPerHour, machineType, useCase, region, "on_demand")
```
simaarfania: Should `on_demand` be guarded the same way `spot` is on the next line? `ParseSkus` (`pricing_map.go:263`) creates the `ComputePricing` entry on the first matching SKU of either tier, so if a region has only a `Spot Custom ...` SKU, `OnDemandPerHour` stays `0` and we emit `{price_tier="on_demand"} 0`. Would wrapping this in `if pricing.OnDemandPerHour > 0` keep compute consistent with them?

pkg/google/vertex/vertex.go
```diff
@@ -0,0 +1,196 @@
...
+
+// Collector collects Vertex AI pricing metrics.
+type Collector struct {
+	pricingMap *PricingMap
+	logger     *slog.Logger
+}
+
+// New creates and initialises a Vertex AI Collector.
+// Pricing is fetched at construction time; an error means the collector cannot be used.
+func New(ctx context.Context, logger *slog.Logger, gcpClient client.Client) (*Collector, error) {
```
simaarfania, line 63: I recall the other collectors taking in the `Config` struct, is there a reason we aren't taking it in here?

pkg/google/vertex/vertex_test.go
```diff
@@ -0,0 +1,325 @@
+package vertex
```
simaarfania, line 1: Could this use the generated `client.Mock` in `pkg/google/client/client_mock.go` instead of hand-implementing it?

pkg/google/vertex/vertex.go
stephan-rayner: I agree that would be better for consistency, I have added that in commit 699a316

pkg/google/vertex/vertex.go
stephan-rayner: I think you are on the nose here. I have added a guard in 4ea49fe

pkg/google/vertex/vertex_test.go
stephan-rayner: Ya, you are right this could really create a drift problem but the client.Mock wraps real billing clients rather than using testify expectations, and porting to it would require a gRPC test server while losing the per-service error injection that the current tests rely on. To this end I think we should probably keep it. There is a helper but the single-service helpers in billing_test_helpers.go can't cover Vertex's dual-service test cases I believe.

stephan-rayner: How does that sit with you?

stephan-rayner: Keeping the hand-implemented mock: addressed in the thread above.

pkg/google/vertex/vertex.go
```diff
@@ -0,0 +1,208 @@
...
+// Pricing is fetched at construction time; an error means the collector cannot be used.
+func New(ctx context.Context, config *Config, logger *slog.Logger, gcpClient client.Client) (*Collector, error) {
+	logger = logger.With("collector", subsystem)
+
+	pm, err := NewPricingMap(ctx, logger, gcpClient)
+	if err != nil {
+		return nil, fmt.Errorf("failed to initialize pricing map: %w", err)
+	}
+
+	refreshInterval := config.ScrapeInterval
```
simaarfania: The refresh goroutine tickers on `config.ScrapeInterval`, falling back to `PriceRefreshInterval` (24h) only when it's zero. But I don't think it's ever 0... `-scrape-interval` defaults to `1h`. So I think we refetch the full Vertex AI + Discovery Engine SKU catalog every hour, not "every 24 hours" as the PR states.

The `bedrock` collector tickers on the fixed 24h `pricingstore.PriceRefreshInterval` and doesn't take a `ScrapeInterval` at all. Was the `ScrapeInterval` wiring intentional? If not, I'd ticker on `PriceRefreshInterval` to match the docs and the other collectors:

```go
ticker := time.NewTicker(PriceRefreshInterval)
```

stephan-rayner: Good catch. Addressed in 9b74ccf: the goroutine now tickers on `PriceRefreshInterval` directly, matching GKE.

pkg/google/vertex/pricing_map.go
```diff
@@ -0,0 +1,381 @@
...
+			case pat.direction == "input" && pat.billingType == "token":
+				target = snap.tokenInput
+			case pat.direction == "output" && pat.billingType == "token":
+				target = snap.tokenOutput
+			case pat.direction == "input" && pat.billingType == "char":
+				target = snap.charInput
+			default:
+				target = snap.charOutput
+			}
+			applyPrice(target, model, pat.tier, price, regions)
```
simaarfania, line 244: It seems that the token matcher collapses a model's modalities onto a single price.

`modalityGroup` matches `text|audio|image|video`, and the patterns capture only the model name (`matches[1]`) plus a hard-coded `tier`, so the modality is matched but discarded. When `applyPrice(target, model, pat.tier, ...)` writes into `target[region][model][tier]`, every modality of a model lands on the same key and the last SKU processed wins.

GCP prices the modalities very differently, so the surviving value is effectively random. From a live `GetPricing` dump of the `Vertex AI` service, from a random model (Gemini Flash 1.5), I see:

- Image Input   $2e-05
- Video Input   $2e-05
- Audio Input   $2e-06
- Text Input    $1.8e-08

All four match `^(.+?) <modality> Input - Predictions$` and resolve to `(model=gemini-1.5-flash, price_tier=on_demand)`, so `cloudcost_gcp_vertex_input_usd_per_1k_tokens{model_id="gemini-1.5-flash", price_tier="on_demand"}` ends up as whichever SKU was parsed last

To fix it, I think we should keep the modality distinct by adding a label or something. Or actually, If only  scoping `modalityGroup` to `text` resolves it with the least code, with a short comment that the other modalities are intentionally dropped.

stephan-rayner: I am leaning toward the second option. At present no end user (mostly us) makes use of the audio, image, or video endpoints in vertex

simaarfania: Good with me!

simaarfania (commented):
I think there are a couple merge conflicts we need to resolve. Overall looks good!

go.mod
```diff
@@ -22,6 +22,7 @@ require (
 	github.com/aws/aws-sdk-go-v2/service/sts v1.42.3
 	github.com/google/go-cmp v0.7.0
 	github.com/googleapis/gax-go/v2 v2.22.0
+	github.com/grafana/grafana-foundation-sdk/go v0.0.16
```
simaarfania: I don't think we use this anywhere, running `go mod tidy` should drop this!

pkg/google/gcp.go
```diff
@@ -209,6 +210,16 @@ func New(ctx context.Context, config *Config) (*GCP, error) {
 					slog.String("message", err.Error()))
 				continue
 			}
+		case "VERTEX":
```
simaarfania: The switch hardcodes to `case "VERTEX" whereas the other services use a named constant. Can we match that pattern?

pkg/google/gcp.go
```diff
@@ -18,6 +18,7 @@ import (
 	cloudcost_exporter "github.com/grafana/cloudcost-exporter"
```
simaarfania, line 18: Also, it seems that Vertex is missing from Services(), so it won't appear in `-list-services`. Should we add an entry?

docs/metrics/gcp/vertex.md
```diff
@@ -0,0 +1,110 @@
+# Vertex AI Metrics
```
simaarfania, line 1: We should liknk to this from `docs/metrics/README.md` or `docs/README.md` so it is more easily discoverable.

pkg/google/gcp.go
stephan-rayner: Oh that is funny I'll fix that

pkg/google/gcp.go
stephan-rayner: Ya I like that better good catch. I'll go with a name constant

go.mod
stephan-rayner: Ya I don't know how that snuck in I think it was a merge I'll pull that

stephan-rayner: I have dealt with this

docs/metrics/gcp/vertex.md
stephan-rayner: Ya that is a good call, I'm going to do both

pkg/google/gcp.go
stephan-rayner: This is sorted now

pkg/google/vertex/vertex.go
```diff
@@ -0,0 +1,203 @@
...
+)
+
+const (
+	subsystem            = "gcp_vertex"
+	PriceRefreshInterval = 24 * time.Hour
+)
+
+// Config holds configuration for the Vertex AI collector.
+type Config struct {
+	ScrapeInterval time.Duration
```
leonorfmartins: I don't think this is never consumed down the line?

pkg/google/vertex/vertex_test.go
```diff
@@ -0,0 +1,342 @@
...
+	assert.Equal(t, "google", rerank.Labels["family"]) // "semantic" prefix maps to google
+	assert.Equal(t, "global", rerank.Labels["region"])
+	assert.InDelta(t, 0.001, rerank.Value, 1e-9)
+}
+
+func TestCollect_DiscoveryEngineUnavailable_RerankingOmitted(t *testing.T) {
+	c, err := New(t.Context(), &Config{}, testLogger(),
+		&stubVertexClient{
+			serviceName:      "services/vertex-ai",
+			deServiceNameErr: fmt.Errorf("discovery engine billing API unavailable"),
```
leonorfmartins: this log message could be more specific, since in this case we know the discovery engine is vertex-ai :)

pkg/google/vertex/vertex.go
stephan-rayner: You are correct that isn't used

pkg/google/vertex/vertex_test.go
stephan-rayner: Ya that is fair, that is a good call. I'll add something in there to make it more about vertex

pkg/google/vertex/vertex.go
```diff
@@ -0,0 +1,198 @@
...
+	return nil
+}
+
+// Name implements provider.Collector.
+func (c *Collector) Name() string {
+	return subsystem
+}
+
+// Collect emits Vertex AI pricing metrics.
+func (c *Collector) Collect(ctx context.Context, ch chan<- prometheus.Metric) error {
```
simaarfania, line 114: Collect emits from the cached pricing map and only checks `ctx.Err()` after iterating every region/model/tier. `vpc` is the closest analog (also emits from a cached pricingMap) it seems, and it checks `ctx.Done()` between iterations, returning ctx.Err() mid-loop ([link](https://github.com/grafana/cloudcost-exporter/blob/main/pkg/google/vpc/vpc.go#L172)). Matching that here would honor a cancelled or timed-out scrape promptly instead of running the full loop first.

stephan-rayner: That is fair, That is straight forward enough, I can make that fix for sure

pkg/google/vertex/pricing_map.go
```diff
@@ -0,0 +1,385 @@
...
+	// MaaS token format — specific before generic
+	{regexp.MustCompile(`(?i)^(.+?)\s+Batch\s+Input\s+Tokens?$`), "input", "token", "batch"},
+	{regexp.MustCompile(`(?i)^(.+?)\s+Batch\s+Output\s+Tokens?$`), "output", "token", "batch"},
+	{regexp.MustCompile(`(?i)^(.+?)\s+Cached(?:\s+Text)?\s+Input\s+Tokens?$`), "input", "token", "cached"},
+	{regexp.MustCompile(`(?i)^(.+?)\s+Input\s+Tokens?$`), "input", "token", "on_demand"},
+	{regexp.MustCompile(`(?i)^(.+?)\s+Output\s+Tokens?$`), "output", "token", "on_demand"},
+
+	// Character-billed models
+	{regexp.MustCompile(`(?i)^(.+?)\s+Input\s+Characters?$`), "input", "char", "on_demand"},
+	{regexp.MustCompile(`(?i)^(.+?)\s+Output\s+Characters?$`), "output", "char", "on_demand"},
```
simaarfania, line 135: `skuPatterns` has only `on_demand` for character SKUs, but `docs/metrics/gcp/vertex.md` lists `batch` as a valid `price_tier` for character metrics. If GCP has a SKU like `"Translation LLM Input Characters Batch"`, it would fall through all three regex paths and be silently dropped.

should we add batch character patterns here, or remove `batch` from the docs table if GCP doesn't offer it today?

docs/metrics/gcp/vertex.md
```diff
@@ -0,0 +1,110 @@
...
+#### `price_tier` Values for Token Metrics
+
+Tiers are composed from up to three modifiers: a `thinking_` prefix, a `cached_` prefix, and a `_long_context` suffix. Not all combinations exist in GCP's SKU catalogue; only tiers with a matching SKU are emitted.
+
+**Simple tiers**
+
+| Value | Description |
+|-------|-------------|
+| `on_demand` | Standard real-time inference |
+| `batch` | Batch prediction |
```
simaarfania, line 30: The table lists `batch` as a valid `price_tier` for character metrics, but the GCP pricing page shows `Translation LLM` is flat-rate only with no batch tier. should we drop batch to avoid this confusion?

simaarfania (dismissed):

docs/metrics/gcp/vertex.md
stephan-rayner: Let me double check that

simaarfania (approved):

PR #969, opened 2026-05-14
chore(ci): use go.mod as single source of truth for Go version

rooneyshuman:
## Summary

Make `go.mod` the single source of truth for Go version, replacing a custom composite action with the standard `actions/setup-go` mechanism.

**Why:** Vulnerability scanners read the installed Go version from `go.mod`'s `go` directive, not from binary metadata. With `go 1.25.0` declared, released artifacts were flagged for 1.25.x-era CVEs regardless of the toolchain actually used to compile them.

**Changes:**
- Bump `go` directive in `go.mod`: `1.25.0` → `1.26.3`
- Add `toolchain go1.26.2` directive so `setup-go` installs the same Go release as the Dockerfile
- Replace `./.github/actions/setup-goversion` with `actions/setup-go@v6.4.0` + `go-version-file: go.mod` across `tests.yml`, `release-on-pr-merge.yml`, `release-on-tag-push.yml`
- Delete the now-unused `setup-goversion` composite action
- Keep `cache: false` on release workflows to avoid the zizmor cache-poisoning finding (release jobs publish artifacts; tests workflow keeps caching since it doesn't)

**Going forward:** Renovate's gomod manager treats the `toolchain` directive as a pinned dependency and will open PRs to bump it alongside the Dockerfile `golang` tag, eliminating the prior drift between the two sources.

## Test plan

- [x] `go build ./...` passes
- [x] `go vet ./...` passes
- [x] `go test ./...` — all packages `ok`

Tracking: https://github.com/grafana/deployment_tools/issues/588539
Closes https://github.com/grafana/deployment_tools/issues/588512


Duologic:
Have you considered using `go-version-file`? https://github.com/actions/setup-go/blob/main/docs/advanced-usage.md#using-the-go-version-file-input

Then the action takes the verson from go.mod

The Dockerfile may still need to be updated as well to align these.

go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/grafana/cloudcost-exporter
 
-go 1.25.0
+go 1.26.0
```
Duologic: Any reason not to go for latest?
```suggestion
go 1.26.3
```

Duologic (approved):

PR #974, opened 2026-05-20
fix(release): use cloudcost-exporter-updater app for token broker

simaarfania:
Updates the `release-on-pr-merge` workflow to use `cloudcost-exporter-updater` as the GitHub App slug instead of the placeholder `github-app`.

This aligns with the corresponding change in `grafana/deployment_tools` which registers the Vault token broker config for the `cloudcost-exporter-updater` app.

The two PRs must merge together for the token generation to work end-to-end.


cindy (approved):

eloymg (approved):
LGTM! Thanks

PR #999, opened 2026-06-03
feat(gke): collect node and disk metadata in background stores

### LINKED ISSUE #272: GCP GKE: Collect node metrics in the background
# Overview 

We're hitting a scaling point within Grafana Labs where the time to collect metrics for GKE instances is >40s. The current configuration is setup to scrape metrics every 60 seconds. This aligns with our TCO recording rules that calculate our TCO every minute. We'd like to avoid bumping the scrape duration above 60 seconds so that our TCO recording rules are as fresh as possible.

Here's a 90 day view of our scrape duration increased over that time: 

image

## Goals

- [ ] collect disk metadata in a background process and store results in a cache
- [ ] collect pricing metadata in a background process and store results in a cache
- [ ] Scraping `/metrics` should be as fast as reading from the cache

@logyball already provided a nice framework to consider this work in his implementation of AKS metrics.

stephan-rayner:
Closes #272
Internal discussion: https://github.com/grafana/deployment_tools/issues/608719

## Problem

GKE scrape duration has been climbing for 90 days and is now regularly exceeding 40s against a 60s scrape interval. Every scrape was making live GCP API calls: one \`Zones.List\` plus two list calls per zone (\`Instances.List\` and \`Disks.List\`) for each configured project. At our current scale, that is over 260 API calls per scrape per project, all blocking on the Prometheus scrape path.

## What Changed

Node and disk metadata now collect in background goroutines and cache results in memory. \`Collect()\` reads from those caches, making scrape time effectively independent of GCP API latency.

Two new stores handle this:

- **NodeStore**: refreshes instance inventory every 5 minutes
- **DiskStore**: refreshes disk inventory every 15 minutes, with deduplication

Each store signals readiness via a channel. If a store has not completed its first populate when \`Collect()\` runs, that metric type is skipped with a log line rather than blocking or erroring.

This follows the same pattern as the AKS collector.

## Tradeoffs

- Node data is up to 5 minutes stale; disk data up to 15 minutes
- No GKE metrics are emitted on the first scrape after startup until both stores complete initial population
- Background populate failures surface as missing metrics, not scrape errors: monitor via \`cloudcost_exporter_gcp_gke_populate_errors_total\` or logs

## Testing

- Existing \`TestCollector_Collect\` updated to wait for store readiness before asserting metrics
- \`TestCollector_ZoneConcurrencyLimit\` replaced by \`TestNodeStore_Populate_ConcurrencyLimit\`, which tests concurrency where it now lives


stephan-rayner:
**Partial Failure Handling: Options for Discussion**

The current `Populate` implementation has an asymmetry worth resolving before this ships. Wanted to write up the options and get eyes on it before picking one.

**The Problem**

There are three distinct failure scenarios:

- `GetZones` fails: project is skipped, never enters `updates`. `maps.Copy` leaves the existing cache entry untouched. Stale data is preserved.
- All zone calls fail: every goroutine logs and returns `nil`, leaving `instances` as `nil`. `updates[project] = nil` is written, and `maps.Copy` overwrites the cache with nil. Stale data is wiped.
- Some zone calls fail: the store is updated with data from the zones that succeeded.

The last two cases can produce misleading metrics. A transient GCP outage wipes your node data entirely; a partial outage makes it look like your costs dropped.

**Option A: Preserve stale on total failure only (`successCount > 0`)**

Only skip the cache update if zero zone calls succeeded. A partial failure still updates with whatever came back.

Pros: fixes the "total failure wipes cache" bug with minimal change.
Cons: partial failure still emits an incomplete picture. For a cost exporter, a sudden apparent drop in 50% of your nodes looks like real cost reduction.

**Option B: Preserve stale on any failure (`failCount == 0`)**

Only update the cache when the populate was entirely clean. Any zone error leaves the previous data in place.

Pros: cache always represents a complete, consistent snapshot. Stale but complete is more trustworthy than fresh but partial.
Cons: a single persistently broken zone stalls all updates for that project indefinitely. You'd need to watch the error logs to know it's happening, since metrics would silently reflect the last clean state.

**Option C: Always replace (current partial-failure behaviour, fix the total-failure case)**

Same as Option A but make the behaviour explicit and documented rather than accidental.

**My lean**

Option B for a cost exporter. The audience for these metrics is making spend decisions, so a consistent but stale view is less harmful than a fresh but incomplete one. The persistent-zone-failure risk is real but addressable: it's loud in logs and a populate-failure metric counter (future work already noted in the PR) would make it observable. Open to pushback on this.

pkg/google/gke/gke.go
```diff
@@ -73,163 +62,70 @@ func (c *Collector) Register(_ provider.Registry) error {
...
-			close(disks)
-		}()
+	now := time.Now()
+	nodeCount, diskCount := 0, 0
 
-		for group := range instances {
-			if instances == nil {
-				continue
-			}
-			for _, instance := range group {
```
logyball, line 148: [NIT] - We can make the node and disk retrieval concurrent using a waitgroup as these two things shouldn't be dependent on one another.

We will get some modest performance gains I imagine from not blocking the node metric scraping on the disk metric scraping, no reason to not fetch them in parallel IMO.

pkg/google/gke/disk_store.go
```diff
@@ -0,0 +1,106 @@
...
+				if err != nil {
+					ds.logger.LogAttrs(egCtx, slog.LevelError, "failed to list disks in zone",
+						slog.String("project", project),
+						slog.String("zone", zone.Name),
+						slog.String("error", err.Error()))
+					return nil
+				}
+				mu.Lock()
+				for _, raw := range results {
+					key := raw.Zone + "/" + raw.Name
```
logyball: I like what you're doing here by making sure to de-dupe, but it does not look like it follows the [previously established method of doing it:](https://github.com/grafana/cloudcost-exporter/blob/4e3acabc1a12a34da5b8c45ffa236762a68c5f82/pkg/google/gke/disk.go#L84-L96)

using `d.Name()`

```go
// Name will return the name of the disk. If the disk has a label "kubernetes.io/created-for/pv/name" it will return the value stored in that key.
// otherwise it will return the disk name that is directly associated with the disk.
func (d Disk) Name() string {
	if d.description == nil {
		return d.name
	}
	// first check that the key exists in the map, if it does return the value
	name := coalesce(d.description, pvNameKey, pvNameShortKey)
	if name != "" {
		return name
	}
	return d.name
}
```

so I would suggest something like:



```suggestion
					d := NewDisk(raw, project)
					if seen[d.Name()] {
						continue
					}
					seen[d.Name()] = true
					disks = append(disks, d)
```

it does require an up-front init of disk that you might end up throwing away, but doesn't introduce new logic for how we are determining if a disk is a duplicate or not.

pkg/google/gke/gke_test.go
```diff
@@ -567,24 +590,60 @@ func TestCollector_ZoneConcurrencyLimit(t *testing.T) {
...
+
+func TestDiskStore_Populate_Deduplication(t *testing.T) {
+	fakeClient := &diskDedupClient{
+		zoneDisks: map[string][]*computev1.Disk{
+			"us-central1-a": {
+				{Name: "shared-name", Zone: "us-central1-a"},
+				{Name: "shared-name", Zone: "us-central1-a"}, // within-zone duplicate: should be dropped
+			},
+			"us-central1-b": {
+				{Name: "shared-name", Zone: "us-central1-b"}, // same name, different zone: should be kept
```
logyball: [see this comment](https://github.com/grafana/cloudcost-exporter/pull/999/changes#r3364025838)

This test is validating new behavior that exists as of this PR.  Previously the deduplication logic was already tested [up here](https://github.com/grafana/cloudcost-exporter/pull/999/changes#diff-0b43eaf66c6e6dd7174a93fb2c7c918a92275f6bbe92bc28c6ed980e85c152edL441)

Now it's possible that zone + name is a better way to do it, but I think we should validate that and test it separately out of this PR, which should be scoped to performance

pkg/google/gke/disk_store.go
```diff
@@ -0,0 +1,106 @@
...
+	logger    *slog.Logger
+	gcpClient client.Client
+	projects  []string
+
+	mu    sync.RWMutex
+	disks map[string][]*Disk
+
+	initialPopulationOnce sync.Once
+	initialPopulation     chan struct{}
+}
```
logyball, line 32: I feel like "Store()" here could be an interface that is implemented by DiskStore and NodeStore

- Done() and Populate() have identical type signatures
- New() has the same params 
- `Get[Nodes|Disks]()` could probably be genericized with a bit of finangling 

That would make a bit of the maintainance easier

pkg/google/gke/gke_test.go
```diff
@@ -555,9 +552,35 @@ func (c *concurrentGCPClient) ListDisks(_ context.Context, _ string, _ string) (
 	return nil, nil
 }
 
-func TestCollector_ZoneConcurrencyLimit(t *testing.T) {
-	// 8 zones → 16 goroutines (2 per zone: ListInstancesInZone + ListDisks),
-	// but zoneCollectConcurrencyLimit=10 caps the total.
+func TestNodeStore_Populate_ConcurrencyLimit(t *testing.T) {
+	// 8 zones → up to 8 goroutines for ListInstancesInZone during NodeStore.Populate.
+	// nodePopulateConcurrencyLimit caps the total.
+	const numZones = 8
```
logyball: this should be >=10 to test the actual concurrency limit

- [nodes](https://github.com/grafana/cloudcost-exporter/pull/999/changes#diff-ce546d364de8a4d4a87d09e8f6f026bd819b91e5dedf3e71f787a76db444a55aR17)
- [disks](https://github.com/grafana/cloudcost-exporter/pull/999/changes#diff-92ae644a034154c392fbb4338df201e8b8c2d5566209d1e441ccb435be040b6aR17)

pkg/google/gke/node_store.go
```diff
@@ -0,0 +1,98 @@
...
+						slog.String("error", err.Error()))
+					return nil
+				}
+				mu.Lock()
+				instances = append(instances, results...)
+				mu.Unlock()
+				return nil
+			})
+		}
+		_ = eg.Wait()
```
logyball: no real point in using an errgroup if we don't ever return an error - I'd say that based on our decision about what option to go with, we may want to use this (or not) 🤷

logyball (dismissed):
> My lean
>
> Option B for a cost exporter. The audience for these metrics is making spend decisions, so a consistent but stale view is less harmful than a fresh but incomplete one. The persistent-zone-failure risk is real but addressable: it's loud in logs and a populate-failure metric counter (future work already noted in the PR) would make it observable. Open to pushback on this.

I tend to lean towards option A.  I don't think that we need to block the population of fresh metrics on the failure of a single zone in a project.

Consider the psuedocode, simplified from the PR:

```
for p := range projects {
  zs, err := p.getZones()
  if err != nil {
     // can't get _any_ zones, skip this project
     // log failure, increment project level(?) failure metric
     continue
  }
  for z in zs:
    disks, err := z.listDisks()
    if err {
      // err listing particular zone
      // skip for this zone
     // log failure, increment project+zone level(?) failure metric
      continue
    }
    cache.update(disks)
}
```

This seems like the most reasonable way towards the eventually consistent way of doing things.  We don't want to couple project-level success (think `grafanalabs-dev`) to the success of all sub-zones (hundreds 😿 ):

Image

Basically:

1. we get all the relevant projects from the command line
2. we get the relevant zones from the GCP client, per-project
  - if this fails, we bail, there's nothing to do here, cache remains untouched
3. we get the relevant disk for each zone in the project
  - if we cant get the disks, we move on
  - if we CAN get the disks, we update the cache

So at scrape time, all data is as up-to-date per-project-zone as it can be.  

> Cons: partial failure still emits an incomplete picture. For a cost exporter, a sudden apparent drop in 50% of your nodes looks like real cost reduction.

I'm not sure how this is intended.  I feel like not updating the cache is not wiping it to "drop" metrics, it's simply returning "stale" data for those areas that can't be updated.

for disks: Rather than have the cache be one big list of disks per project, why not have it split by project+zone?

For nodes: rather than project -> list of nodes, why not project+zone -> list of nodes?

pkg/google/gke/gke_test.go
stephan-rayner: Agreed!

stephan-rayner:
> I tend to lean towards option A. I don't think that we need to block the population of fresh metrics on the failure of a single zone in a project.
//src: https://github.com/grafana/cloudcost-exporter/pull/999#pullrequestreview-4437370638

I see your logic here, and I dig it @logyball thank you!

pkg/google/gke/disk_store.go
```diff
@@ -0,0 +1,129 @@
...
+				}
+				successfulZones++
+				mu.Unlock()
+			}()
+		}
+		wg.Wait()
+
+		// ctx.Err() != nil means we bailed for shutdown, not a real total-failure.
+		if ctx.Err() == nil && successfulZones == 0 && len(zones) > 0 {
+			ds.logger.LogAttrs(ctx, slog.LevelError, "all zone listings failed, wiping cached data",
```
leonorfmartins: We can choose to keep stale cache instead of wiping it. Regardless of what we choose, we should log an error so we can act on it.

stephan-rayner: Both points are addressed in e5452a3. The cache is now keyed by project → zone, so a failed zone call returns early without touching the cache entry for that zone — stale data is preserved. The error is logged at `slog.LevelError` on the same path.

pkg/google/gke/gke.go
stephan-rayner: This has been addressed

docs/metrics/gcp/gke.md
```diff
@@ -3,9 +3,27 @@
...
+| cloudcost_gcp_gke_instance_memory_usd_per_gib_hour         | Gauge       | The memory cost of a GCP Compute Instance, associated to a GKE cluster, in USD/(GiB*h)      | `cluster_name`=&lt;name of the cluster the instance is associated with&gt; <br/> `instance`=&lt;name of the compute instance&gt; <br/> `region`=&lt;GCP region code&gt; <br/> `family`=&lt;broader compute family (n1, n2, c3 ...) &gt; <br/> `machine_type`=&lt;specific machine type, e.g.: n2-standard-2&gt; <br/> `project`=&lt;GCP project, where the instance is provisioned&gt; <br/> `price_tier`=&lt;spot\|ondemand&gt;                                            |
```
logyball, line 7: nice catch

pkg/google/gke/node_store.go
```diff
@@ -0,0 +1,122 @@
...
+			}
+			wg.Add(1)
+			go func() {
+				defer wg.Done()
+				defer func() { <-sem }()
+				// Re-check: sem<- may have raced with cancellation.
+				if ctx.Err() != nil {
+					return
+				}
+				results, err := ns.gcpClient.ListInstancesInZone(project, zone.Name)
```
logyball: NIT - and not introduced by your PR, and not needing to be fixed by your PR.  but this should probably accept the `ctx` so it can be canceled if it timesout in flight:

```go
func (c *Compute) listInstancesInZone(ctx context.Context, projectId, zone string) ([]*MachineSpec, error) {
	var allInstances []*MachineSpec

	err := c.computeService.Instances.List(projectId, zone).Pages(ctx, func(page *compute.InstanceList) error {
		if page == nil {
			return nil
		}
		for _, instance := range page.Items {
			allInstances = append(allInstances, NewMachineSpec(instance))
		}
		return nil
	})
	return allInstances, err
}
```

but that's neither here nor there

pkg/google/gke/disk_store.go
```diff
@@ -0,0 +1,134 @@
...
+			ds.logger.LogAttrs(ctx, slog.LevelError, "failed to get zones",
+				slog.String("project", project),
+				slog.String("error", err.Error()))
+			continue
+		}
+
+		sem := make(chan struct{}, ds.concurrency)
+		var wg sync.WaitGroup
+
+	zoneLoop:
```
logyball: I've peeled off a branch to make some changes to this and I'll raise a PR against this PR - but I don't think we need to have a labeled line here that we jump back to.  Effectively, we're recreating an `errGroup` here  to bound concurrent requests without actually returning an error (correct behavior, just a little overcomplicated IMO), so we could do that with an errGroup itself, and just not return an error.

pkg/google/gke/disk_store.go
```diff
@@ -0,0 +1,134 @@
...
+		close(ds.initialPopulation)
+	})
+
+	for _, project := range ds.projects {
+		zones, err := ds.gcpClient.GetZones(project)
+		if err != nil {
+			ds.logger.LogAttrs(ctx, slog.LevelError, "failed to get zones",
+				slog.String("project", project),
+				slog.String("error", err.Error()))
+			continue
```
logyball: One quesstion that we should think about addressing here (maybe related to standardizing error collection) is to increment an operational error metric here.  

This dev doesnt represent a degrading of behavior (see [here](https://github.com/grafana/cloudcost-exporter/blob/169cdf2498b1687f9161cf0f5a03b6dea0b6cdca/pkg/google/gke/gke.go#L127) - the error behavior is largely unchanged).  But it will be challenging to see stale data accumulating unless we're alerting on logs, which is more tricky than alerting on metrics.

pkg/google/gke/disk_store.go
```diff
@@ -0,0 +1,134 @@
...
+	// the last-writer race on ds.disks.
+	if !ds.populating.CompareAndSwap(false, true) {
+		ds.logger.LogAttrs(ctx, slog.LevelInfo, "populate already in progress, skipping tick")
+		return
+	}
+	defer ds.populating.Store(false)
+
+	defer ds.initialPopulationOnce.Do(func() {
+		close(ds.initialPopulation)
+	})
```
logyball, line 88: this is a clever way of keeping things from stepping on one another without mucking about with too many locks/mutexes, nice

logyball (dismissed):
please check out this PR: https://github.com/grafana/cloudcost-exporter/pull/1042

It encompasses some of the ideas that I saw from this PR and addresses some of the comments I had.

simaarfania (commented):
I think we need to update [cmd/exporter/exporter.go](https://github.com/grafana/cloudcost-exporter/blob/main/cmd/exporter/exporter.go). The flag help for `gcp.gke.zone-concurrency`  still says collection happens "during a GKE scrape" and "divided by 2" for two goroutines per zone. After this PR, collection is background and the node and disk stores each apply the cap independently, would you mind updating that ?

simaarfania (approved):
LGTM!

rooneyshuman (approved):

PR #1008, opened 2026-06-05
fix(collectors): propagate Collect() errors in GKE, EC2, and RDS

stephan-rayner:
GKE zone-level goroutines for ListInstancesInZone and ListDisks returned nil on failure, preventing the collectormetrics wrapper from incrementing the error counter. They now return errors to the errgroup, which Collect() surfaces after emitting partial metrics for successful zones. Also fixes a latent bug where the nil-group guard checked the channel instead of the group value.

EC2 fetchInstancesData and fetchVolumesData silently dropped API errors. Both now return errors, collected via a buffered channel and joined before Collect() returns.

RDS collected regional listing failures into a slice, logged them, then returned nil. Collect() now returns a joined error after emitting metrics for any regions that succeeded


leonorfmartins (approved):

stephan-rayner:
Tracking the staged rollout work in #1080 before this gets merged.

PR #934, opened 2026-04-29
feat(aws/bedrock): add marketplace pricing for Claude 4.x and Cohere models 

simaarfania:
Part of https://github.com/grafana/cloudcost-exporter/issues/933.

Adds a second pricing source to the Bedrock collector and reconciles it with the existing one. The collector queried only `AmazonBedrock` in the AWS Pricing API, which covers Claude 2/3 and Amazon-native models. The models most used in production (Claude 4.x, Cohere Rerank/Embed) are priced under a separate service code, `AmazonBedrockFoundationModels`, and were missing entirely.

## Metrics emitted

Three gauges, all list-price rates (not measured spend):

| Metric | Unit |
|---|---|
| `cloudcost_aws_bedrock_input_usd_per_1k_tokens` | USD per 1,000 input tokens |
| `cloudcost_aws_bedrock_output_usd_per_1k_tokens` | USD per 1,000 output tokens |
| `cloudcost_aws_bedrock_search_unit_usd_per_1k_search_units` | USD per 1,000 search units (e.g. Cohere Rerank) |

Labels on every series: `account_id`, `region`, `model_id`, `family`, `price_tier`.

```
cloudcost_aws_bedrock_input_usd_per_1k_tokens{account_id="123456789012",region="us-east-1",model_id="claude-sonnet-4.6",family="anthropic",price_tier="on_demand"} 0.0033
cloudcost_aws_bedrock_output_usd_per_1k_tokens{account_id="123456789012",region="us-east-1",model_id="claude-sonnet-4.6",family="anthropic",price_tier="on_demand"} 0.0165
cloudcost_aws_bedrock_input_usd_per_1k_tokens{account_id="123456789012",region="us-east-1",model_id="claude-sonnet-4.6",family="anthropic",price_tier="cache_read"} 0.00033
cloudcost_aws_bedrock_input_usd_per_1k_tokens{account_id="123456789012",region="us-east-1",model_id="claude-sonnet-4.6",family="anthropic",price_tier="cross_region_batch"} 0.0015
cloudcost_aws_bedrock_search_unit_usd_per_1k_search_units{account_id="123456789012",region="us-east-1",model_id="cohere-rerank-v3.5",family="cohere",price_tier="on_demand"} 2
```

## Marketplace pricing source

- `pkg/aws/client/`: add `ListBedrockMarketplacePrices` and implement it against `AmazonBedrockFoundationModels`.
[...]

PR #1010, opened 2026-06-05
docs(aws/bedrock): document marketplace pricing source, families, and tiers

simaarfania:
Updates the Bedrock metrics docs to match the `AmazonBedrockFoundationModels`
(Bedrock Marketplace) pricing source.

## Summary

- Documents the two pricing sources: `AmazonBedrock` (Claude 2/3, Nova, Titan)
  and `AmazonBedrockFoundationModels` (Claude 4.x, Cohere Rerank/Embed, Meta,
  AI21, Writer, TwelveLabs), and how they merge.
- Expands the `family` label set to `anthropic`, `amazon`, `cohere`, `meta`,
  `ai21`, `stability`, `writer`, `twelvelabs`, `unknown`.
- Adds the cross-region compound `price_tier` values (`cross_region_batch`,
  `cross_region_flex`, `cross_region_priority`).
- Documents the two `model_id` conventions (standard `usagetype` slug vs
  marketplace lowercase-hyphen `servicename`) and clarifies that `model_id` is a
  normalized pricing identifier, not a canonical Bedrock model ID.
- Notes that marketplace pricing is best-effort and the per-source unit
  conversions ($/1M and $/unit to $/1k).
- Links `bedrock.md` from the docs and metrics indexes (it was unlinked).

## Why

The collector gained the `AmazonBedrockFoundationModels` source, but the docs
still described only first-party families and tiers. Without this, the documented
`family`/`price_tier` values and `model_id` format do not match emitted metrics.

## Depends on

Describes behavior added in #934. Merge this after #934 so the docs do not
describe pricing the collector does not yet emit.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -9,7 +9,7 @@
   - [Release Process](contribute/releases.md)
 - [Metrics](metrics/README.md) - See what each service exposes
   - [Providers](metrics/providers.md)
-  - **AWS:** [EC2](metrics/aws/ec2.md), [S3](metrics/aws/s3.md), [RDS](metrics/aws/rds.md), [MSK](metrics/aws/msk.md), [ELB](metrics/aws/elb.md), [NAT Gateway](metrics/aws/natgateway.md), [VPC](metrics/aws/vpc.md)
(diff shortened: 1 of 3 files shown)


simaarfania:
Folded into #934, which now carries the docs commit (52a5de4). Closing.

PR #1016, opened 2026-06-08
docs: add deprecation policy

nikimanoledaki:
## Why

`cloudcost-exporter` reached `v1.0.0` and is used in production. Removing user-facing features (Prometheus metric names and labels, CLI flags) requires explicit guardrails. Major releases also require a deprecation policy in place so that removing features can be planned as part of the release process.

## What

Adds `docs/contribute/deprecation-policy.md` covering:

- When the deprecation policy applies (Prometheus metrics and labels, CLI flags)
- Steps to mark a feature as deprecated (warning log, docs update, replacement guidance, target removal version)
- Minimum notice window: 3 minor releases or 3 months, whichever is longer
- Removal at a major release

Links the policy from `docs/contribute/README.md`.

### DIFF
--- a/docs/contribute/deprecation-policy.md
+++ b/docs/contribute/deprecation-policy.md
@@ -6,7 +6,7 @@ Backward compatibility and feature stability matter, especially since cloudcost-
 
 1. Mark a feature as deprecated - example: https://github.com/grafana/cloudcost-exporter/pull/467:
    1. Open a PR so that cloudcost-exporter logs a deprecation warning.
-   2. Update docs with a warning as well.
+   2. Update any docs with a warning as well. For example, update the service's docs if this is a Prometheus metric or label change; if this is a CLI flag change, find any references to it in the docs and add a warning, including in the Helm chart and its docs.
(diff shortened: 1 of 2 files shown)


docs/contribute/deprecation-policy.md
```diff
@@ -0,0 +1,13 @@
+# Deprecation Policy
+
+Backward compatibility and feature stability matter, especially since cloudcost-exporter is used in production systems and reached its first major (`x.0.0`) [semver](https://semver.org/#semantic-versioning-specification-semver) version. The project should follow a deprecation policy for removing existing features to minimize breaking changes. This applies to features like Prometheus metric names and labels and CLI flags.
+
+## Steps
+
+1. Mark a feature as deprecated - example: https://github.com/grafana/cloudcost-exporter/pull/467:
+   1. Open a PR so that cloudcost-exporter logs a deprecation warning.
+   2. Update docs with a warning as well.
```
suntala: It would be helpful to specify which doc.

nikimanoledaki: thanks @suntala, I added more info

suntala (approved):
Good idea!

PR #1018, opened 2026-06-08
fix(rds): cap per-region DescribeDBInstances with a timeout

rooneyshuman:
## What

Cap each region's `ListRDSInstances` (`DescribeDBInstances`) call with a per-region context timeout so a slow or unreachable region fails fast instead of consuming the whole collector budget.

## Why

After parallelizing the RDS region fan-out (#1017), the `prod-us-east-0` Scrape Availability SLO was still burning.

`scrape_duration_seconds` was repeatedly hitting ~55s and `up` dropped to 0 at exactly those points: a single slow/unreachable region rode the full ~60s collector budget via the AWS SDK's retries (`maxRetryAttempts = 10`), pushing the `/metrics` scrape past its `scrape_timeout` (~55s) and failing the whole scrape. Because per-region list errors are swallowed (logged, not returned), `cloudcost_exporter_collector_last_scrape_error` stayed clean while `up` went red — so only `up` reflected the problem.

The fan-out stopped one slow region from starving the others, but didn't bound how long that one region could take. This change adds the missing wall-clock ceiling.

## How

- Wrap each region's `ListRDSInstances` call in `context.WithTimeout(ctx, regionListTimeout)` inside `fetchInstancesData`.
- Default `regionListTimeout = 15s`, held as a `Collector` field (set in `New`) so it's injectable in tests.
- Behavior on timeout is unchanged from the existing per-region error path: the slow region is logged and skipped; healthy regions still report.

## Testing

- Added `TestCollector_Collect_SlowRegionFailsFast`: one region blocks until its context is canceled, the other is healthy. Asserts `Collect` returns quickly (runs in ~0.05s under a 50ms test timeout), the healthy region still emits a metric, and the slow region is skipped without blocking.
- `go build ./...`, `go vet ./pkg/aws/rds/`, and `go test -race ./pkg/aws/rds/` all pass.

### DIFF
--- a/pkg/aws/rds/rds.go
+++ b/pkg/aws/rds/rds.go
@@ -32,13 +32,14 @@ var (
 
 // Collector is a prometheus collector that collects metrics from AWS RDS clusters.
 type Collector struct {
-	regions        []types.Region
-	regionMap      map[string]client.Client
-	scrapeInterval time.Duration
-	Client         client.Client
-	pricingMap     *pricingMap
-	accountID      string
-	logger         *slog.Logger
+	regions           []types.Region
+	regionMap         map[string]client.Client
+	scrapeInterval    time.Duration
+	regionListTimeout time.Duration
+	Client            client.Client
+	pricingMap        *pricingMap
+	accountID         string
+	logger            *slog.Logger
 }
 
 type Config struct {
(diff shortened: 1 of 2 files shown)


stephan-rayner (approved):

pkg/aws/rds/rds.go
```diff
@@ -51,18 +52,25 @@ type Config struct {
 
 const (
 	serviceName = "RDS"
+
+	// regionListTimeout caps a single region's DescribeDBInstances call so a slow
+	// or unreachable region fails fast instead of consuming the whole collector
+	// budget. Without it, one region riding the full ~60s budget overruns the
+	// Prometheus scrape_timeout and fails the scrape (up=0).
+	regionListTimeout = 15 * time.Second
```
logyball, line 60: This should be overrideable by the user and defaulted to a 15s valude, not hardcoded.  The 60s scrape interval is arbitrary - I can imagine users would much rather have complete slow data rather than incomplete fast data.  We should at least make this a config/CLI option

Issue #1043, opened 2026-06-11
Add experimental service flags to iterate on new collectors without semver breakage


simaarfania:
## Problem

New collectors land while their metric and label schema is still settling. `bedrock` (#934) and `vertex` (#929) are good examples, both have iterated on label names, `price_tier` shapes, and `model_id` derivation. Because emitted metrics fall under the exporter's backward-compatibility contract, any later label change is a breaking change, which pushes us to get everything right in a single PR and makes post-merge iteration costly.

## Proposal

Add an opt-in experimental service set per provider that is explicitly outside the backward-compatibility contract:

```
--aws.experimental.services=bedrock
--gcp.experimental.services=vertex
```

Collectors enabled this way may change or remove labels and metrics without a major version bump. They are off by default, so a stable deployment is unaffected and nothing breaks for users who have not opted in.

## Behavior

- A parallel `*.experimental.services` `StringSliceFlag` alongside the existing `*.services`, wired into the same provider registration path.
- Experimental collectors are logged as experimental at startup so it is obvious in a deployment.
- A collector graduates by moving from `*.experimental.services` to `*.services` once its schema is stable, at which point the contract applies.

## Scope

- Design the mechanism once and share it across providers rather than per-collector one-offs.
- Initial candidates: `bedrock` and `vertex`.

## Open questions

- Naming: `*.experimental.services` vs an `experimental.` prefix convention.
- Does `bedrock` ship stable via #934 and experimental applies to future work, or does it start experimental-only so we can keep iterating on its labels?
- How to mark experimental beyond the flag: a startup log line, a note in the metric `HELP` text, or the flag alone.
- Graduation criteria (what "stable" means for a collector).

## Acceptance criteria

- [x] `*.experimental.services` flag exists for AWS and GCP and registers collectors from the union with `*.services`.
- [x] Experimental collectors are surfaced as experimental at startup.
- [x] Docs describe the experimental flag and that its metrics are not under the compatibility contract.
- [ ] At least one collector (`bedrock` or `vertex`) wired through it.

PR #1046, opened 2026-06-12
feat(provider): add experimental.services flags for AWS, GCP, and Azure

simaarfania:
Adds `--aws.experimental.services`, `--gcp.experimental.services`, and `--azure.experimental.services`: opt-in flags that enable collectors whose metrics are not covered by the backward-compatibility contract and may change. They let a collector iterate on labels without forcing a major version bump.

## Summary

- Each flag takes a comma-separated and/or repeated list, the same handling as the matching `--{provider}.services`.
- Each provider concatenates its stable and experimental service lists and registers them through its existing dispatch `switch`, so an experimental service uses the same collector. Registration is handled inline per provider, matching how the repo keeps provider logic self-contained.
- When registering a service from the experimental portion, the provider logs a warning: `registering experimental collector; its metrics are not covered by the backward-compatibility contract and may change`.
- This is an opt-in plus a signal; no collector is moved to experimental here.

## Tests

- `Test_NewWithDependencies` (AWS) and `Test_New_experimentalService` (Azure) assert an experimental service registers a collector and emits the warning.
- `Test_selectProvider_forwardsExperimentalServices` asserts each provider's `Config.ExperimentalServices` receives its flag value.

## Follow-up

Nothing restricts which collector may be experimental yet, so `--gcp.experimental.services=GKE` would register `GKE` with the warning. Registry-driven enforcement (mark a collector experimental, reject non-experimental names) is a follow-up. Part of #1043.


stephan-rayner (dismissed):
Hey Sima, one small thing in Test_New_experimentalService in azure_test.go (around line 41): the doc comment mentions "routes it through
  provider.MergeServiceEntries" but that function doesn't exist anywhere in the repo. Looks like a leftover from an earlier design.

I'm easy on this as it is a comment and we can clean it up later. I'm approving

jjo:
### Code review

Found 1 issue:

1. Experimental services are not deduplicated against stable services. Passing the same service in both `-{provider}.services` and `-{provider}.experimental.services` registers its collector twice, so `RegisterCollectors` hits a Prometheus duplicate-descriptor error and the exporter fails to start. This is the exact case a dedup guard earlier in this branch handled — its own comment said it "keeps a collector graduating from experimental to stable from registering twice and failing collector registration" — but the final per-provider inlining refactor (`a55ae24`) dropped it. The same `slices.Concat` + index pattern with no dedup is repeated in all three providers (`pkg/azure/azure.go#L113-L120`, `pkg/google/gcp.go#L116-L125`).

https://github.com/grafana/cloudcost-exporter/blob/a55ae249f8c620f21edbe7e04c01ab8f42706341/pkg/aws/aws.go#L176-L188

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

jjo (dismissed):
LGTM, don't think https://github.com/grafana/cloudcost-exporter/pull/1046#issuecomment-4694068418 Claude's review is blocking, as nevertheless it's op' resposibility to use experimental svcs correctly :), that said, it'd be cool to get it fixed

simaarfania:
> LGTM, don't think [#1046 (comment)](https://github.com/grafana/cloudcost-exporter/pull/1046#issuecomment-4694068418) Claude's review is blocking, as nevertheless it's op' resposibility to use experimental svcs correctly :), that said, it'd be cool to get it fixed

Definitely think it's in our best practices to fix this. Pushing up the fix, thanks @jjo and Claude for the stellar catch!

stephan-rayner (approved):

rooneyshuman (approved):

jjo (approved):

PR #1051, opened 2026-06-16
feat(exporter): add opt-in leader election to deduplicate collection across replicas

### LINKED ISSUE #1027: Support leader election to avoid duplicate collection across replicas
When cloudcost-exporter runs with more than one replica, each replica independently collects the same pricing data from the cloud provider APIs. This multiplies load on the CSP pricing/list APIs for identical data with no benefit — every replica exports the same rates.

## Proposal

Add an opt-in leader-election mechanism (e.g. Kubernetes lease-based) so only one replica actively collects from the upstream cloud provider APIs at a time.

## Considerations

- Flag-gated, default off — preserves current single-replica behavior.
- Sharding (splitting collection work across replicas) e.g. `kube-state-metrics` would be excellent, but is out of scope for an initial go at it
- Define non-leader behavior: non-leader replicas should serve basic up/down metrics

## Acceptance criteria

- Running multiple replicas results in a single set of upstream API calls.
- Behavior is configurable via flag.
- Documented in the config/metrics docs.

somaz94:
Adds opt-in, lease-based leader election so running multiple replicas no longer multiplies load on the cloud provider pricing APIs. Off by default, so single-replica deployments are unchanged.

When enabled (`-leader-election.enabled`):
- Replicas elect one leader via a Kubernetes `Lease`.
- The leader registers the provider collectors and calls the cloud provider APIs.
- Non-leaders serve `/metrics` with the base operational metrics plus a `cloudcost_exporter_leader_election_is_leader` gauge, but do not call the provider APIs.
- Losing the lease stops collection and shuts the replica down so it rejoins as a candidate.

The result is a single set of upstream API calls regardless of replica count. Sharding (splitting work across replicas) is intentionally out of scope, per the issue.

### Validation
- `make lint` — 0 issues
- `go test ./...` — all pass (new `pkg/leaderelection` tests including a fake-clientset leadership-acquisition test; updated `cmd/exporter` tests for the split registry handler)
- `make build-binary` — success
- New dependencies: `k8s.io/client-go`, `k8s.io/apimachinery`, `k8s.io/api` @ `v0.34.2` (stable)

### Docs
- `docs/deploying/leader-election.md` — flags, required RBAC (`leases` in `coordination.k8s.io`), and the leader gauge
- README link from the Kubernetes deployment section

closes: #1027


somaz94:
Hi — this PR closes #1027, a `good first issue` filed by @logyball asking for exactly this.

Two small asks:

1. **Approve the queued workflow runs.** All five workflows (`Tests`, `PR Title`, `Suggest Release Label`, `zizmor`, `TruffleHog`) sit at `action_required` — the first-time-contributor gate — so the suite has never actually executed. I can't self-serve that one.
2. A sanity check on the shape, if the approach looks right.

Just rebased onto `main` (only `go.mod`/`go.sum` conflicted — Renovate churn) and re-signed the commit, so the `signed-commits` report is now clean. Locally `go build ./...`, `go vet`, and `go test` on `cmd/exporter{,/config,/web}` + `pkg/leaderelection` pass.

Leader election is off by default (`--leader-election.enabled=false`), so single-replica installs are behaviourally byte-identical.

somaz94:
Following up on my 2026-08-06 comment. The branch has since gone out of date with main and now conflicts, so I would rather ask than keep rebasing it.

This closes #1027, a good first issue filed by @logyball, and adds opt-in leader election so multiple replicas do not duplicate collection. It is +569/-41 across 9 files and off by default.

If there is interest I will rebase and ask again for the workflow approval. If the direction has changed, I am happy to close it.

logyball:
Hi @somaz94 ! Apologies for taking this long (I've been on leave).  We're discussing this internally as we speak and will get back to you soon about the direction we'd like to take this.

logyball:
> Hi @somaz94 ! Apologies for taking this long (I've been on leave). We're discussing this internally as we speak and will get back to you soon about the direction we'd like to take this.

To follow up on this: we're currently working internally on a design to better capture the "HA" story than just leader-election.

somaz94:
Sounds good — I'll close this in favour of your HA design. Thanks for following up, and good luck with it!

PR #1057, opened 2026-06-23
fix(gke): add mutex to PricingMap to prevent concurrent map read/write panic

### LINKED ISSUE #1036: fix(gke): add mutex to PricingMap to prevent concurrent map read/write panic
## Background

`PricingMap` has no synchronization. `ParseSkus` writes to `pm.compute` and `pm.storage` in-place when the 24-hour refresh ticker fires. `Collect` reads those same maps on every scrape via `GetCostOfInstance` and `GetCostOfStorage`.

Go's runtime detects concurrent map read/write as a fatal error, so if a scrape coincides with a refresh tick the exporter panics and restarts.

## Impact

Low probability: the race window is open for roughly the duration of `ParseSkus` once every 24 hours. When it does trigger, the exporter crashes and recovers on restart. No data corruption, but an unnecessary outage.

## Fix

Add a `sync.RWMutex` to `PricingMap`: write-lock in `ParseSkus`, read-lock in `GetCostOfInstance` and `GetCostOfStorage`. Alternatively, build a fresh map in `Populate` and swap it in atomically under a write-lock.

ohyesgocool:
## What

`PricingMap.compute` and `.storage` (pkg/google/gke) are written in place by
`ParseSkus` when the 24h refresh ticker fires, and read by `Collect` on every
scrape via `GetCostOfInstance` and `GetCostOfStorage`. With no synchronization,
a scrape coinciding with a refresh tick trips Go's "concurrent map read and map
write" detector and crashes the exporter (it recovers on restart).

Fixes #1036.

## Change

- Add a `sync.RWMutex` to `PricingMap`: write-lock `ParseSkus`, read-lock
  `GetCostOfInstance` and `GetCostOfStorage`. Network calls in `Populate` stay
  outside the lock (mirrors the `pkg/google/networking` pattern).
- `GetCostOfStorage` returns a copy of `StoragePrices` so the value read after
  the lock is released can't be mutated by a later refresh.
- Pointer-ize the `TestStructuredPricingMap_GetCostOfInstance` table so the new
  mutex field doesn't trip `go vet` copylocks.
- Add `TestPricingMap_ConcurrentAccess`, a regression test that runs `ParseSkus`
  concurrently with the getters.

## Testing

`go test -race ./pkg/google/gke/` — trips the race detector on `main`, clean
with this change.

Suggested release label: `release:patch` (bug fix affecting the binary).

### DIFF
--- a/pkg/google/gke/pricing_map.go
+++ b/pkg/google/gke/pricing_map.go
@@ -7,6 +7,7 @@ import (
 	"log/slog"
 	"regexp"
 	"strings"
+	"sync"
 	"time"
 
 	"cloud.google.com/go/billing/apiv1/billingpb"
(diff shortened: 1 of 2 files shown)


ohyesgocool:
Hi @nikimanoledaki Can i have some review here please ? 😄

PR #1060, opened 2026-06-25
feat(bedrock): label token prices by type, region, and quota tier

simaarfania:
Replaces the overloaded `price_tier` label and the per-direction metric names on the AWS Bedrock pricing metrics with a separate label per dimension, so downstream joins key on each dimension directly.

## Summary

- Merges `cloudcost_aws_bedrock_input_usd_per_1k_tokens` and `cloudcost_aws_bedrock_output_usd_per_1k_tokens` into one `cloudcost_aws_bedrock_usd_per_1k_tokens`, distinguished by a `token_type` label (`input`, `output`, `cache_read`, `cache_write`).
- Replaces the composed `price_tier` string with a label per dimension: `region_tier` (`in`/`cross`) and `quota_tier` (`standard`/`batch`/`flex`/`priority`/`latency_optimized`) on both the token and search metrics, plus `cache_ttl` (`""`/`5m`/`1h`) on the token metric (set for `cache_write`).
- `cloudcost_aws_bedrock_search_unit_usd_per_1k_search_units` keeps its name and gains `region_tier`/`quota_tier` (no `token_type`/`cache_ttl`, a different unit).
- The pricingstore composite key is widened to carry the decomposed fields; `Collect` expands them back into labels.

## Why

The old `price_tier` packed three independent dimensions into one string (`composeTier(crossRegion, op, quota)`, e.g. `cross_region_cache_read`, `cache_write_5m`, `on_demand_batch`), and token direction lived in the metric name. Any consumer wanting to filter or join by tier or token type had to parse the composed string or union the two metric names. The gen-AI Bedrock cost recording rules in `grafana/deployment_tools` are the first such consumer: their price-side join is an 8-branch `or` with per-branch `label_replace`. With a label per dimension that collapses to a single selector keyed on `(model_id, region_tier, token_type)`.

Decomposing also fixes a latent bug in the standard (`AmazonBedrock`) source: `extractPriceTier` returned `cross_region` early and dropped the quota qualifier, so cross-region prices that differ only by quota (`batch`/`flex`/`priority`) collided into one series and overwrote each other. Verified against live pricing (one account, `amazon|anthropic`): in-region series are identical (1992 to 1992, a 1:1 re-encoding), and cross-region recovers the dropped prices (1782 to 2196, +414), with `flex`/`priority` cross-region buckets that did not exist before.

## After merge

This is a precondition for the gen-AI Bedrock cost recording rules in `grafana/deployment_tools` (grafana/deployment_tools#621203, still in draft). Those rules will switch their price-side join to the new labels once a release carrying this change is scraped; nothing depends on this metric in production yet.

## Metric change

Bedrock is an experimental collector (enabled via `-aws.experimental.services`), so its metrics are outside the backward-compatibility contract and may change without a major release. This renames the token price metrics and replaces `price_tier`: `cloudcost_aws_bedrock_input_usd_per_1k_tokens` and `cloudcost_aws_bedrock_output_usd_per_1k_tokens` become `cloudcost_aws_bedrock_usd_per_1k_tokens{token_type=...}`, and `price_tier` is replaced by `token_type`/`region_tier`/`quota_tier`/`cache_ttl`. Anyone scraping these should move to the new labels. `docs/metrics/aws/bedrock.md` is updated. The Vertex collector is unaffected (it keeps the `input`/`output` token-cost suffixes).

### DIFF
--- a/pkg/aws/bedrock/bedrock.go
+++ b/pkg/aws/bedrock/bedrock.go
@@ -49,14 +49,10 @@ const (
 	cacheTTL5m = "5m"
 	cacheTTL1h = "1h"
 
-	// metricKind selects the output metric when a pricing key is decoded at collection time.
-	kindToken  = "token"
-	kindSearch = "search"
-
 	// compositeKeySep separates the fields encoded in the pricingstore usagetype key:
-	// kind|family|model_id|token_type|region_tier|quota_tier|cache_ttl
+	// family|model_id|token_type|region_tier|quota_tier|cache_ttl
 	compositeKeySep    = "|"
-	compositeKeyFields = 7
+	compositeKeyFields = 6
 
 	// searchUnitsPerKilo converts from per-unit pricing (as published by the AWS Pricing API)
 	// to per-1000-units pricing (as emitted by SearchUnitCostDesc).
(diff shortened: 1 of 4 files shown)


pkg/aws/bedrock/bedrock.go
```diff
@@ -389,36 +440,43 @@ var tokenTypeMarkers = []string{
...
-	case strings.HasSuffix(lower, "-batch"):
-		return priceTierOnDemandBatch
-	case strings.HasSuffix(lower, "-flex"):
-		return priceTierOnDemandFlex
-	case strings.HasSuffix(lower, "-priority"):
-		return priceTierOnDemandPriority
-	default:
-		return priceTierOnDemand
-	}
+	case strings.Contains(lower, "batch"):
```
leonorfmartins, line 473: should we also include a  `latency_optimized` case since it seems like one of the tiers? https://github.com/grafana/cloudcost-exporter/pull/1060/changes#diff-cf77c926b9166f64b82fcf1933dc3ec66dd50dd7691f5b04d37b50e517c268f0R52

simaarfania: Good catch! `standardTier` mapped `batch`/`flex`/`priority` but not `latency_optimized`, even though `marketplaceTier` and `quota_tier` already had it. Fixed this.

nikimanoledaki (approved):

nikimanoledaki:
LGTM! Really happy to see that it was relatively easy to simplify the [previous query](https://ops.grafana-ops.net/goto/swvt9q?orgId=stacks-27821)!

PR #1083, opened 2026-07-13
feat(vertex): align with the Bedrock collector and price Claude on Vertex

### LINKED ISSUE #1067: feat(vertex): align GCP Vertex metric shape with AWS Bedrock
#1060 simplified the Bedrock token metrics: direction-named metrics (`input_`/`output_`) were replaced by a single `cloudcost_aws_bedrock_usd_per_1k_tokens` metric with a `token_type` label, and the composed `price_tier` label was decomposed into orthogonal `region_tier`, `quota_tier`, and `cache_ttl` labels.

The Vertex collector predates that work and still uses the old pattern. Applying the same simplification would greatly simplify our cost metrics and cost recording rules.

## Current Metrics

```
cloudcost_gcp_vertex_input_usd_per_1k_tokens{model_id, family, region, price_tier}
cloudcost_gcp_vertex_output_usd_per_1k_tokens{model_id, family, region, price_tier}
cloudcost_gcp_vertex_input_usd_per_1k_characters{model_id, family, region, price_tier}
cloudcost_gcp_vertex_output_usd_per_1k_characters{model_id, family, region, price_tier}
cloudcost_gcp_vertex_instance_total_usd_per_hour{machine_type, use_case, region, price_tier}
cloudcost_gcp_vertex_search_unit_usd_per_1k_search_units{model_id, family, region}
```

## Target Metrics

```
cloudcost_gcp_vertex_usd_per_1k_tokens{project_id, model_id, family, region, token_type, region_tier, quota_tier}
cloudcost_gcp_vertex_usd_per_1k_characters{project_id, model_id, family, region, token_type, region_tier, quota_tier}
cloudcost_gcp_vertex_instance_total_usd_per_hour{project_id, machine_type, use_case, region, region_tier}
cloudcost_gcp_vertex_search_unit_usd_per_1k_search_units{project_id, model_id, family, region, region_tier, quota_tier}
```

## Proposed changes

- **Consolidate input/output into a single metric with `token_type`:** Replace `cloudcost_gcp_vertex_input_usd_per_1k_tokens` + `cloudcost_gcp_vertex_output_usd_per_1k_tokens` with `cloudcost_gcp_vertex_usd_per_1k_tokens{token_type="input"|"output"}`, matching the `TokenCostSuffix` + `token_type` pattern already used by Bedrock. Apply the same consolidation to the character metrics: replace the two separate metrics with `cloudcost_gcp_vertex_usd_per_1k_characters{token_type="input"|"output"}`.
- **Add `project_id` label:** Bedrock carries `account_id` to identify the billing scope. Vertex should carry `project_id` for the equivalent. This is available from `Config.Projects`.
- **Align tier label naming:** Vertex has a single `price_tier` label. Bedrock splits this into `region_tier` and `quota_tier`. If Vertex pricing has equivalent dimensions, match the label names. If not, document why `price_tier` stays consolidated.
- **Add `project_id` to the rerank metric:** `cloudcost_gcp_vertex_search_unit_usd_per_1k_search_units` is currently missing `price_tier` (present on all other Vertex metrics) and would need `project_id` added alongside the above.

These are experimental metrics (`--gcp.experimental.services=vertex`), so no backward-compatibility constraint applies. This is the right time to make breaking schema changes.

simaarfania:
Bring the GCP Vertex collector in line with the AWS Bedrock collector, and price the model the Assistant actually runs on Vertex (Claude).

Closes #1067

## Summary

Experimental Vertex metrics (`--gcp.experimental.services=vertex`).

**Metric shape (#1067)**
- Consolidate the split `_input_`/`_output_` token and character metrics into `cloudcost_gcp_vertex_usd_per_1k_tokens` / `_usd_per_1k_characters`, keyed by `gen_ai_token_type`.
- Rename `model_id` -> `gen_ai_request_model` (matches the Bedrock token metric).
- Single `project_id` label (the auth project), stamped on every series, mirroring Bedrock's single `account_id` from STS. No per-project fan-out (prices are project-independent).

**Family filter**
- `--gcp.vertex.families`, a regex on `family`, mirroring `--aws.bedrock.families`. Default `.*`.

**Focus on model pricing (drop what doesn't map)**
- Remove the `compute` metric: it priced infra nodes (out of scope) and its regex never matched the live catalog, so it emitted zero series.
- Fix reranking: the regex expected `... Ranking Requests`, but the live SKU is `Vertex AI Search: Ranking` (the Semantic Ranker), so it emitted nothing. Match it, label `gen_ai_request_model=semantic-ranker`.
- Skip `AI Dev Tools:` (Agent Builder) SKUs that were leaking into the token metric as junk `ai-dev-tools:-claude-*` models.

**Price Claude on Vertex (`family="anthropic"`)**
- Claude prices are not in the public Cloud Catalog; they are account-scoped. Add a v1beta Cloud Billing client that lists the per-model `Claude ...` services under the project's billing account and reads their **list** prices, with `gen_ai_token_type` including `cache_read`/`cache_write`.
- The billing account is resolved from `--project-id` via `getBillingInfo` (read-only), so no new flag is needed, consistent with the other GCP collectors and with how the AWS collectors derive the account from STS.
- Best-effort: if the billing account can't be resolved or has no Claude SKUs, Claude is left unpriced and the rest still emits. Model slugs match Sigil's `gen_ai_request_model`; regional endpoints carry their own +10% price via the `region` label.

Resulting schema:

```
cloudcost_gcp_vertex_usd_per_1k_tokens{project_id, region, gen_ai_request_model, family, gen_ai_token_type, price_tier}
cloudcost_gcp_vertex_usd_per_1k_characters{project_id, region, gen_ai_request_model, family, gen_ai_token_type, price_tier}
cloudcost_gcp_vertex_search_unit_usd_per_1k_search_units{project_id, region, gen_ai_request_model, family, price_tier}
```

## Why

#1060 simplified the Bedrock token metrics and the labels were later renamed to `gen_ai_request_model` / `gen_ai_token_type`. The Vertex collector predated that and kept the old pattern, so aligning the two simplifies the shared cost recording rules. Claude is the Assistant's Vertex completion model, so pricing it is what makes the price-vs-usage join actually work for Vertex.

`price_tier` stays a single composed label rather than splitting into Bedrock's `region_tier` / `quota_tier`: Vertex has no cross-region inference tier, and `price_tier` composes dimensions that do not decompose cleanly. Documented in `docs/metrics/gcp/vertex.md`.

Experimental metrics, so the schema change carries no backward-compatibility constraint.

## Notes

- Claude list prices come from the account-scoped v1beta API because GCP does not publish them in the public catalog (verified). The collector reads `listPrice`, which is account-independent; only the account's negotiated `contractPrice` would differ, and it is not used.
- The account-scoped fetch adds startup latency proportional to the number of Claude SKUs (one price call each); it runs only when the project's billing account exposes `Claude ...` services.

### DIFF
--- a/cmd/exporter/config/config.go
+++ b/cmd/exporter/config/config.go
@@ -28,6 +28,7 @@ type Config struct {
 			Services                 StringSliceFlag
 			ExperimentalServices     StringSliceFlag
 			GKEZoneConcurrency       int
+			VertexFamilyFilter       string
 		}
 		Azure struct {
 			Services             StringSliceFlag
--- a/cmd/exporter/exporter.go
+++ b/cmd/exporter/exporter.go
@@ -87,6 +87,7 @@ func providerFlags(fs *flag.FlagSet, cfg *config.Config) {
 	fs.Var(&cfg.Providers.Azure.ExperimentalServices, "azure.experimental.services", "Experimental Azure service(s); their metrics are not covered by the backward-compatibility contract and may change. Run with -list-services to see available values.")
 	fs.Var(&cfg.Providers.GCP.Services, "gcp.services", "GCP service(s). Run with -list-services to see available values.")
 	fs.Var(&cfg.Providers.GCP.ExperimentalServices, "gcp.experimental.services", "Experimental GCP service(s); their metrics are not covered by the backward-compatibility contract and may change. Run with -list-services to see available values.")
+	fs.StringVar(&cfg.Providers.GCP.VertexFamilyFilter, "gcp.vertex.families", ".*", "Regex matched against the Vertex model family label. Only matching families are emitted.")
 	flag.StringVar(&cfg.Providers.AWS.Region, "aws.region", "", "AWS region")
 	flag.StringVar(&cfg.Providers.AWS.RoleARN, "aws.roleARN", "", "Optional AWS role ARN to assume for cross-account access.")
 	fs.StringVar(&cfg.Providers.AWS.BedrockFamilyFilter, "aws.bedrock.families", ".*", "Regex matched against the Bedrock model family label. Only matching families are emitted.")
@@ -288,6 +289,7 @@ func selectProviderWith(
 			ExperimentalServices: strings.Split(cfg.Providers.GCP.ExperimentalServices.String(), ","),
 			CollectorTimeout:     collectorTimeout,
 			GKEZoneConcurrency:   cfg.Providers.GCP.GKEZoneConcurrency,
+			VertexFamilyFilter:   cfg.Providers.GCP.VertexFamilyFilter,
 		})
 
 	default:
--- a/docs/metrics/gcp/vertex.md
+++ b/docs/metrics/gcp/vertex.md
@@ -32,7 +32,9 @@ Claude on Vertex (`family="anthropic"`) is priced automatically, no extra config
 are account-scoped rather than public catalog SKUs, so the collector resolves the billing account
 from `--project-id` (via `getBillingInfo`) and reads them from the Cloud Billing v1beta pricing API.
 The lookup is best-effort: if the billing account cannot be resolved or has no Claude SKUs, Claude is
-left unpriced and the other Vertex prices are still emitted.
+left unpriced and the other Vertex prices are still emitted. It runs off the startup path (the
+collector starts with catalog prices and adds Claude on the first background refresh), so Claude
+series appear a few seconds after start.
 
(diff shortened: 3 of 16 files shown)


stephan-rayner (approved):

rooneyshuman (approved):

stephan-rayner:
With this being merged https://github.com/grafana/cloudcost-exporter/issues/1085 can be closed

Issue #1107, opened 2026-08-06
Roll out the gcp_gke latency fix (PR #999, -73% latency) to the worst-offending cloudcost-exporter collectors


stephan-rayner:
Reference Pattern: #999

### Background

[cloudcost_exporter_collector_duration_seconds](https://ops.grafana-ops.net/explore?schemaVersion=1&panes=%7B%223p8%22:%7B%22datasource%22:%22edprtf91hz01se%22,%22queries%22:%5B%7B%22refId%22:%22B%22,%22expr%22:%22sort_desc%28%20%20%5Cn%20%20histogram_quantile%280.99,%20%20%5Cn%20%20%20%20sum%20by%28collector%29%20%28rate%28cloudcost_exporter_collector_duration_seconds%5B1h%5D%29%29%20%20%5Cn%20%20%29%20%20%5Cn%29%20%20%22,%22range%22:true,%22instant%22:true,%22datasource%22:%7B%22type%22:%22prometheus%22,%22uid%22:%22edprtf91hz01se%22%7D,%22editorMode%22:%22code%22,%22legendFormat%22:%22__auto%22,%22hide%22:false%7D%5D,%22range%22:%7B%22from%22:%22now-1h%22,%22to%22:%22now%22%7D,%22compact%22:false%7D%7D&from=2026-08-06T19:45:47.990Z&to=2026-08-06T19:45:47.990Z&timezone=utc) is a native histogram (defined in `pkg/collectormetrics/collectormetrics.go`) tracking how long each cloud-cost collector takes per scrape, labeled by `collector`, `provider`, and `region`. A stack rank by average and p99 duration surfaced clear outliers worth investigating.

The codebase has 18 collector implementations that resolve to 17 distinct `collector` label values (both `aws/vpc` and `google/vpc` emit `collector="VPC"`). The rankings below show 16 rows: `azure_blob` produced no data in the measured window and is absent. See "Measurement caveats" before reading the numbers as absolute.

---

### Measurement caveats

- **Per-provider deployment.** Each exporter deployment runs exactly one provider, so collectors only compete within a single scrape when they share a provider. `aws_rds`'s p99 only affects AWS deployments; `ForwardingRule` only affects GCP. There is no single global scrape cycle across all 16 rows.
[...]

PR #1109, opened 2026-08-06
perf(aws/rds): pricing and inventory in a background store

### LINKED ISSUE #1108: aws_rds: Move Pricing off the Scrape Path to Fix the P99
Part of #1107. Reference pattern: #999.

### Background

Issue #1107 ranks `aws_rds` as the worst p99 offender in the collector fleet (p99 ~42.55s, avg ~4.48s). Before applying the PR #999 background-store pattern, we tested its key precondition: is `aws_rds` scrape latency actually dominated by live AWS API calls on the scrape path? It is, and the investigation pinned the exact cause of the p99 tail.

### Finding: Latency Is API-Bound on the Scrape Path

`Collect()` (`pkg/aws/rds/rds.go:76-156`) makes live AWS calls in two phases that map onto the avg-vs-p99 split:

1. **Instance listing (the ~4.5s average).** Per-region `DescribeDBInstances`, paginated at 100/page (`pkg/aws/client/rds.go:19-34`), run concurrently one goroutine per region (`rds.go:85-103`). Wall-clock is bounded by the slowest region. This is comparable to EC2's ~4.6s average.

2. **Pricing lookups (the ~42s p99 tail).** The pricing map (`pkg/aws/rds/pricing_map.go`) is never pre-warmed and never refreshed in the background. `New()` allocates an empty map (`rds.go:64`); it fills lazily, sequentially, one instance at a time inside the Collect loop (`rds.go:109-154`). Each cache miss makes a blocking pricing `GetProducts` call via `GetRDSUnitData` (`rds.go:126`, `pkg/aws/client/pricing.go:229-281`) before moving to the next instance.

### Root Cause of the P99 Spikes

In steady state the pricing map is warm, so pricing makes zero API calls and Collect is just the concurrent listing (~4.5s). After a pod restart, or when many new instance shapes appear, the map is cold and Collect issues U sequential pricing calls (U = distinct region + instance class + engine + deployment option + location keys), serialized inside the loop. That serialized fan-in is the "occasional severe spike" the parent issue flagged. The p99 is specifically the cold-pricing-map path, not the listing.

### aws_rds Is the Outlier

Every other AWS collector already moves pricing off the scrape path:

- **EC2** pre-warms two pricing maps in `New()` and refreshes them on `ScrapeInterval` tickers (`pkg/aws/ec2/ec2.go:90-124`); `Collect()` only reads them.
- **MSK** and **Bedrock** use the shared `pkg/aws/pricingstore`, pre-warmed in `New()` and refreshed on a `PriceRefreshInterval` ticker (`msk.go:113-133`, `bedrock.go:186-205`).

`aws_rds` is the only AWS collector still fetching pricing lazily on the scrape path with a cold-start empty map. This fix applies an established in-repo pattern to the one collector that skipped it.

### Proposed Fix

**Primary (fixes the p99): move RDS pricing off the scrape path.** Mirror EC2's pre-warm plus ticker approach rather than the GKE inventory store or the `pricingstore` package. `pricingstore` is not a clean fit: its snapshot parser expects EC2-shaped `productInfo` JSON keyed by `usagetype` (`pricing_store.go:214-227`), which does not match RDS's `region + class + engine + deployment + location` key scheme. EC2's own approach matches RDS directly and keeps the existing `pricingMap` type.

- Add a bulk `GenerateRDSPricingMap` that fetches all RDS on-demand prices per region via paginated `GetProducts` (reuse the paginating helper `getPricesFromProductList`, `pkg/aws/client/pricing.go:207-227`, instead of the single-product `GetRDSUnitData`) and parses them into the existing key scheme via `validateRDSPriceData`.
- Call it once in `New()` to pre-warm, then start a background ticker to refresh it, following `ec2.go:97-124`. RDS `New()` currently starts no goroutines.
- Remove the lazy `GetRDSUnitData`/`Set` path from `Collect()` (`rds.go:122-142`); Collect becomes a pure map read with a skip-and-log on pricing miss.

Result: `Collect()` makes zero pricing API calls and the cold-start serialized fan-in disappears, collapsing the p99.

**Secondary (optional, addresses the average): background instance store.** Apply the GKE background-store pattern (`region -> instances`, single-phase since RDS is regional) to move `ListRDSInstances` off the scrape path. EC2 deliberately keeps listing in `Collect()` via a worker pool and accepts a similar average, so this is a judgement call, not required. Recommend deferring unless the average becomes a problem.

### Critical Files

- `pkg/aws/rds/rds.go`: `New()` (add pre-warm plus ticker), `Collect()` (drop lazy pricing fetch, read-only pricing).
- `pkg/aws/rds/pricing_map.go`: add `GenerateRDSPricingMap`; keep the existing `pricingMap` type, `Get`/`Set`, and `validateRDSPriceData`.
- `pkg/aws/client/pricing.go`: reuse `getPricesFromProductList` (`:207-227`) for a bulk RDS product fetch.
- `pkg/aws/aws.go`: RDS wiring (`:226-245`) already passes `ScrapeInterval`; no new config expected.
- Reference precedents: `pkg/aws/ec2/ec2.go:90-124`, `pkg/aws/msk/msk.go:113-133`, `pkg/google/gke/{node_store,disk_store}.go` (for the optional secondary).

### Acceptance Criteria

- RDS pricing is pre-warmed in `New()` and refreshed on a background ticker; `Collect()` makes zero pricing API calls.
- Unit tests: cold-start Collect serves all prices from the pre-warmed map (no `GetRDSUnitData` calls), the ticker refresh path is covered, and the pricing-miss skip-and-log is covered.
- Metrics emit on the first scrape after startup.
- Post-deploy, the reproduction PromQL from #1107 shows `aws_rds` p99 dropped out of the tail, closing the root cause.

### Verification

- `make generate && make test` (regenerate mocks first; RDS client mocks change if a bulk pricing method is added).
- Run locally against a profile with RDS instances: `go run cmd/exporter/exporter.go -provider aws -aws.profile $AWS_PROFILE -aws.services RDS`. Confirm pricing populates in `New()` (startup logs), not during Collect, and that metrics emit on the first scrape.
- Re-run the #1107 reproduction queries after deploy to confirm the p99 improvement.

### LINKED ISSUE #1134: aws/rds: Aurora I/O-Optimized and Standard storage collide on the same pricing key
## Context

Raised by @simaarfania during review of #1109: https://github.com/grafana/cloudcost-exporter/pull/1109#discussion_r3962417201

The current pricing key (`createPricingKey` in `pkg/aws/rds/rds.go`) doesn't distinguish Aurora I/O-Optimized storage from Aurora Standard storage. Confirmed against the live AWS Pricing API (`AmazonRDS` bulk offer file, us-east-1): for the same `instanceType`, `databaseEngine`, `deploymentOption`, `licenseModel`, and `locationType`, there are two distinct `Database Instance` SKUs for many Aurora instance classes, differing only in the `storage` attribute:

| SKU | storage | usagetype |
|---|---|---|
| `SVB4AAU3H83DPGMK` | `EBS Only` | `InstanceUsage:db.r6g.large` |
| `WJD7AAF57MBUC6E6` | `Aurora IO Optimization Mode` | `InstanceUsageIOOptimized:db.r6g.large` |

Both produce the same pricing key today, so whichever price is parsed last for that key wins in the pricing map. This is a live correctness bug for any Aurora cluster running I/O-Optimized storage: it can report the Standard price instead (or vice versa).

## Complication

The `storage` attribute isn't a clean discriminator on its own. For instance-store-backed Aurora instance classes, the same attribute instead describes local ephemeral disk (e.g. `1 x 950 NVMe SSD`), unrelated to the I/O-Optimized/Standard distinction. A fix can't add `storage` verbatim to the key: it needs to normalize to "is this the I/O-Optimized SKU" (`usagetype` contains `IOOptimized`, or `storage == "Aurora IO Optimization Mode"`) and treat everything else as Standard.

It's also not yet confirmed whether `DescribeDBInstances` surfaces the I/O-Optimized setting on the member DB instance (`StorageType` field), or only `DescribeDBClusters` does. If it's cluster-only, matching an instance to the right price may need an additional API call and a new IAM permission, which is a bigger change than a key fix alone.

## Task

- Confirm whether `rds:DescribeDBInstances` returns enough information (e.g. `StorageType`) to determine I/O-Optimized vs Standard per instance, or whether `DescribeDBClusters` is required.
- Add an explicit I/O-Optimized attribute to the pricing key, derived correctly from the Pricing API response rather than the raw `storage` string.
- Add table-driven tests covering both Aurora storage configurations.

Related to #1131 (different root cause: engine/edition matching, not storage configuration).

stephan-rayner:
## Summary

Moves AWS RDS pricing and instance listing off the Prometheus scrape path into two fully independent background stores, collapsing the collector's p99.

Part of #1107. Closes #1108.

## Problem

`aws_rds` was the fleet's worst p99 offender (p99 ~42.55s, avg ~4.48s). `Collect()` filled the pricing map lazily: on a cache miss it made a blocking `GetProducts` call per instance, one at a time, inside the scrape loop. In steady state the map is warm and a scrape is just the concurrent listing (~4.5s). After a pod restart, or when many new instance shapes appear, the map is cold and a scrape issues a serialized fan-in of pricing calls, which is the p99 tail. Every other AWS collector already moves pricing off the scrape path; RDS was the outlier.

## Change

Apply the background-store pattern (the GKE `node_store`/`pricing_map` shape) to RDS, with instance listing and pricing fully decoupled per review feedback:

- `instanceStore` refreshes `DescribeDBInstances` inventory per region on a ticker (`--scrape-interval`), starting at process startup.
- `pricingStore` independently refreshes RDS pricing on its own fixed 24 hour ticker. It lists every RDS Database Instance price in a region in bulk (`GetProducts`) rather than per instance, and keys each price by product attributes (instance type, engine, edition, deployment option, license, location type, and for Aurora, storage mode). It has no dependency on `instanceStore` and vice versa.
- `Collect()` joins an instance to its price by that key. It makes zero AWS calls, checks each store's readiness independently, and skips the whole scrape (no metrics, no error) if either is still cold, covering every not-ready combination (neither, instances-only, pricing-only). A pricing miss for an individual instance is skipped with a log rather than failing the scrape.

This addresses both the p99 (cold pricing fan-in leaves the scrape path) and the average (listing leaves the scrape path), and lets pricing and inventory refresh on cadences that fit how often each actually changes.

### Bounding and safety

Each store's per-region background work is wrapped in its own timeout. A positive `--aws.rds.region-timeout` is honoured by both stores as the operator's fail-fast budget; `0` (the default) falls back to an internal 2 minute ceiling per store. On the old scrape-path code, `0` was implicitly bounded by the collector interval. The background populate has no such outer deadline, so the ceiling guarantees a hung AWS call can never wedge either refresh, leave its readiness channel open, and skip every future tick.

Partial instances (AWS can return an instance missing its AZ, class, engine, or multi-AZ flag mid create or delete) are skipped rather than dereferenced, so a partial record cannot panic the background goroutine.

### Observability

Operational counter `cloudcost_exporter_aws_rds_populate_errors_total`, labeled `store` (`instances` or `pricing`), `region`, `operation`, surfaces populate failures. Instance-side operations are `lookup_client`, `list_instances`; pricing-side operations are `list_prices`, `parse_pricing`. Populate failures become missing metrics plus a bumped counter, not scrape errors, so alerting on this counter is recommended.

## Tradeoffs

- Cached inventory is stale by up to `--scrape-interval`; cached pricing by up to 24 hours.
- No metrics emit until both stores complete their first refresh. Since the two stores populate concurrently rather than sequentially, this cold-start gap is bounded by the slower of the two, not their sum.
- Multi-edition engine (Oracle, SQL Server) pricing-key attributes are best-effort and untested against a live Pricing API response; tracked separately in #1131.
- Aurora's pricing key additionally disambiguates Standard vs I/O-Optimized storage (they carry different instance-hour prices and previously collided on one key, closes #1134). This relies on `DBInstance.StorageType` surfacing `aurora-iopt1` at the instance level, which needs confirming against a live Aurora I/O-Optimized cluster before merge; flagged with a `VERIFY:` comment on `auroraStorageModeFromInstance` in `pkg/aws/rds/rds.go`.

## Testing

- `go test ./...` passes.
- `instanceStore`: warms instance inventory; a refresh overwrites the cache; readiness is signalled once; the overlap guard skips a concurrent populate; a failing or client-less region is counted and skipped without dropping its siblings; the per-region timeout is applied (configured value and the default ceiling).
- `pricingStore`: warms pricing; prices are listed once per region regardless of instance count; a refresh overwrites cached prices; readiness is signalled once; the overlap guard skips a concurrent populate; list and parse errors are counted under `store="pricing"` and skipped without dropping siblings; the per-region timeout is applied (configured value and the default ceiling).
- `Collector.Collect`: cold-start for every not-ready combination (neither store warm, instances-only, pricing-only) emits nothing and does not error; a warm `Collect` serves every instance from both stores and makes zero AWS calls; a pricing miss for one instance is skipped without failing the scrape; multi-region instances are all emitted.
- Aurora storage mode: Standard and I/O-Optimized produce distinct pricing keys on both the Pricing API attribute side and the instance `StorageType` side, and a missing or unrecognized value on either side is skipped rather than mis-keyed.

## Verification

- `make generate && make test`.
- Run locally against a profile with RDS instances and confirm both stores populate at startup (not during `Collect`) and metrics emit once both complete their first refresh.
- Re-run the #1107 reproduction queries after deploy to confirm `aws_rds` drops out of the p99 tail.

## Docs

`docs/metrics/aws/rds.md` documents the operational metric's `store` label values, the two independent refresh intervals, and the cold-start behavior.

### DIFF
--- a/docs/metrics/aws/rds.md
+++ b/docs/metrics/aws/rds.md
@@ -8,7 +8,7 @@
 
 | Metric name                                      | Metric type | Description                                                              | Labels                                                                                                                              |
 |--------------------------------------------------|-------------|-------------------------------------------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------|
-| cloudcost_exporter_aws_rds_populate_errors_total | Counter     | Errors during background store population, by store, region, and operation | `store`=&lt;always `instances`&gt; <br/> `region`=&lt;AWS region&gt; <br/> `operation`=&lt;`lookup_client`, `list_instances`, `get_pricing`, `validate_pricing`&gt; |
+| cloudcost_exporter_aws_rds_populate_errors_total | Counter     | Errors during background store population by store, region, and operation | `store`=&lt;always `instances`&gt; <br/> `region`=&lt;AWS region&gt; <br/> `operation`=&lt;`lookup_client`, `list_instances`, `get_pricing`, `validate_pricing`&gt; |
 
 ## Overview
 
--- a/docs/metrics/aws/rds.md
+++ b/docs/metrics/aws/rds.md
@@ -4,9 +4,17 @@
 |------------------------------------------------|-------------|--------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------|
 | cloudcost_aws_rds_hourly_rate_usd_per_hour     | Gauge       | Hourly cost of AWS RDS instances by region, tier, and instance ID. Cost represented in USD/hour | `account_id`=&lt;AWS account ID&gt; <br/> `region`=&lt;AWS region&gt; <br/> `tier`=&lt;RDS instance tier&gt; <br/> `id`=&lt;RDS instance ID&gt; <br/> `arn_name`=&lt;RDS instance ARN name&gt; |
 
+### Operational Metrics
+
+| Metric name                                      | Metric type | Description                                                              | Labels                                                                                                                              |
+|--------------------------------------------------|-------------|-------------------------------------------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------|
+| cloudcost_exporter_aws_rds_populate_errors_total | Counter     | Errors during background store population, by store, region, and operation | `store`=&lt;always `instances`&gt; <br/> `region`=&lt;AWS region&gt; <br/> `operation`=&lt;`lookup_client`, `list_instances`, `get_pricing`, `validate_pricing`&gt; |
+
 ## Overview
 
 The RDS collector exports pricing metrics for Amazon Relational Database Service instances across all configured AWS regions. It collects hourly pricing rates for RDS database instances based on their instance type and tier.
+
+Instance inventory (`DescribeDBInstances`) and pricing (`GetProducts`) are refreshed in the background on a ticker rather than on the Prometheus scrape path. A scrape reads the warm cache and makes zero AWS calls, so scrape latency no longer tracks AWS API latency or a cold pricing map. Warming starts at startup, so there is a short cold-start gap: no metrics emit until the first background refresh finishes. When a refresh fails to warm a price for an instance, that instance is skipped and `cloudcost_exporter_aws_rds_populate_errors_total` is incremented; the scrape still succeeds.
 
 ## Configuration
 
@@ -23,15 +31,15 @@ Or via command line:
 --aws.services=ec2,s3,rds
 ```
 
-### Per-region list timeout
+### Per-region Timeout
 
-The collector queries `DescribeDBInstances` in each region concurrently. `--aws.rds.region-timeout` bounds each region's call:
+The background refresh queries `DescribeDBInstances` and pricing in each region concurrently. `--aws.rds.region-timeout` bounds each region's work:
 
 ```bash
 --aws.rds.region-timeout=15s
 ```
 
-The default is `0`, which leaves each region bounded only by the collector context (`--collector-interval`) — a slow or unreachable region can consume the whole budget and overrun the Prometheus `scrape_timeout`, failing the scrape (`up=0`). Set a positive value (e.g. `15s`) to fail such regions fast: the region is logged and skipped while the others still report, keeping total collection under the scrape timeout and protecting scrape availability. Raise it to wait for complete data from slow regions at the cost of longer scrapes.
+The default is `0`, which applies an internal safety ceiling so a hung AWS call can never stall the background refresh. Set a positive value (e.g. `15s`) to fail slow or unreachable regions faster: the region is logged, counted in `cloudcost_exporter_aws_rds_populate_errors_total`, and skipped while the others still warm. Because the refresh runs off the scrape path, a slow region delays only the freshness of that region's data, not the scrape itself.
 
 ## Labels
 
@@ -44,7 +52,8 @@ The default is `0`, which leaves each region bounded only by the collector conte
 ## Notes
 
 - Pricing data is fetched from the AWS Pricing API (us-east-1 region)
-- Metrics are refreshed periodically based on the configured scrape interval
+- Inventory and pricing are refreshed in the background on the configured scrape interval; scrapes read the warm cache and make no AWS calls
+- No metrics emit until the first background refresh completes after startup
 - All costs are represented in USD per hour
 
 ## IAM Permissions
--- a/docs/metrics/aws/rds.md
+++ b/docs/metrics/aws/rds.md
@@ -8,13 +8,11 @@
 
 | Metric name                                      | Metric type | Description                                                              | Labels                                                                                                                              |
 |--------------------------------------------------|-------------|-------------------------------------------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------|
-| cloudcost_exporter_aws_rds_populate_errors_total | Counter     | Errors during background store population by store, region, and operation | `store`=&lt;always `instances`&gt; <br/> `region`=&lt;AWS region&gt; <br/> `operation`=&lt;`lookup_client`, `list_instances`, `get_pricing`, `validate_pricing`&gt; |
+| cloudcost_exporter_aws_rds_populate_errors_total | Counter     | Errors during background store population by store, region, and operation | `store`=&lt;always `instances`&gt; <br/> `region`=&lt;AWS region&gt; <br/> `operation`=&lt;`lookup_client`, `list_instances`, `list_prices`, `parse_pricing`&gt; |
 
 ## Overview
 
 The RDS collector exports pricing metrics for Amazon Relational Database Service instances across all configured AWS regions. It collects hourly pricing rates for RDS database instances based on their instance type and tier.
-
-Instance inventory (`DescribeDBInstances`) and pricing (`GetProducts`) are refreshed in the background on a ticker rather than on the Prometheus scrape path. A scrape reads the warm cache and makes zero AWS calls, so scrape latency no longer tracks AWS API latency or a cold pricing map. Warming starts at startup, so there is a short cold-start gap: no metrics emit until the first background refresh finishes. When a refresh fails to warm a price for an instance, that instance is skipped and `cloudcost_exporter_aws_rds_populate_errors_total` is incremented; the scrape still succeeds.
 
 ## Configuration
 
@@ -33,7 +31,7 @@ Or via command line:
 
 ### Per-region Timeout
 
-The background refresh queries `DescribeDBInstances` and pricing in each region concurrently. `--aws.rds.region-timeout` bounds each region's work:
+The background refresh lists instances (`DescribeDBInstances`) and prices (`GetProducts`) in each region concurrently. `--aws.rds.region-timeout` bounds each region's work:
 
 ```bash
 --aws.rds.region-timeout=15s
@@ -51,7 +49,8 @@ The default is `0`, which applies an internal safety ceiling so a hung AWS call 
 
 ## Notes
 
-- Pricing data is fetched from the AWS Pricing API (us-east-1 region)
+- Pricing is listed in bulk per region from the AWS Pricing API and keyed by instance type, engine, deployment option, and license so `Collect` matches each instance to its price
+- Instances whose engine is not recognized are skipped; recognized engines are MySQL, MariaDB, PostgreSQL, Aurora MySQL, Aurora PostgreSQL, Oracle, and SQL Server
 - Inventory and pricing are refreshed in the background on the configured scrape interval; scrapes read the warm cache and make no AWS calls
 - No metrics emit until the first background refresh completes after startup
 - All costs are represented in USD per hour
--- a/docs/metrics/aws/rds.md
+++ b/docs/metrics/aws/rds.md
@@ -50,6 +50,7 @@ The default is `0`, which applies an internal safety ceiling so a hung AWS call 
 ## Notes
 
 - Pricing is listed in bulk per region from the AWS Pricing API and keyed by instance type, engine, deployment option, and license so `Collect` matches each instance to its price
+- Aurora MySQL and Aurora PostgreSQL additionally key on storage mode (Standard vs I/O-Optimized), since I/O-Optimized carries a different instance-hour price; an Aurora instance with an unrecognized or missing storage mode is skipped rather than priced against the wrong SKU
 - Instances whose engine is not recognized are skipped; recognized engines are MySQL, MariaDB, PostgreSQL, Aurora MySQL, Aurora PostgreSQL, Oracle, and SQL Server
 - Instance inventory and pricing refresh independently in the background, each in its own store: inventory refreshes on `--scrape-interval` (default `1h`); pricing refreshes on a fixed 24-hour interval, since it's a stable bulk-per-region fetch that doesn't need frequent refresh
 - Scrapes read both warm stores and make no AWS calls
--- a/docs/metrics/aws/rds.md
+++ b/docs/metrics/aws/rds.md
@@ -8,7 +8,7 @@
 
 | Metric name                                      | Metric type | Description                                                              | Labels                                                                                                                              |
 |--------------------------------------------------|-------------|-------------------------------------------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------|
-| cloudcost_exporter_aws_rds_populate_errors_total | Counter     | Errors during background store population by store, region, and operation | `store`=&lt;always `instances`&gt; <br/> `region`=&lt;AWS region&gt; <br/> `operation`=&lt;`lookup_client`, `list_instances`, `list_prices`, `parse_pricing`&gt; |
+| cloudcost_exporter_aws_rds_populate_errors_total | Counter     | Errors during background store population by store, region, and operation | `store`=&lt;`instances` or `pricing`&gt; <br/> `region`=&lt;AWS region&gt; <br/> `operation`=&lt;`lookup_client`, `list_instances`, `list_prices`, `parse_pricing`&gt; |
 
 ## Overview
 
@@ -31,13 +31,13 @@ Or via command line:
 
 ### Per-region Timeout
 
-The background refresh lists instances (`DescribeDBInstances`) and prices (`GetProducts`) in each region concurrently. `--aws.rds.region-timeout` bounds each region's work:
+Instance inventory (`DescribeDBInstances`) and pricing (`GetProducts`) refresh independently, each in its own background store with its own per-region call. `--aws.rds.region-timeout` bounds each region's work in both stores:
 
 ```bash
 --aws.rds.region-timeout=15s
 ```
 
-The default is `0`, which applies an internal safety ceiling so a hung AWS call can never stall the background refresh. Set a positive value (e.g. `15s`) to fail slow or unreachable regions faster: the region is logged, counted in `cloudcost_exporter_aws_rds_populate_errors_total`, and skipped while the others still warm. Because the refresh runs off the scrape path, a slow region delays only the freshness of that region's data, not the scrape itself.
+The default is `0`, which applies an internal safety ceiling so a hung AWS call can never stall either background refresh. Set a positive value (e.g. `15s`) to fail slow or unreachable regions faster: the region is logged, counted in `cloudcost_exporter_aws_rds_populate_errors_total`, and skipped while the others still warm. Because the refresh runs off the scrape path, a slow region delays only the freshness of that region's data, not the scrape itself.
 
 ## Labels
 
@@ -51,8 +51,9 @@ The default is `0`, which applies an internal safety ceiling so a hung AWS call 
 
 - Pricing is listed in bulk per region from the AWS Pricing API and keyed by instance type, engine, deployment option, and license so `Collect` matches each instance to its price
 - Instances whose engine is not recognized are skipped; recognized engines are MySQL, MariaDB, PostgreSQL, Aurora MySQL, Aurora PostgreSQL, Oracle, and SQL Server
-- Inventory and pricing are refreshed in the background on the configured scrape interval; scrapes read the warm cache and make no AWS calls
-- No metrics emit until the first background refresh completes after startup
+- Instance inventory and pricing refresh independently in the background, each in its own store: inventory refreshes on `--scrape-interval` (default `1h`); pricing refreshes on a fixed 24-hour interval, since it's a stable bulk-per-region fetch that doesn't need frequent refresh
+- Scrapes read both warm stores and make no AWS calls
+- No metrics emit until both the instance and pricing background stores complete their first refresh after startup
 - All costs are represented in USD per hour
 
 ## IAM Permissions
--- a/pkg/aws/rds/pricing_map.go
+++ b/pkg/aws/rds/pricing_map.go
@@ -10,7 +10,26 @@ import (
 )
 
 type AWSPriceData struct {
-	Terms *AWSTerms `json:"terms"`
+	Product *AWSProduct `json:"product"`
+	Terms   *AWSTerms   `json:"terms"`
+}
+
+type AWSProduct struct {
+	Attributes *AWSProductAttributes `json:"attributes"`
+}
+
+// AWSProductAttributes holds the RDS Database Instance product attributes that
+// make up a pricing key. These come straight from the AWS Pricing API, so the
+// engine and edition are display names ("PostgreSQL", "Oracle"), not the
+// instance Engine codes ("postgres", "oracle-ee").
+type AWSProductAttributes struct {
+	InstanceType     string `json:"instanceType"`
+	RegionCode       string `json:"regionCode"`
+	DatabaseEngine   string `json:"databaseEngine"`
+	DatabaseEdition  string `json:"databaseEdition"`
+	DeploymentOption string `json:"deploymentOption"`
+	LicenseModel     string `json:"licenseModel"`
+	LocationType     string `json:"locationType"`
 }
 
 type AWSTerms struct {
@@ -50,24 +69,20 @@ func (pm *pricingMap) Get(key string) (float64, bool) {
 	return v, ok
 }
 
-func validateRDSPriceData(ctx context.Context, priceList string) (float64, error) {
-	var priceData AWSPriceData
-	if err := json.Unmarshal([]byte(priceList), &priceData); err != nil {
-		slog.ErrorContext(ctx, "error unmarshaling price JSON", "error", err)
-		return 0, err
-	}
-
-	if priceData.Terms == nil {
+// priceFromTerms extracts the single on-demand USD price from a product's
+// pricing terms.
+func priceFromTerms(ctx context.Context, terms *AWSTerms) (float64, error) {
+	if terms == nil {
 		slog.ErrorContext(ctx, "Terms is nil")
 		return 0, fmt.Errorf("terms is nil")
 	}
-	if priceData.Terms.OnDemand == nil {
+	if terms.OnDemand == nil {
 		slog.ErrorContext(ctx, "OnDemand is nil")
 		return 0, fmt.Errorf("OnDemand is nil")
 	}
 
 	var term *AWSTerm
-	for _, t := range priceData.Terms.OnDemand {
+	for _, t := range terms.OnDemand {
 		if t == nil || t.PriceDimensions == nil {
 			slog.ErrorContext(ctx, "PriceDimensions is nil")
 			return 0, fmt.Errorf("PriceDimensions is nil")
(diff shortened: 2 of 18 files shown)


docs/metrics/aws/rds.md
```diff
@@ -4,10 +4,18 @@
...
 
+### Operational Metrics
+
+| Metric name                                      | Metric type | Description                                                              | Labels                                                                                                                              |
+|--------------------------------------------------|-------------|-------------------------------------------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------|
+| cloudcost_exporter_aws_rds_populate_errors_total | Counter     | Errors during background store population, by store, region, and operation | `store`=&lt;always `instances`&gt; <br/> `region`=&lt;AWS region&gt; <br/> `operation`=&lt;`lookup_client`, `list_instances`, `get_pricing`, `validate_pricing`&gt; |
```
logyball: super NIT


```suggestion
| cloudcost_exporter_aws_rds_populate_errors_total | Counter     | Errors during background store population by store, region, and operation | `store`=&lt;always `instances`&gt; <br/> `region`=&lt;AWS region&gt; <br/> `operation`=&lt;`lookup_client`, `list_instances`, `get_pricing`, `validate_pricing`&gt; |
```

docs/metrics/aws/rds.md
```diff
@@ -4,10 +4,18 @@
...
+
 ## Overview
 
 The RDS collector exports pricing metrics for Amazon Relational Database Service instances across all configured AWS regions. It collects hourly pricing rates for RDS database instances based on their instance type and tier.
 
+Instance inventory (`DescribeDBInstances`) and pricing (`GetProducts`) are refreshed in the background on a ticker rather than on the Prometheus scrape path. A scrape reads the warm cache and makes zero AWS calls, so scrape latency no longer tracks AWS API latency or a cold pricing map. Warming starts at startup, so there is a short cold-start gap: no metrics emit until the first background refresh finishes. When a refresh fails to warm a price for an instance, that instance is skipped and `cloudcost_exporter_aws_rds_populate_errors_total` is incremented; the scrape still succeeds.
```
logyball: Would recommend ditching this.  This is describing a "how", which isn't required in a top-level MD document.

pkg/aws/rds/rds.go
```diff
@@ -55,125 +53,81 @@ const (
...
 
-// New creates an rds collector. A RegionListTimeout of 0 leaves each region's
-// DescribeDBInstances call bounded only by the shared collector context
-// (-collector-interval); a positive value caps it per region so a slow or
-// unreachable region fails fast instead of overrunning the scrape.
-func New(_ context.Context, config *Config, logger *slog.Logger) (*Collector, error) {
+// New creates an rds collector.
+//
+// Instance inventory and pricing are refreshed in the background on a ticker
+// rather than on the scrape path (see background-store-pattern.md). New()
```
logyball: does this doc exist?

logyball (changes requested):
Hey @stephan-rayner - this is close to what we're looking for in my opinion.  It definitely offers a performance improvement over what is happening currently.

My issue with it and what I think we should address before merging is that it doesn't quite follow the same pattern as the other "background fetches" for those resources that we do fetch in the background.

The general process is split the price of things and the listing of things into two separate processes that are united with a "key" of some sort, so that you can map the instance <--> price.

For example, in the nodes example, we:
- [list the nodes](https://github.com/grafana/cloudcost-exporter/blob/fb2c9f503867ce5e38d62b60f7467a6e3bb8706a/pkg/google/gke/node_store.go#L127) using an errgroup
- [get the pricing data](https://github.com/grafana/cloudcost-exporter/blob/fb2c9f503867ce5e38d62b60f7467a6e3bb8706a/pkg/google/gke/pricing_map.go#L231) separately

You can see the pricing and the node/volumes being collected [separately here](https://github.com/grafana/cloudcost-exporter/blob/fb2c9f503867ce5e38d62b60f7467a6e3bb8706a/pkg/google/gke/gke.go#L185) when the GKE Collector is initialized.

And when a metric is scraped (just using nodes for example):

- [a list of nodes is read](https://github.com/grafana/cloudcost-exporter/blob/fb2c9f503867ce5e38d62b60f7467a6e3bb8706a/pkg/google/gke/gke.go#L103) - recall these were collected in the background
- [for each of those instances, costs are grabbed from the pricing map](https://github.com/grafana/cloudcost-exporter/blob/fb2c9f503867ce5e38d62b60f7467a6e3bb8706a/pkg/google/gke/gke.go#L112) using that "key" that unites the "thing" to the "price"

What's different about what you're doing here that I think needs to be changed before we can merge this is that you're create a store here that starts population when the collector is registered (good), but it makes the pricing _dependent_ on the instances:

- [instances are listed](https://github.com/grafana/cloudcost-exporter/blob/ce2e7cce9bad5091d4073660a81c75ebf67ef385/pkg/aws/rds/store.go#L174)
- THEN [pricing is determined](https://github.com/grafana/cloudcost-exporter/blob/ce2e7cce9bad5091d4073660a81c75ebf67ef385/pkg/aws/rds/store.go#L210) for the individual instances

This keeps things coupled.  You could argue that this prevents fetching a bit list of prices that we'll never use (True), but I'd argue that the storage/listing of prices is _very_ cheap and does not need to be done often at all (1 time a day, at most), and ensures that we won't have RDS instance <--> pricing errors.

So my suggestion is to follow the established pattern:

- define "key" that uniquely maps an instance to a price
- list All RDS prices per region (seems wasteful, but you're only going to store basically key -> price in the pricing map)
- list all RDS instances we have 
- store the two things above in separate maps
- when collect is capped, iterate through the instances and match them up to prices

WDYT?

pkg/aws/rds/rds.go
stephan-rayner: ah woops, technically it does on my local I created it for myself. I'll pull that reference

docs/metrics/aws/rds.md
stephan-rayner: Agreed this doesn't need to be here it can be deleted

pkg/aws/rds/store.go
```diff
@@ -0,0 +1,220 @@
...
+	timeout := s.regionListTimeout
+	if timeout <= 0 {
+		timeout = defaultPopulateTimeout
+	}
+	ctx, cancel := context.WithTimeout(ctx, timeout)
+	defer cancel()
+
+	s.populateInstances(ctx, regionName, regionClient)
+	s.populatePricing(ctx, regionName)
+}
```
logyball, line 180: Here we're still coupling the instances and the pricing - they're still underneath the same root call:

1. `newInstanceStore()` OR Collector -> `startRefreshTicker`
2. `Populate()`
3. for region -> go `populateRegion()`

Either way, you have the initial population, or you have the more common case: startRefreshTicker, using the scrape interval from the config (default 1hr) tells `populate` to be called every time the ticker... ticks.

What we've got here then does do one of the important parts of this development which is decoupling the scrape itself from the population of the data, but it's still not decoupling the pricing fetching from the instance fetching.  Even though they're on separate function calls, they currently all fall under the same code path when they could be fully independent of one another.

One thing you will have to contend with is their independent ability to be "ready", e.g. there are cases where a scrape can happen and:
1. no instances are present in the cache
2. no prices are present in the cache

any combination of (1) and (2), and all should likely be handled with skipping the metrics

pkg/aws/rds/store.go
```diff
@@ -0,0 +1,220 @@
...
+// newInstanceStore returns a store that begins warming immediately in the
+// background. Warming at startup rather than on the first scrape closes the
+// cold-start pricing gap that drove the RDS p99.
+func newInstanceStore(ctx context.Context, logger *slog.Logger, config *Config, pm *pricingMap, populateErrors *prometheus.CounterVec) *instanceStore {
+	s := &instanceStore{
+		logger:            logger.With("store", "instances"),
+		regions:           config.Regions,
+		regionMap:         config.RegionMap,
+		pricingClient:     config.Client,
+		pricingMap:        pm,
```
logyball, line 97: if we decouple pricing and instances, we don't need pointers to the pricing map/client in the instance store

pkg/aws/rds/pricing_map.go
```diff
@@ -100,3 +115,45 @@ func validateRDSPriceData(ctx context.Context, priceList string) (float64, error
...
+	price, err := priceFromTerms(ctx, priceData.Terms)
+	if err != nil {
+		return "", 0, false
+	}
+	return key, price, true
+}
+
+// priceKeyFromAttributes builds the pricing key from a product's attributes. It
+// returns ok=false when an attribute the key depends on is missing.
+func priceKeyFromAttributes(product *AWSProduct) (string, bool) {
```
logyball, line 143: are we certain about these being the uniqueness parameters here?

I'm doing some research here, but I'm wondering if there are multiple prices available for these things, especially with the aurora vs non-aurora disticntion.

logyball: Ok I've been struggling to find a good way to do this, and I haven't yet found a 100% way to easily map DB instances -> unique prices 😮‍💨

logyball: The pricing description is a pretty good proxy, I imagine:

`$ 0.805 per RDS db.m5d.xlarge Multi-AZ instance hour (or partial hour) running Oracle SE2 (BYOL)"`

`<price> per RDS <instance size> <deployment type> instance hour (or partial hour) running <engine> <license>`

So I think the only thing we're missing here is the engine coule also include aurora, as we have things like PostgreSQL and Aurora PostgreSQL. 

Not sure how much we're using them, but worth checking

logyball: If we're interested in fixing this here, would recommend adding Aurora as an additional attribute on the pricing key / AWS Attributes, but that can definitely wait til next PR in my opinion.

logyball (changes requested):
See this comment here: https://github.com/grafana/cloudcost-exporter/pull/1109/changes#r3824386222

Essentially what we're looking for here is _decoupling_ the two calls:
- listing RDS instances
- listing RDS prices

Right now they're still called under one `populateRegion` function.  

If they're separated (check out how [the GKE collector does it](https://github.com/grafana/cloudcost-exporter/blob/main/pkg/google/gke/gke.go)) then we can have separate intervals to refresh everything:
1. long interval for prices
2. relatively short interval for RDS instances

The critical portion will be making sure that the key that we use to map instance <--> price is durable.

Last thing: can you update the PR description to reflect the new changes?

pkg/aws/rds/store.go
stephan-rayner: Good spot, very good spot. Let's make them independent. Ya I think skipping when things are not fully ready is the way!

pkg/aws/rds/store.go
stephan-rayner: 100% agree

pkg/aws/rds/pricing_map.go
stephan-rayner: > Ok I've been struggling to find a good way to do this, and I haven't yet found a 100% way to easily map DB instances -> unique prices 😮‍💨

I'm glad it wasn't just me to struggled with this haha.

Ya I can add the engine and Aurora for sure that makes sense to me

pkg/aws/rds/store.go
stephan-rayner: Done, pulled the pricing client/map pointers out of instanceStore entirely. It doesn't know pricing exists anymore.

pkg/aws/rds/store.go
stephan-rayner: Split this into two fully independent stores now, instanceStore and pricingStore, each with their own Populate/ticker/readiness signal. Instances still refresh on the scrape interval, pricing gets its own 24h ticker since it's cheap and stable. Collect checks both stores' readiness separately and skips for any combination of not-ready (neither, instances-only, pricing-only) instead of one shared gate.

pkg/aws/rds/pricing_map.go
stephan-rayner: Filed this as a follow-up since we agreed it can wait: https://github.com/grafana/cloudcost-exporter/issues/1131. Left a TODO near `engineAttributes` pointing at it too.

pkg/aws/rds/rds.go
```diff
@@ -195,8 +173,97 @@ func isOutpostsInstance(instance rdsTypes.DBInstance) string {
...
+// API's display names ("PostgreSQL", "Oracle" plus edition "Enterprise"), so we
+// translate explicitly. Engines absent from this map cannot be priced and are
+// skipped by pricingKeyFor.
+//
+// VERIFY: the open-source rows (MySQL/MariaDB/PostgreSQL/Aurora) are confirmed,
+// but the Oracle and SQL Server databaseEdition strings are best-effort and
+// should be checked against a real GetProducts response before relying on their
+// prices.
+//
+// TODO: harden Aurora and multi-edition engine matching against a real
```
simaarfania: It seems like Serverless v2 might be a separate problem from the Aurora key rather than a wider version of it. Those SKUs look like they carry `productFamily = "ServerlessV2"` rather than Database Instance, so listRDSPrices may not return them at all, maybe we should remove this TODO here?

pkg/aws/rds/rds.go
```diff
@@ -195,8 +173,97 @@ func isOutpostsInstance(instance rdsTypes.DBInstance) string {
...
+// consult this map; see pricingKeyFor.
+//
+// VERIFY: the "License included" / "Bring your own license" strings are
+// best-effort and should be checked against a real GetProducts response.
+var licenseModelAttributes = map[string]string{
+	"license-included":       "License included",
+	"bring-your-own-license": "Bring your own license",
+}
+
+func createPricingKey(region, instanceType, databaseEngine, databaseEdition, depOption, licenseModel, locationType string) string {
```
simaarfania: I think it's possible we aren't getting the right rates for Aurora instances, it seems the pricing key doesn't separate Aurora I/O-Optimized from Aurora Standard, so the two SKUs may be colliding on one entry and we could end up with whichever price gets parsed last.

simaarfania (dismissed):

pkg/aws/rds/rds.go
stephan-rayner: Oh that is worth looking into let me take a look at that. If you are right that really needs fixing

pkg/aws/rds/rds.go
stephan-rayner: Ya I am with you I fought with claude about that but it was insistent and I think I should have trusted my gut. Thanks for the backup Sima I'll take that out.

pkg/aws/rds/rds.go
stephan-rayner: Confirmed, you're right, checked it against the live pricing feed. Same instanceType/engine/deployment/license/location produces two SKUs for a lot of Aurora classes (`EBS Only` vs `Aurora IO Optimization Mode` in `storage`), and they collide on the current key today. It's trickier than just adding `storage` to the key though, that field means something completely different for NVMe-backed instance classes, and I'm not sure yet whether `DescribeDBInstances` even exposes I/O-Optimized at the instance level or only `DescribeDBClusters` does. Filed it as its own follow-up with the details: https://github.com/grafana/cloudcost-exporter/issues/1134

pkg/aws/rds/rds.go
stephan-rayner: Done, pulled the Serverless v2 mention out of the TODO. Just points at #1131 now.

pkg/aws/rds/pricing_map.go
```diff
@@ -100,3 +118,55 @@ func validateRDSPriceData(ctx context.Context, priceList string) (float64, error
 
 	return price, nil
 }
+
+// parseRDSPriceProduct turns a single Pricing API product into its pricing key
+// and on-demand USD price. ok is false when the product lacks the attributes or
+// on-demand price needed to key it, so the caller can count and skip it.
+func parseRDSPriceProduct(ctx context.Context, priceList string) (key string, price float64, ok bool) {
```
logyball, line 125: [NIT] 

regarding error handling here - right now this logs errors using `slog.ErrorContext()`, which does keep the context (nice) but drops the component and other nice things that we get from the component logger.  Not blocking, but I think it'd be preferred to return an actual structured/named error here, which can be logged by the original caller.  

Note that some folks will disagree with me :) (not to name any names, cough cough `@\inkel`) - but I think that passing the error up the chain and then logging at the caller is the right move.  Currently when Calling `parseRDSPriceProduct` from populateRegion, we get an incremented error if something goes wrong, but I don't think we ever actually log anywhere _what_ went wrong.

So my sugggestion would be to return an actual fmt.Error() with details about what happened _where_ the error occurs, and log it at the original caller to gain some visibility.

pkg/aws/rds/pricing_store.go
```diff
@@ -0,0 +1,133 @@
...
+		logger:            logger.With("store", "pricing"),
+		pricingClient:     config.Client,
+		regions:           config.Regions,
+		regionListTimeout: config.RegionListTimeout,
+		concurrency:       populateConcurrency,
+		populateErrors:    populateErrors,
+		prices:            newPricingMap(),
+		initialPopulation: make(chan struct{}),
+	}
+	go s.Populate(ctx)
```
logyball, line 56: I'm thinking we should have a check here that blocks healthiness on a total pricing call failure.  If for some reason we don't populate _anything_ here, we are still returning a "healthy" pricing store, which will cause all subsequent scrapes to fail for 24 hours.  

Take a look at [GKE](https://github.com/grafana/cloudcost-exporter/blob/b530dc41ebaa1f56e572e1880451672329a8ca60/pkg/google/gke/pricing_map.go#L119): 

```go
// NewPricingMap returns a new PricingMap in a way that can be used afterwards.
func NewPricingMap(ctx context.Context, gcpClient client.Client) (*PricingMap, error) {
	pm := &PricingMap{
		compute:   map[string]*FamilyPricing{},
		storage:   map[string]*StoragePricing{},
		gcpClient: gcpClient,
	}

	if err := pm.Populate(ctx); err != nil {
		return nil, err
	}

	return pm, nil
}
```

When running initial population (New), the pricing call is synchronous and fails with an error initially, and then subsequently logs it.

I think this is the correct approach as:
1. case 1: initial failure.   Current approach in RDS: We have a blank exporter (nothing functioning correctly for 24 hours).  Proposed approach in RDS: we have a fast failing exporter, which will cause some action.
2. case 2: subsequent failure: we have a logged and recorded error and stale pricing
3. case 3: no failures: happy days

WDYT?

logyball: Actually now that I'm looking at it - I don't think there's any way for a pricing store error to be surfaced to the collector.  A full or partial outage (1 ... N regions) will be swallowed by the Populate() call, and the collector has no way of knowing besides the metric 🤔

logyball (changes requested):
One more change req: pricing failure on init is not picked up, will only self-resolve after 24hours

pkg/aws/rds/pricing_store.go
stephan-rayner: Fixed in 4a9c498. Didn't go with the exact GKE/`NewPricingMap` pattern (synchronous populate in the constructor, fail the constructor on error), since that would re-block `New()` on the Pricing API, which is the thing this PR was built to avoid.

Instead, mirrored the AKS VM price store's adaptive-interval pattern (`aks.go`/`vm_price_store.go`): `Populate` now returns whether every region's listing succeeded. `startPricingRefreshTicker` owns the whole lifecycle including the first attempt, and after any populate with a region failure, the next attempt runs in 30 minutes instead of waiting the full 24h refresh interval. So the "ready but broken for a full day" window you flagged shrinks to 30 minutes, and it self-heals without any change to `New()`'s non-blocking startup.

Doesn't fully solve "no way for the collector to know," in the sense that `Collect()` still has no signal beyond the counter, but it directly addresses the actual failure mode (a broken store looking healthy for a long time) and I think closes the gap enough in practice. Let me know if you think the collector-visibility half still needs its own fix.

pkg/aws/rds/pricing_map.go
stephan-rayner: Agreed this is the right direction, haven't gotten to it yet: still logging from inside `parseRDSPriceProduct` via `slog.ErrorContext` rather than returning a structured error for `populateRegion` to log with full component context. Left it out of this pass since it's observability-only, not a correctness issue. Will follow up in a future PR rather than block on it here, unless you feel otherwise.

stephan-rayner:
One subtlety worth calling out explicitly: `Done()` on both stores closes after the *first* populate attempt, regardless of whether that attempt succeeded (`pricing_store.go`, `store.go`). So if every region fails on the very first populate (bad IAM, a region-wide outage), the store latches "ready" forever with nothing cached, rather than staying "not ready" until something actually warms.

I don't think this needs different behaviour, for two reasons:

1. A store with nothing cached and a store that isn't "ready" produce the same observable outcome from `Collect`: zero metrics. There's no stale-but-wrong data being served, because nothing was ever cached to begin with.
2. The failure is still visible: every populate tick that fails increments `cloudcost_exporter_aws_rds_populate_errors_total`, which is the metric this PR already recommends alerting on.

So "ready" here means "we've attempted a populate," not "we've succeeded at least once," and that distinction doesn't currently change behaviour. Added `TestPricingStore_Done_ClosesEvenWhenFirstPopulateFailsEverywhere` and `TestInstanceStore_Done_ClosesEvenWhenFirstPopulateFailsEverywhere` to pin this down, since it's exactly the kind of thing that's easy to assume works the other way at a glance.

logyball (approved):
LGTM! Thanks for working through these issues.  I think this PR is in a great place.

Be sure to tag it appropriately so the release can go out.  I agree with the bot, this should be a minor version bump.

PR #1129, opened 2026-09-01
add Environment Variables to Container

### LINKED ISSUE #1128: helm: add 'env:' to deployment container spc
### Cloud Provider

_No response_

### Feature Request

I would like to extend the helm deployment template, so we can set environment variables for the pod.

specific usecase is a need for proxy configs in our environment,

schwatvogel:
resolves #1128

just adds the `env:` structure to the deployment and to values.yaml


logyball:
TBH looks fine to me.  @nikimanoledaki any thoughts here?  We don't use environment variables for any `cloudcost-exporter` configuration AFAIK, but it doesn't break default behavior or have any downsides that I can think of.

